PR #16, opened 2019-04-11
Fix bug with Priority calculation in Candidate

### LINKED ISSUE #14: Candidate.Priority() is completely wrong
https://github.com/pion/ice/blob/7245ac0866456f7c39a534f496d0f346ab7aad30/candidate.go#L212

`^` is XOR in Go, not exponentiation. `2^24 == 26`.

It should be `(1<<24) * c.Type().Preference()` or simply `c.Type().Preference() << 24`.

`uint16` is also not large enough to hold the correct value.

maxhawkins:
Was using XOR not <<.

Fixes #14

### DIFF
--- a/candidate.go
+++ b/candidate.go
@@ -202,16 +202,16 @@ func (c *Candidate) writeTo(raw []byte, dst *Candidate) (int, error) {
 }
 
 // Priority computes the priority for this ICE Candidate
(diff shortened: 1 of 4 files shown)


Sean-Der:
@maxhawkins anything you want to keep from this PR, or ok to close? thanks!

PR #17, opened 2019-04-11
Candidate priority is incorrectly calculated

### LINKED ISSUE #14: Candidate.Priority() is completely wrong

mjmac:
Some constants were being incorrectly generated using 2^N (XOR)
instead of 1 << N. This resulted in incorrect candidate priorities
being used to evaluate candidate pairs.

Resolves #14

### DIFF
--- a/candidate.go
+++ b/candidate.go
@@ -202,16 +202,16 @@ func (c *Candidate) writeTo(raw []byte, dst *Candidate) (int, error) {
 }
 
 // Priority computes the priority for this ICE Candidate
-func (c *Candidate) Priority() uint16 {
+func (c *Candidate) Priority() uint32 {
 	// The local preference MUST be an integer from 0 (lowest preference) to
 	// 65535 (highest preference) inclusive.  When there is only a single IP
 	// address, this value SHOULD be set to 65535.  If there are multiple
 	// candidates for a particular component for a particular data stream
 	// that have the same type, the local preference MUST be unique for each
 	// one.
-	return (2^24)*c.Type.Preference() +
-		(2^8)*c.LocalPreference +
-		(2^0)*(256-c.Component)
+	return (1<<24)*uint32(c.Type.Preference()) +
+		(1<<8)*uint32(c.LocalPreference) +
+		uint32(256-c.Component)
 }
 
 // Equal is used to compare two CandidateBases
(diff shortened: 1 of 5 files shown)


candidatepair.go
```diff
@@ -73,7 +73,7 @@ func (p *candidatePair) Priority() uint32 {
 		return 0
 	}
 
-	return (2^32)*min(g, d) + 2*max(g, d) + cmp(g, d)
+	return (2<<32)*uint64(min(g, d)) + uint64(2*max(g, d)+cmp(g, d))
```
maxhawkins: Shouldn't this be `1 << 32 `?

maxhawkins: `2*max(g, d)` overflows if `g == 1<<32-1`

mjmac: yep. fixed.

maxhawkins: I think there's still an overflow with `2*max(g, d)`

candidate.go
```diff
@@ -202,16 +202,16 @@ func (c *Candidate) writeTo(raw []byte, dst *Candidate) (int, error) {
...
 	// The local preference MUST be an integer from 0 (lowest preference) to
 	// 65535 (highest preference) inclusive.  When there is only a single IP
 	// address, this value SHOULD be set to 65535.  If there are multiple
 	// candidates for a particular component for a particular data stream
 	// that have the same type, the local preference MUST be unique for each
 	// one.
-	return (2^24)*c.Type.Preference() +
-		(2^8)*c.LocalPreference +
-		(2^0)*(256-c.Component)
+	return (1<<24)*uint32(c.Type.Preference()) +
```
kixelated: `uint32(c.Type.Preference()) << 24` is the same result

mjmac: > `uint32(c.Type.Preference()) << 24` is the same result

Granted, but it's less readable IMO. Especially when comparing the implementation to the spec.

PR #18, opened 2019-04-12
Send connectivity checks right on agent start

### LINKED ISSUE #15: startConnectivityChecks adds unnecessary delay
It creates a ticker, but tickers don't fire immediately. This adds 2 seconds before connectivity checks are actually initiated.

Sean-Der:
Connectivity checks are done via a Ticker, the inital tick does not
happen immediately (by design) causing the startup time to be the Duration the
Ticker is created with.

This change adds another chan forceCandidateContact that can be used to
force contact at anytime. Currently it is only called on startup, but
could be useful in the future for reuse.

Resolve #15

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -19,7 +19,7 @@ import (
 
 const (
 	// taskLoopInterval is the interval at which the agent performs checks
-	taskLoopInterval = 2 * time.Second
+	defaultTaskLoopInterval = 2 * time.Second
 
 	// keepaliveInterval used to keep candidates alive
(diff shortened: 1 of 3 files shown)


maxhawkins:
No tests?

Sean-Der:
Good point! I will write some that assert that we get an initial request (and that we can request at anytime)

Sean-Der:
@maxhawkins test added! can I get a re-review w/e you have a chance. thanks.

PR #22, opened 2019-04-13
Assert MessageIntegrity/Username for inbound

### LINKED ISSUE #19: stun.Username and stun.MessageIntegrity are not checked
Pions does not actually check if the remote username/password is correct. A bad entity could use this to hijack a session by sending a STUN packet with maxed out priority. The contents would still be encrypted using DTLS/SRTP, but it would kill the session and open up those protocols to man-in-the-middle attacks.

Sean-Der:
When we get an inbound message assert these values, also discard any
other packet types besides binding. In the future we should extend to
handle inbound error messages

Resolves #19

----

Still need to write tests, but after #15 lands and I rebase this would appreciate if people could help me manually test everything (before I make a stupid mistake)

* [x] FireFox (Offer + Answer)
* [x] Chrome (Offer + Answer)
* [ ] PRFLX
* [x] SRFLX

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -3,6 +3,8 @@
 package ice
 
 import (
+	"bytes"
+	"encoding/binary"
 	"fmt"
 	"math/rand"
 	"net"
@@ -29,6 +31,8 @@ const (
 
 	// the number of bytes that can be buffered before we start to error
 	maxBufferSize = 1000 * 1000 // 1MB
+
+	stunAttrHeaderLength = 4
 )
 
 // Agent represents the ICE agent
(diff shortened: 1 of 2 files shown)


maxhawkins:
Could you annotate with references to the spec so I can compare?

Sean-Der:
Hey @maxhawkins 

I didn't go off the spec on this unfortunately :/ I just knew what needed to be done from reading pcaps before.

Looking through RFC now, but I do see https://tools.ietf.org/html/rfc5245#section-2.5

Sean-Der:
@maxhawkins FireFox+Chrome Host+Srflx all work for me.

Conceptually I don't see any reason Prflx would regress (and we have tests for it) so I feel good to go.

agent.go
```diff
@@ -768,9 +770,80 @@ func (a *Agent) handleNewPeerReflexiveCandidate(local *Candidate, remote net.Add
...
+
+	if !usernameFound {
+		return fmt.Errorf("inbound packet missing Username")
+	} else if err := usernameAttr.Unpack(m, usernameRawAttr); err != nil {
+		return err
+	}
+
+	expectedUsername := a.localUfrag + ":" + a.remoteUfrag
+	if usernameAttr.Username != expectedUsername {
+		return fmt.Errorf("MessageIntegrity mismatch expected(%x) actual(%x)", expectedUsername, usernameAttr.Username)
```
maxhawkins: s/MessageIntegrity/Username/

agent.go
```diff
@@ -768,9 +770,80 @@ func (a *Agent) handleNewPeerReflexiveCandidate(local *Candidate, remote net.Add
...
+	}
+
+	tailLength := 24
+	rawCopy := make([]byte, len(m.Raw))
+	copy(rawCopy, m.Raw)
+
+	if _, hasFingerprint := m.GetOneAttribute(stun.AttrFingerprint); hasFingerprint {
+		tailLength += 8
+
+		currLength := binary.BigEndian.Uint16(rawCopy[2:4])
```
maxhawkins: Can I see a diagram of this structure or something? Kind of mysterious what values  you're pulling out here.

agent.go
```diff
@@ -768,9 +770,80 @@ func (a *Agent) handleNewPeerReflexiveCandidate(local *Candidate, remote net.Add
...
+		binary.BigEndian.PutUint16(rawCopy[2:], currLength-8)
+	}
+
+	computedMessageIntegrity, err := stun.MessageIntegrityCalculateHMAC(key, rawCopy[:len(rawCopy)-tailLength])
```
maxhawkins:
```go
rawCopy[:len(rawCopy)-tailLength]
```

What if len`(rawCopy) == 1`?

agent_test.go
```diff
@@ -351,3 +366,67 @@ func TestConnectivityOnStartup(t *testing.T) {
...
+			t.Fatal("Binding with invalid Username was able to create prflx candidate")
+		}
+
+		a.handleInbound(buildMsg(stun.ClassRequest, a.localUfrag+":"+a.remoteUfrag, "Invalid"), local, remote)
+		if len(a.remoteCandidates) == 1 {
+			t.Fatal("Binding with invalid MessageIntegrity was able to create prflx candidate")
+		}
+	})
+
+	t.Run("Invalid Binding requests should be discarded", func(t *testing.T) {
```
maxhawkins: is this description supposed to be the same as the last one?

agent.go
```diff
@@ -768,9 +770,80 @@ func (a *Agent) handleNewPeerReflexiveCandidate(local *Candidate, remote net.Add
...
+		return fmt.Errorf("inbound packet missing MessageIntegrity")
+	} else if err := messageIntegrityAttr.Unpack(m, messageIntegrityRawAttr); err != nil {
+		return err
+	}
+
+	tailLength := 24
+	rawCopy := make([]byte, len(m.Raw))
+	copy(rawCopy, m.Raw)
+
+	if _, hasFingerprint := m.GetOneAttribute(stun.AttrFingerprint); hasFingerprint {
```
maxhawkins: I don't see any test that covers this if statement

Sean-Der: Added a test!

agent_test.go
Sean-Der: Fixed

agent.go
Sean-Der: Asserted length and early returned if too short

agent.go
Sean-Der: Updated this with comments,  just rewriting the STUN length header. I used a bunch of named variables (instead of magic numbers) hopefully a little better. If you still think it is bad tell me and can fix up

agent.go
Sean-Der: Fixed

PR #25, opened 2019-04-15
Discard non-STUN if no associated candidate

### LINKED ISSUE #20: CandidatePair is not validated on Read
Another security concern. If a non-STUN packet is received, it is blindly forwarded to the mux. This increases the attack vector of the application by exposing DTLS/SRTP/etc to arbitrary packets from the outside world. You can imagine an attack where a bad actor sends DTLS `ClientHello` packets to all possible ICE ports, and sometimes it will be able to initiate the handshake before the intended   peer (hijacking the connection).

The fix is to validate local/remote pairs. You need to receive a success from a given remote Candidate (including the correct username/password) before allowing non-STUN packets to be forwarded from that pair.

Sean-Der:
If a inbound message is not associated with a valid remote STUN
candidate discard it

Resolves #20

-----

@kixelated I believe all the issues you cited are resolved when this lands.

* Remote STUN messages are checked via MessageIntegrity (so no MITM)
* Remotes can't create prflx candidates (no DoS possible)
* We discard non-STUN messages if we don't have an associated candidate
   - Since we are dealing with lossy messages I think success just matters for outbound. I would prefer not to lose inbound messages from a candidate if it is valid.

### DIFF
--- a/candidate.go
+++ b/candidate.go
@@ -188,13 +188,17 @@ func (c *Candidate) recvLoop() {
 			}
 
 			continue
-		} else {
-			err = c.agent.run(func(agent *Agent) {
-				agent.noSTUNSeen(c, srcAddr)
-			})
-			if err != nil {
-				log.Warnf("Failed to handle message: %v", err)
-			}
+		}
+
+		isValidRemoteCandidate := make(chan bool, 1)
+		err = c.agent.run(func(agent *Agent) {
+			isValidRemoteCandidate <- agent.noSTUNSeen(c, srcAddr)
+		})
+
+		if err != nil {
+			log.Warnf("Failed to handle message: %v", err)
+		} else if !<-isValidRemoteCandidate {
+			log.Warnf("Discarded message from %s, not a valid remote candidate", c.addr())
 		}
 
 		// NOTE This will return packetio.ErrFull if the buffer ever manages to fill up.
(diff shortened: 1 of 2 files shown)


candidate.go
```diff
@@ -188,13 +189,19 @@ func (c *Candidate) recvLoop() {
...
-		} else {
-			err = c.agent.run(func(agent *Agent) {
-				agent.noSTUNSeen(c, srcAddr)
-			})
-			if err != nil {
-				log.Warnf("Failed to handle message: %v", err)
+		}
+
+		var isValidRemoteCandidate uint32
+		err = c.agent.run(func(agent *Agent) {
```
kixelated: AFAIK the `run` function will only block until the given `func(*Agent)` function is passed to the `taskLoop`. It won't actually block on the function being fully executed.

You have to block with a channel here; atomic is not good enough. That or try to follow the path of my PR and have a `runInline` function that does block.

Sean-Der: Fixed.

I am going to move everything that way though, I just want to change things one at a time.

candidate.go
```diff
@@ -188,13 +188,17 @@ func (c *Candidate) recvLoop() {
...
-		} else {
-			err = c.agent.run(func(agent *Agent) {
-				agent.noSTUNSeen(c, srcAddr)
-			})
-			if err != nil {
-				log.Warnf("Failed to handle message: %v", err)
-			}
+		}
+
+		isValidRemoteCandidate := make(chan bool)
```
kixelated: Should be buffered size 1 just in case.

Sean-Der: added buffering

PR #26, opened 2019-04-16
Validate inbound success messages

### LINKED ISSUE #21: Valid peers can create Prflx candidates for addresses they don't control
1. Attacker sends a BindRequest to Pions and gets saved as a validPair.
2. Attacker never replies to Pion's BindRequest messages (so it will stay a validPair not selectedPair).
3. Attacker/Pions performs DTLS handshake.
4. Attacker sends a BindRequest with a higher priority and a spoofed return address (the target), which is also saved as a validPair.
5. Pions now proceeds to send raw SRTP/data to the target address.
6. Attacker can send RTCP/keep-alive information to keep the connection open.

This allows an attacker to DDoS a target without sending the data itself.

The fix involves requiring a SuccessResponse from a CandidatePair before writing to it. This may go against what the ICE RFC requires. It might actually not be possible because of NAT stuff, which is potentially a huge issue.

Sean-Der:
Validate all inbound binding success messages, before
peers could create valid prflx candidates without doing
a full request response

Resolves #21

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -3,6 +3,7 @@
 package ice
 
 import (
+	"bytes"
 	"fmt"
 	"math/rand"
 	"net"
@@ -27,6 +28,9 @@ const (
 
 	// the number of bytes that can be buffered before we start to error
 	maxBufferSize = 1000 * 1000 // 1MB
+
+	// the number of outbound binding requests we cache
+	maxPendingBindingRequests = 50
 
 	stunAttrHeaderLength = 4
 )
@@ -94,6 +98,9 @@ type Agent struct {
 
 	buffer *packetio.Buffer
 
+	// LRU of outbound Binding request Transaction IDs
+	pendingBindingRequests [][]byte
+
 	// State for closing
 	done chan struct{}
 	err  atomicError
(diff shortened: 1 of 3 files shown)


Sean-Der:
Adding @backkem and @mjmac 

----


Updated the Agent to require a full request+response before creating a Valid Candidate. Before we would create prflx candidates just from a inbound request.

We can't do this because this would allow a malicious agent to forge packets and add candidates they don't own.

agent.go
```diff
@@ -297,6 +306,11 @@ func (a *Agent) pingCandidate(local, remote *Candidate) {
 	}
 
 	a.log.Tracef("ping STUN from %s to %s\n", local.String(), remote.String())
+	if overflow := len(a.pendingBindingRequests) - maxPendingBindingRequests; overflow > 1 {
+		a.pendingBindingRequests = a.pendingBindingRequests[overflow:]
```
mjmac: This LRU implementation is clever. I had to think about it for a minute before I understood it. My only concern here is that it might be worth logging a message (at DEBUG or higher) indicating that N unanswered binding requests are being culled. If there is a big flurry of bind requests generated (think multiple interfaces with ipv4/ipv6 addresses) I could see a situation with a peer responding to some requests slower than others and then the resulting rejections might be confusing without this additional piece of information.

Sean-Der: ACK! Will add that before merging

PR #30, opened 2019-04-25
Partial revert #26

hugoArregui:
Revert only the setValidPair changes


Sean-Der:
@backkem @hugoArregui 

So I *think* this re-introduces the issue.

 A pair should only be considered valid from a success response, but would let request create a valid candidate. Basically a `a.setValidPair` should only occur inside a `successResponse` block

PR #32, opened 2019-04-26
Assert that a bindingSuccessResponse is symmetric

Sean-Der:
We previously didn't correctly follow [0] This updates
the agent to assert that a bindingSuccessResponse is
for the address it was sent too.

https://tools.ietf.org/html/rfc8445#section-7.2.5.2.1

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -47,6 +47,11 @@ func (bp byPairPriority) Less(i, j int) bool {
 	return bp.candidatePairs[i].Priority() > bp.candidatePairs[j].Priority()
 }
 
+type bindingRequest struct {
+	transactionID []byte
+	destination   net.Addr
+}
+
 // Agent represents the ICE agent
 type Agent struct {
 	onConnectionStateChangeHdlr       func(ConnectionState)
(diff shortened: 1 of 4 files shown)


Sean-Der:
@kixelated reading the RFC I also need to validate inbound requests, that isn't possible with Windows though :/

That is correct right @backkem? You ran into this with TURN when you did `ipnet` is it worth putting that in a package so `ice` can use it? It sounds like we will need it.

agent.go
```diff
@@ -657,23 +720,18 @@ func (a *Agent) handleInbound(m *stun.Message, local *Candidate, remote net.Addr
...
+			return
 		}
-
-		a.addRemoteCandidate(pflxCandidate)
+	default:
 		return
 	}
-	remoteCandidate.seen(false)
 
+	remoteCandidate.seen(false)
```
hugoArregui: maybe this is not possible for other reasons but the code itself doesn't prevent to have a nil remoteCandidate here if the transaction is valid (as in, it was on pendingBindingRequest and the addr match

hugoArregui: Sean, hope you don't mind I took the liberty of refactoring this function a bit while review it, because the switch was messing me up: https://gist.github.com/hugoArregui/fbb4f020d096cedcae65cf8de1ba49ce if you think is good I can send a commit. The only functional difference I introduce (I hope!) is related to comment I did above, I'm always handling a null remoteCandidate.

PR #33, opened 2019-04-29
refactor handleInboundControlling/Controlled

hugoArregui:
Both methods pretty much do the same


hugoArregui:
@Sean-Der cannot merge because of the coverarge. Not sure if you want to merge it or remove the check.

Sean-Der:
@hugoArregui I think if you capitalize your commit subject line `refactor handleInboundControlling/Controlled` it should pass!

don't know why both aren't failing though :/

PR #34, opened 2019-04-29
Implement regular nomination

### LINKED ISSUE #28: Implement regular nomination
We need to stop using aggressive nomination. 

If we have multiple pairs this can cause us to have different choices on each side (since the last pair with `use-candidate` is chosen. 

Right now we just send every pair a 'USE-CANDIDATE' this means that each side may choose different candidates, all depending on when packets arrive.

Also then if you choose different candidates they will eventually die. ICE only does ping via SendIndications and we have no pongs. So each side is just sending pings and eventually each side dies since they haven't gotten any inbound traffic in 30 seconds.

hugoArregui:
Fixes #28

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -48,8 +48,9 @@ func (bp byPairPriority) Less(i, j int) bool {
 }
 
 type bindingRequest struct {
-	transactionID []byte
-	destination   net.Addr
+	transactionID  []byte
+	destination    net.Addr
+	isUseCandidate bool
 }
 
 // Agent represents the ICE agent
@@ -98,6 +99,7 @@ type Agent struct {
 	remotePwd        string
 	remoteCandidates map[NetworkType][]*Candidate
 
+	selector     pairCandidateSelector
 	selectedPair *candidatePair
 	validPairs   []*candidatePair
 
@@ -222,6 +224,7 @@ func NewAgent(config *AgentConfig) (*Agent, error) {
 	gatherCandidatesReflective(a, config.Urls, config.NetworkTypes)
 
 	go a.taskLoop()
+
 	return a, nil
 }
 
@@ -257,10 +260,18 @@ func (a *Agent) startConnectivityChecks(isControlling bool, remoteUfrag, remoteP
 	case remotePwd == "":
 		return ErrRemotePwdEmpty
 	}
+
 	a.haveStarted.Store(true)
 	a.log.Debugf("Started agent: isControlling? %t, remoteUfrag: %q, remotePwd: %q", isControlling, remoteUfrag, remotePwd)
 
 	return a.run(func(agent *Agent) {
+		if isControlling {
+			a.selector = &controllingSelector{agent: a, log: a.log}
+		} else {
+			a.selector = &controlledSelector{agent: a, log: a.log}
+		}
+		a.selector.Start()
+
 		agent.isControlling = isControlling
 		agent.remoteUfrag = remoteUfrag
 		agent.remotePwd = remotePwd
@@ -272,57 +283,6 @@ func (a *Agent) startConnectivityChecks(isControlling bool, remoteUfrag, remoteP
 		t := time.NewTicker(a.taskLoopInterval)
 		agent.connectivityChan = t.C
 	})
-}
-
-func (a *Agent) pingCandidate(local, remote *Candidate) {
-	var msg *stun.Message
-	var err error
-
-	// The controlling agent MUST include the USE-CANDIDATE attribute in
-	// order to nominate a candidate pair (Section 8.1.1).  The controlled
-	// agent MUST NOT include the USE-CANDIDATE attribute in a Binding
-	// request.
-
-	transactionID := stun.GenerateTransactionID()
-	if a.isControlling {
-		msg, err = stun.Build(stun.ClassRequest, stun.MethodBinding, transactionID,
-			&stun.Username{Username: a.remoteUfrag + ":" + a.localUfrag},
-			&stun.UseCandidate{},
-			&stun.IceControlling{TieBreaker: a.tieBreaker},
-			&stun.Priority{Priority: local.Priority()},
-			&stun.MessageIntegrity{
-				Key: []byte(a.remotePwd),
-			},
-			&stun.Fingerprint{},
-		)
-	} else {
-		msg, err = stun.Build(stun.ClassRequest, stun.MethodBinding, transactionID,
-			&stun.Username{Username: a.remoteUfrag + ":" + a.localUfrag},
-			&stun.IceControlled{TieBreaker: a.tieBreaker},
-			&stun.Priority{Priority: local.Priority()},
-			&stun.MessageIntegrity{
-				Key: []byte(a.remotePwd),
-			},
-			&stun.Fingerprint{},
-		)
-	}
-
-	if err != nil {
-		a.log.Debug(err.Error())
-		return
-	}
-
-	a.log.Tracef("ping STUN from %s to %s\n", local.String(), remote.String())
-	if overflow := len(a.pendingBindingRequests) - maxPendingBindingRequests; overflow > 1 {
-		a.log.Debugf("Discarded %d pending binding requests, pendingBindingRequests is full", overflow)
-		a.pendingBindingRequests = a.pendingBindingRequests[overflow:]
-	}
-
-	a.pendingBindingRequests = append(a.pendingBindingRequests, bindingRequest{
-		transactionID: transactionID,
-		destination:   remote.addr(),
-	})
-	a.sendSTUN(msg, local, remote)
 }
 
 func (a *Agent) updateConnectionState(newState ConnectionState) {
@@ -338,31 +298,50 @@ func (a *Agent) updateConnectionState(newState ConnectionState) {
 	}
 }
 
-func (a *Agent) setValidPair(local, remote *Candidate, selected, controlling bool) {
-	// TODO: avoid duplicates
-	p := newCandidatePair(local, remote, controlling)
-	a.log.Tracef("Found valid candidate pair: %s (selected? %t)", p, selected)
-
-	if selected {
-		// Notify when the selected pair changes
-		if !a.selectedPair.Equal(p) {
-			a.onSelectedCandidatePairChange(p)
+func (a *Agent) findValidPair(local, remote *Candidate) *candidatePair {
+	for _, p := range a.validPairs {
+		if p.local == local && p.remote == remote {
+			return p
 		}
-		a.selectedPair = p
-		a.validPairs = nil
-		// TODO: only set state to connected on selecting final pair?
-		a.updateConnectionState(ConnectionStateConnected)
-	} else {
-		// keep track of pairs with succesfull bindings since any of them
-		// can be used for communication until the final pair is selected:
-		// https://tools.ietf.org/html/draft-ietf-ice-rfc5245bis-20#section-12
-		a.validPairs = append(a.validPairs, p)
-		// Sort the candidate pairs by priority of the remotes
-		sort.Sort(byPairPriority{a.validPairs})
 	}
+	return nil
+}
+
+func (a *Agent) addValidPair(local, remote *Candidate) *candidatePair {
+	p := a.findValidPair(local, remote)
+	if p != nil {
+		a.log.Tracef("Candidate pair is already valid: %s", p)
+		return p
+	}
+
+	p = newCandidatePair(local, remote, a.isControlling)
+	a.log.Tracef("Found valid candidate pair: %s", p)
+
+	// keep track of pairs with succesfull bindings since any of them
+	// can be used for communication until the final pair is selected:
+	// https://tools.ietf.org/html/draft-ietf-ice-rfc5245bis-20#section-12
+	a.validPairs = append(a.validPairs, p)
+	return p
+}
+
+func (a *Agent) setSelectedPair(p *candidatePair) {
+	a.log.Tracef("Set selected candidate pair: %s", p)
+	// Notify when the selected pair changes
+	a.onSelectedCandidatePairChange(p)
+
+	a.selectedPair = p
+	a.updateConnectionState(ConnectionStateConnected)
 
 	// Signal connected
 	a.onConnectedOnce.Do(func() { close(a.onConnected) })
+}
+
+func (a *Agent) getBestValidPair() *candidatePair {
+	if len(a.validPairs) == 0 {
+		return nil
+	}
+	sort.Sort(byPairPriority{a.validPairs})
+	return a.validPairs[0]
 }
 
 // A task is a
@@ -383,28 +362,29 @@ func (a *Agent) run(t task) error {
 }
 
 func (a *Agent) taskLoop() {
-	contactCandidates := func() {
-		if a.validateSelectedPair() {
-			a.log.Trace("checking keepalive")
-			a.checkKeepalive()
-		} else {
-			a.log.Trace("pinging all candidates")
-			a.pingAllCandidates()
-		}
-	}
-
 	for {
-		select {
-		case <-a.forceCandidateContact:
-			contactCandidates()
-		case <-a.connectivityChan:
-			contactCandidates()
-		case t := <-a.taskChan:
-			// Run the task
-			t(a)
+		if a.selector != nil {
+			select {
+			case <-a.forceCandidateContact:
+				a.selector.ContactCandidates()
+			case <-a.connectivityChan:
+				a.selector.ContactCandidates()
+			case t := <-a.taskChan:
+				// Run the task
+				t(a)
 
-		case <-a.done:
-			return
+			case <-a.done:
+				return
+			}
+		} else {
+			select {
+			case t := <-a.taskChan:
+				// Run the task
+				t(a)
+
+			case <-a.done:
+				return
+			}
 		}
 	}
 }
@@ -449,7 +429,7 @@ func (a *Agent) pingAllCandidates() {
 
 			for _, localCandidate := range localCandidates {
 				for _, remoteCandidate := range remoteCandidates {
-					a.pingCandidate(localCandidate, remoteCandidate)
+					a.selector.PingCandidate(localCandidate, remoteCandidate)
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


agent.go
```diff
@@ -260,6 +266,25 @@ func (a *Agent) startConnectivityChecks(isControlling bool, remoteUfrag, remoteP
 	a.haveStarted.Store(true)
 	a.log.Debugf("Started agent: isControlling? %t, remoteUfrag: %q, remotePwd: %q", isControlling, remoteUfrag, remotePwd)
 
+	if isControlling {
```
Sean-Der: I would nominate the first pair that the controlling side sends a `SuccessResponse` for. 

We can come back later and figure out how to pick the 'Highest Priority' later, less chance of bugs for the first round though.

agent.go
```diff
@@ -52,6 +52,11 @@ type bindingRequest struct {
 	destination   net.Addr
 }
 
+type nominatedPair struct {
```
Sean-Der: Can we use the validPair structure here instead?

**Controlling Side**
Maintain a list of pairs that we saw a full Request/Response for. When we get a request (and are about to send a response) for a pair in this list it becomes our `selectedPair`

**Controlled Side**
validPairs is just the list of everything we did a Request+Response for, and we keep hitting them until we get a `USE-CANDIDATE`

hugoArregui: yes, we are actually using it like that, but this addional nominatedPair allow us to:

a) check the transaction without worring about the pendingTransaction structure (because we shouldn't risk to loose this tx in case of overlow)
b) In ase we make a nomination in the controlling side we need to differentiate between a normal success reponse and a response to a use candidate (in the later we select the pair)
b) In case we get a nomination in the controlled side for a pair that is not validate yet, we need to start another cycle of request/response and then answer the nomination.

I'm working in another commit, maybe it will be a bit clearer after a little refactor

selection.go
```diff
@@ -0,0 +1,267 @@
...
+
+	if bytes.Equal(s.validateNominationTxID, m.TransactionID) {
+		s.agent.sendBindingSuccess(s.nominationTxID, local, remote)
+	}
+}
+
+func (s *ControlledSelector) HandleBindingRequest(m *stun.Message, local, remote *Candidate) {
+	_, useCandidate := m.GetOneAttribute(stun.AttrUseCandidate)
+
+	if useCandidate {
```
Sean-Der: Should we allow `useCandidate` for different pairs? 

Maybe some sort of assert/check so we don't end up in the state where local/remote have different pairs.

selection.go
```diff
@@ -0,0 +1,267 @@
...
+
+		if p != nil {
+			// If the state of this pair is Succeeded, it means that the check
+			// previously sent by this pair produced a successful response and
+			// generated a valid pair (Section 7.2.5.3.2).  The agent sets the
+			// nominated flag value of the valid pair to true.
+			s.agent.setSelectedPair(p)
+			s.agent.sendBindingSuccess(m.TransactionID, local, remote)
+		} else {
+			p = newCandidatePair(local, remote, false)
```
Sean-Der: Is this to support aggressive nomination? In the normal case we shouldn't get a `success` for a pair that doesn't exist right?

selection.go
```diff
@@ -0,0 +1,267 @@
...
+			s.pingNominatedCandidate(p, m.TransactionID)
+		}
+	} else {
+		s.agent.sendBindingSuccess(m.TransactionID, local, remote)
+	}
+}
+
+func (s *ControlledSelector) pingNominatedCandidate(pair *candidatePair, nominationTxID []byte) {
+	a := s.agent
```
Sean-Der: (nit please push back if you disagree) I wouldn't put variables on the stack if you don't have to. IMO it makes things harder to grok (more LoC and just more state I need to keep track of when reading)

selection.go
hugoArregui: mmm not sure why not, there is no guarantee AFAIC the pair choosen by the controlling side had been processed by the controlled side

selection.go
```diff
@@ -0,0 +1,267 @@
...
+import (
+	"bytes"
+	"sort"
+	"net"
+	"time"
+
+	"github.com/pion/stun"
+)
+
+type nominatedPair struct {
```
Sean-Der: I would drop the dedicated `nominatedPair` structure, and maybe mark our `validPairs` as nominated (maybe?)

Thinking of your comments from earlier
```
a) check the transaction without worring about the pendingTransaction structure (because we shouldn't risk to loose this tx in case of overflow)
b) In case we make a nomination in the controlling side we need to differentiate between a normal success response and a response to a nomination (in the later we select the pair)
b) In case we get a nomination in the controlled side for a pair that is not validate yet, we need to start another cycle of request/response and then answer the nomination
```

a) I think it is ok to loose the tx, we will keep sending the binding-success
b) I would add another field to transaction something like `useCandidateResponse`, keep the code simpler
c) If we get a `useCandidate` and the pair isn't valid I think we will need to use if (if we are controlled) if we are dealing with aggressive remotes.

selection.go
Sean-Der: I don't believe so, we aren't just nominating a validPair with the best priority. We want to nominate a pair that has seen a full Request/Response done by both sides.

```
   L                        R
   -                        -
   STUN request ->             \  L's
             <- STUN response  /  check

              <- STUN request  \  R's
   STUN response ->            /  check

   STUN request + attribute -> \  L's
             <- STUN response  /  check

(Flight 0)
- Controlled does nothing
- Controlling pings all pairs (Should respond with success)

(Flight 1)
- Controlled pings every candidate pair it got a BindingRequest on (Should respond with success)
- Controlling does nothing

(Flight 2)
- Controlled is waiting for a useCandidate
- Controlling should send `useCandidate` to any pair it get a BindingRequest from (Since controlled and controlling both had a full Request/Response each way)
```

hugoArregui: but we don't know if the full request/response was done by both sides AFAIK that's why we use the attribute. Maybe I'm missing something. The controlling knows which candidates have been accepted because of the response, but doesn't know which responses have been received by the controlled (the best thing it could do is to track his own responses, but that doesn't mean it had been received by the controlled one).

This is the relevant part of the spec:

```
https://tools.ietf.org/html/rfc8445#section-7.3.1.5
   If the controlled agent receives a Binding request with the USE-
   CANDIDATE attribute set, and if the ICE agent accepts the request,
   the following action is based on the state of the pair computed in
   Section 7.3.1.4:

   o  If the state of this pair is Succeeded, it means that the check
      previously sent by this pair produced a successful response and
      generated a valid pair (Section 7.2.5.3.2).  The agent sets the
      nominated flag value of the valid pair to true.

   o  If the received Binding request triggered a new check to be
      enqueued in the triggered-check queue (Section 7.3.1.4), once the
      check is sent and if it generates a successful response, and
      generates a valid pair, the agent sets the nominated flag of the
      pair to true.  If the request fails (Section 7.2.5.2), the agent
      MUST remove the candidate pair from the valid list, set the
      candidate pair state to Failed, and set the checklist state to
      Failed.
```

Sean-Der: Aye that is all right, thanks for pulling the RFC keep trucking @hugoArregui 

I am so excited to land this, so many bad connections were probably caused by this bad ICE implementation :(

hugoArregui: I'll put the second item in the code (the first one is already there), I love when the comments point to the RFC, it makes it so easy to follow!

selection.go
hugoArregui: no problem! done

selection.go
hugoArregui: Can you expand on your first point? This is something I'm not sure how to do, we need to *always* reply to a nomination, but I'm not sure how to do that. It seems to me we cannot loose that tx and we will need to reply even if there if is not a valid pair and we need to perform a new request/response cycle

hugoArregui: https://tools.ietf.org/html/rfc8445#section-7.3.1.5

If the controlled agent does not accept the request from the
   controlling agent, the controlled agent MUST reject the nomination
   request with an appropriate error code response (e.g., 400)
   [RFC5389].

Sean-Der: We need to call `nominatePair` in a loop, since the connection may be lossy our original might get lost.

I would entire failure by just having a timer that records how long we have been in checking (if we are stuck in checking for more then 30 seconds we should go to fail)

I would update `taskLoop` to call the `ControllingSelector`, with logic like this
* If we have a `selectedPair` ping it
* If we have a CandidatePair that is nominated sending a Request with `useCandidate`
* If we don't have any nominated CandidatePairs ping everyone

----

To distinguish if the response is for a `useCandidate` or an old transaction I would update like this
```
type bindingRequest struct {
    transactionID []byte
    destination   net.Addr
    isUseCandidate bool
}
```

and if we get a response back for `isUseCandidate == true` this becomes our selected pair (on the controlling side)

hugoArregui: I did this change, however, I feel a bit unsure about this part of the standard

```
-       // If the controlled agent does not accept the request from the
-       // controlling agent, the controlled agent MUST reject the nomination
-       // request with an appropriate error code response (e.g., 400)
-       // [RFC5389].
```

in pion<->pion this would work because of the general timeout, but not sure when interacting with other implementations

selection.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+type ControllingSelector struct {
+	agent         *Agent
+	nominatedPair *candidatePair
+	log           logging.LeveledLogger
+}
+
+func (s *ControllingSelector) Start() {
+	go func() {
```
Sean-Der: Nice, this all looks really good!

I think the last thing is we can drop this part (now that `nominatePair` is called in `ContactCandidates`) and then move the `ConnectionStateFailed` into a timer that gets reset when `updateConnectionState` == `Checking` and if we are stuck in `Checking` for more then n seconds we go to failed?

hugoArregui: Sorry the code was a bit confusing, I updated it. We still use this to actually select the nominatedPair.

So we need two things:

- a way to nominate a pair, I'm using a timeout in here, but we could use another criteria. I know you suggested to nominate the first pair we get, but I don't really like that too much.. I think this way (which is not ideal either) at least has the advantage of selecting by priority.
- the timeout you mentioned. We are already doing that in here, we could check if we are stuck in checking instead, but not sure if that's is actually better, just because there are other reasons to be stuck on a checking state (like if we don't get an aswer from the signaling system), I feel a timeout to wait for candidates may be easier to understand from the point of view of the user of the library.

selection.go
```diff
@@ -85,6 +68,12 @@ func (s *controllingSelector) nominatePair(pair *candidatePair) {
 
 func (s *controllingSelector) HandleBindingRequest(m *stun.Message, local, remote *Candidate) {
 	s.agent.sendBindingSuccess(m, local, remote)
+
+	p := s.agent.findValidPair(local, remote)
+	if p != nil && s.nominatedPair == nil && s.agent.selectedPair == nil {
```
Sean-Der: @hugoArregui Feel free to revert, but I think this is the best thing to do for now (and we can keep the change small/test it well)

We can follow up and make the timeout configurable (and allow the user to configure it!) but for now choosing the first candidate is best (IMO) connection startup time is really important IMO for the default.

Sean-Der:
@hugoArregui This is amazing, I used it locally and it works great! This is going to make a huge difference for a lot of people :)

If you don't mind
* Squashing everything down (just so we have a single commit/clean history) if you want to split into multiple clean commits that is fine also! Just want to avoid commits that aren't useful in the history.

* Finish up testing (I can grab this also!) I played around with it locally `Pion <-> Pion` and `Chrome <-> Pion` both worked great so that is fantastic :D 

* Push and enjoy your status as ICE hero!

selection.go
```diff
@@ -107,6 +96,7 @@ func (s *controllingSelector) HandleSucessResponse(m *stun.Message, local, remot
 	p := s.agent.addValidPair(local, remote)
 
 	if pendingRequest.isUseCandidate {
+		s.nominatedPair = nil
```
hugoArregui: why setting it to nil? Remember we cannot nominate two different pairs

selection.go
hugoArregui: ok, that makes sense. My main concern is, it seems to me (but maybe this won't happen in reallity), that we could end up with a relay candidate instead of a host or a reflexive candidate. That will be bad because we have to pay for the use of the TURN server :joy: 

Otherwise I'm ok with selecting the first pair, but it seems to be we need to reintroduce the timeout right>

selection.go
Sean-Der: If the`validPair` times out we might need to nominate a different pair. We can move this logic/setting to `validateSelectedPair` instead though!

selection.go
Sean-Der: Yea I agree with everything here! My vote is a follow up PR (people are hitting this bug today, so want to get it out)

In that PR I think we can implement
* Setting the connection to failed after `n` time in checking
* Allow setting the `Nomination Policy` and we will collect `validPairs` for that long. However if we get a pairing with the lowest possible priority (`host <-> host`) we should take it right away!
* This `Nomination Policy` can be configured via `webrtc.SettingEngine`
* I would be curious to know what Chromium's time for this is. It feels very fast usually.

selection.go
hugoArregui: I don't think so, from the RFC:

```
https://tools.ietf.org/html/rfc8445#section-8.1.1
  The criteria for stopping the connectivity checks and for picking a
   pair for nomination are outside the scope of this specification.
   They are a matter of local optimization.  The only requirement is
   that the agent MUST eventually pick one and only one candidate pair
   and generate a check for that pair with the USE-CANDIDATE attribute
   set.

   Once the controlling agent has successfully nominated a candidate
   pair (Section 7.2.5.3.4), the agent MUST NOT nominate another pair
   for same component of the data stream within the ICE session.  Doing
   so requires an ICE restart.
```

Sean-Der: Aye nice check, mind killing that line?

estasig:
"Next we can land configurable Nomination Policy setting the time to wait etc..."
---
Hi Sean-Der
Do we have the configurable Nomination Policy setting now? (not found in the doc.)
I'm testing Pion with Chrome today. Seems chrome is using aggressive nomination and the connection breaks about 30 seconds after chrome chooses another pair.
Both M77 and M76 have the problem.

Sean-Der:
@estasig That doesn't sound good! It sounds like Chrome is re-nominating maybe? Which we don't support yet, can you share a SDP? It shouldn't try to switch nominated/selected pairs once it makes a decision.

Is something accidentally adding it to the SDP? Maybe try only giving Chrome one candidate for now (that you know is valid) and see if it stays connected.

estasig:
```
v=0
o=- 624190682 1569291915 IN IP4 0.0.0.0
s=-
t=0 0
a=fingerprint:sha-256 62:18:D0:DC:7C:92:94:33:2E:2C:67:1B:08:9C:56:F2:A3:C9:96:3C:86:7C:46:C5:52:68:3E:09:D6:39:38:79
a=group:BUNDLE 0 1
m=audio 9 UDP/TLS/RTP/SAVPF 0 8 111
c=IN IP4 0.0.0.0
a=setup:active
a=mid:0
a=ice-ufrag:hMuWFSvrvuStGQIF
a=ice-pwd:GYSaonWMIhKicfoqytgQPzptgqrxhvcl
a=rtcp-mux
a=rtcp-rsize
a=rtpmap:0 PCMU/8000/1
a=rtpmap:8 PCMA/8000/1
a=rtpmap:111 opus/48000/2
a=fmtp:111 minptime=10;useinbandfec=1
a=ssrc:920256325 cname:abc
a=ssrc:920256325 msid:abc audio
a=ssrc:920256325 mslabel:abc
a=ssrc:920256325 label:audio
a=msid:abc audio
a=sendrecv
a=candidate:foundation 1 udp 2130706431 x.x.x.x 42974 typ host generation 0
a=candidate:foundation 2 udp 2130706431 x.x.x.x 42974 typ host generation 0
a=candidate:foundation 1 udp 2130706431 x.x.x.x 40572 typ host generation 0
a=candidate:foundation 2 udp 2130706431 x.x.x.x 40572 typ host generation 0
a=candidate:foundation 1 udp 1694498815 x.x.x.x 50131 typ srflx raddr 0.0.0.0 rport 50131 generation 0
a=candidate:foundation 2 udp 1694498815 x.x.x.x 50131 typ srflx raddr 0.0.0.0 rport 50131 generation 0
a=candidate:foundation 1 udp 1694498815 x.x.x.x 44656 typ srflx raddr 0.0.0.0 rport 44656 generation 0
a=candidate:foundation 2 udp 1694498815 x.x.x.x 44656 typ srflx raddr 0.0.0.0 rport 44656 generation 0
a=end-of-candidates
m=video 9 UDP/TLS/RTP/SAVPF 102
c=IN IP4 0.0.0.0
a=setup:active
a=mid:1
a=ice-ufrag:hMuWFSvrvuStGQIF
a=ice-pwd:GYSaonWMIhKicfoqytgQPzptgqrxhvcl
a=rtcp-mux
a=rtcp-rsize
a=rtpmap:102 H264/90000
a=fmtp:102 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42001f
a=ssrc:1634910179 cname:abc
a=ssrc:1634910179 msid:abc video
a=ssrc:1634910179 mslabel:abc
a=ssrc:1634910179 label:video
a=msid:abc video
a=sendrecv
a=candidate:foundation 1 udp 2130706431 x.x.x.x 42974 typ host generation 0
a=candidate:foundation 2 udp 2130706431 x.x.x.x 42974 typ host generation 0
a=candidate:foundation 1 udp 2130706431 x.x.x.x 40572 typ host generation 0
a=candidate:foundation 2 udp 2130706431 x.x.x.x 40572 typ host generation 0
a=candidate:foundation 1 udp 1694498815 x.x.x.x 50131 typ srflx raddr 0.0.0.0 rport 50131 generation 0
a=candidate:foundation 2 udp 1694498815 x.x.x.x 50131 typ srflx raddr 0.0.0.0 rport 50131 generation 0
a=candidate:foundation 1 udp 1694498815 x.x.x.x 44656 typ srflx raddr 0.0.0.0 rport 44656 generation 0
a=candidate:foundation 2 udp 1694498815 x.x.x.x 44656 typ srflx raddr 0.0.0.0 rport 44656 generation 0
a=end-of-candidates
```

This is the answer SDP sent by Pion. The IP address is hiden.
Why there are two components while the offer SDP from Chrome has one.

PR #37, opened 2019-05-04
Select a pair once

hugoArregui:
Although we only nominate and select one candidate pair, since we retry the request we can end up selecting the same candidate multiple times


selection.go
```diff
@@ -200,7 +200,9 @@ func (s *controlledSelector) HandleBindingRequest(m *stun.Message, local, remote
 			// previously sent by this pair produced a successful response and
 			// generated a valid pair (Section 7.2.5.3.2).  The agent sets the
 			// nominated flag value of the valid pair to true.
-			s.agent.setSelectedPair(p)
+			if s.agent.selectedPair != nil {
```
Sean-Der: `if s.agent.selectedPair == nil {` I think

LGTM beyond that!

hugoArregui: thanks!, fixed now

PR #41, opened 2019-05-09
Respect port range for srv reflexive candidates

masterada:
Modified gatherCandidatesReflective to respect port min and max
constraints when allocating connections for server reflexive
candidates.

#### Reference issue
Related to https://github.com/pion/webrtc/issues/668

#### Depends on
https://github.com/pion/stun/pull/9

### DIFF
--- a/stun.go
+++ b/stun.go
@@ -3,10 +3,7 @@ package ice
 import (
 	"bytes"
 	"encoding/binary"
-	"errors"
 	"fmt"
-	"net"
-	"time"
 
 	"github.com/pion/stun"
 )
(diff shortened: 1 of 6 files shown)


stun.go
```diff
@@ -67,35 +103,45 @@ func assertInboundMessageIntegrity(m *stun.Message, key []byte) error {
...
+	}
+	err = conn.SetWriteDeadline(time.Now().Add(deadline))
 	if err != nil {
-		return nil, nil, flattenErrs([]error{errors.New("failed to make STUN request"), err})
+		return nil, err
 	}
 
-	if err = client.Close(); err != nil {
-		return nil, nil, flattenErrs([]error{errors.New("failed to close STUN client"), err})
+	resp, err := stun.Request(&stunConn{conn, stunAddr})
```
masterada: Maybe instead of stun.Request i could add a GetMappedAddress method that does everything that request does, but also reads and returns the XorAddress. What do you think?

Sean-Der: Yea 👍 from me!

We should try and keep `pion/ice` as small as possible, it is about to get much more bloated with TURN coming :/

PR #42, opened 2019-05-14
Adding checklist for candidate validation

### LINKED ISSUE #35: ICE connection state stuck at 'checking' forever
## Summary
When there's no connectivity between two endpoints, both ends keep sending pings forever (never fails).

## Motivation
This forces a developer to implement application-level timeout.

(see Motivation)

## Additional context
RFC 5389 section 7.2.1.  Sending over UDP
```
   Retransmissions continue until a response is received, or until a
   total of Rc requests have been sent.  Rc SHOULD be configurable and
   SHOULD have a default of 7.  If, after the last request, a duration
   equal to Rm times the RTO has passed without a response (providing
   ample time to get a response if only this final request actually
   succeeds), the client SHOULD consider the transaction to have failed.
```

hugoArregui:
Fixes #35 

This implements a basic validation schema using a checklist. We try every pair at least maxTries (7 by default see #35), and mark it as failed if we don't get a success response after that many requests. Once we get a success response, we check if it belongs to the best candidate available so far, if it does we nominate it, otherwise we continue.

Also, after a given timeout, if no candidate has been nominated, we simply choose the best valid candidate we got so far (if no candidate is valid, we mark the connection as failed).

Finally, the nomination request also has a maximum of maxTries, we mark the connection as failed if after that many attempt we fail to get a success response.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
-	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -26,6 +25,24 @@ const (
 	// defaultConnectionTimeout used to declare a connection dead
 	defaultConnectionTimeout = 30 * time.Second
 
+	// timeout for candidate selection, after this time, the best candidate is used
+	defaultCandidateSelectionTimeout = 10 * time.Second
+
+	// wait time before nominating a host candidate
+	defaultHostAcceptanceMinWait = 0
+
+	// wait time before nominating a srflx candidate
+	defaultSrflxAcceptanceMinWait = 500 * time.Millisecond
+
+	// wait time before nominating a prflx candidate
+	defaultPrflxAcceptanceMinWait = 1000 * time.Millisecond
+
+	// wait time before nominating a relay candidate
+	defaultRelayAcceptanceMinWait = 2000 * time.Millisecond
+
+	// max binding request before considering a pair failed
+	defaultMaxBindingRequests = 7
+
 	// the number of bytes that can be buffered before we start to error
 	maxBufferSize = 1000 * 1000 // 1MB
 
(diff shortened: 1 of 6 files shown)


enobufs:
From a quick review, it looks good to me. I guess this PR does not address adding a new prflx "local" candidate. Correct? (I guess that would be the next step after this)
One test (TestTimeout) is failing, btw.

hugoArregui:
@enobufs :thinking: not sure what you mean by local candidate

enobufs:
It's related to things we talked about earlier around "we should check (xor-)mapped address in the binding-response message".. so that we can have a complete list of local candidates that would include prflx candidate also.

Say if you have this srflx candidate as your local candidate.
```
Candidates [
{
Type: srflx
Local transport addr: 1.2.3.4:6000
Related Addr: 192.1.1.2:5000
}]
```
Later, you may find a new prflx candidate from the binding-response:
```
Candidates [
{
Type: srflx
Local transport addr: 1.2.3.4:6000
Related Addr: 192.1.1.2:5000
}, {
Type: prflx
Local transport addr: 1.2.3.4:6001
Related Addr: 192.1.1.2:5000
}]

I believe do this does not affect "connectivity", but it may help later in selection logic (I think prflx is preferred more than srflx, etc) and diagnostic purpose (you can tell which mapped-addr is actually being used, etc)

**Comment by hugoArregui** [comment]
@enobufs oh yes, you are right: https://github.com/pion/ice/blob/master/candidatetype.go#L35 didn't realise that, I thought it was just a useful debugging property.

**Comment by enobufs** [comment]
TestTimeout is failing on my mac...(rev @ae901)
```
=== RUN   TestTimeout
ice ERROR: 2019/05/21 15:47:47 error processing checkCandidatesTimeout handler the agent is closed
ice ERROR: 2019/05/21 15:47:47 error processing checkCandidatesTimeout handler the agent is closed
--- FAIL: TestTimeout (30.00s)
    transport_test.go:43: Connection timed out early. (after 29900 ms)
```

**Comment by hugoArregui** [comment]
@enobufs yes, that's a known issue, also happens with master. Fails on mac and not on linux, not sure why (I never actually checked)

**Comment by enobufs** [comment]
I see. I don't know exactly what the testing is testing, but as it says, `Connection timed out early. (after 29900 ms)` - it is trying validate that the timeout shouldn't occur too early. The error says, it at least did not timeout in 29.9s - pretty close!  Maybe the timeout threshold in the test code is the problem? (just guessing...)

**Comment by enobufs** [comment]
> I see. I don't know exactly what the testing is testing, but as it says, `Connection timed out early. (after 29900 ms)` - I'd assume it is trying validate that the timeout shouldn't occur too early. The error says, it at least did not timeout in 29.9s - pretty close! Maybe the timeout threshold in the test code is the problem? (just guessing...)

This has been addressed in #49 !

File: agent.go [hunk 1]
```diff
@@ -215,6 +228,41 @@ func NewAgent(config *AgentConfig) (*Agent, error) {
 	}
 	a.haveStarted.Store(false)
 
+	a.maxBindingRequests = config.MaxBindingRequests
```
Sean-Der: Fantastic work!

Would you mind moving the defaultValues into `const` values?

Sean-Der:
Amazing work @hugoArregui !

This is an approve from me, would you mind just moving the 'default values' into constants. Just easier for people to audit/twiddle that way.

Beyond that I am gonna go home and play with this a little bit, but it is a approve from me :)

agent.go
hugoArregui: done!

PR #46, opened 2019-05-22
Add vnet support

hugoArregui:
Add vnet support by taking a vnet.Net in the AgentConfig

Requires https://github.com/pion/transport/pull/14


Sean-Der:
Exciting :D

One thing that might cause us issues is `vnet` being in production code paths? 

When debugging it might be nice if we can completely keep `vnet` out of peoples code paths, it would be nice if you could pass a `listenUDP` override. Just so the only people that actually even know about `vnet` is us.

hugoArregui:
@Sean-Der not sure to understand what you are talking about, being in production code paths in the library or in the user code?. If you are talking about the later, you can leave net out of the config and by default it will use the normal networking. If you are talking about library code paths.. well I think that's the vnet approach, we may need to change that

ernado:
Can we abstract networking entirely?
E.g. define some `Net` interface and implement it for `system` (default) and for `vnet`?

hugoArregui:
@Sean-Der @ernado I think the issues you are mentioning are more related to vnet than to this PR, do you mind bringing them up in https://github.com/pion/transport/pull/14 ?

enobufs:
> One thing that might cause us issues is vnet being in production code paths?

For production code path, it will traverse only one `if` statement in the current design - it is already a small price to pay for ensuring the quality of ICE to my knowledge. Also, you can introduce vnet selectivity gradually.  For instance, we could start just with unit pion/ice (enabling virtual network only in its unit tests, etc). We will still need a virtual STUN server. If we don't want to touch pion/stun for any doubt, I can write a dead simple STUN server good enough for testing ICE logic.

ernado:
I'm talking about following:
```go
type Net interface {
    ListenPacket(net, addr string) (net.PacketConn, error)
//...
```

And implement that interface for vnet and for go standard net as different network providers.

enobufs:
That was my initial idea, but there are some problems I had:
* Types / properties cannot be exposed through the interface.
* net.ListenUDP() for instance returns a struct (not interface) and there are associated methods to it. That means, the approach cannot be a drop-in replacement. (net package uses interfaces for conn, addr which is good and we can take advantage of them, but not fully "virtualized" unfortunately...)
* We'd also like to avoid introducing global variables.

agent.go
```diff
@@ -154,6 +156,9 @@ type AgentConfig struct {
 
 	LoggerFactory logging.LoggerFactory
 
+	// Net is the our abstracted network interface (see github.com/pion/transport/vnet)
```
enobufs: We should just say "Internal development purpose only." or something for now.

ernado:
> Types / properties cannot be exposed through the interface.

What is the use case?

> net.ListenUDP() for instance returns a struct (not interface) and there are associated methods to it. That means, the approach cannot be a drop-in replacement. (net package uses interfaces for conn, addr which is good and we can take advantage of them, but not fully "virtualized" unfortunately...)

Right, so the general refactoring is needed to make it work.
Anyway we can't make a drop-in replacement, because `net.ListenUDP` is package-level function.

The main point is to access everything via interface so we can really mock/abstract network. 

I'll take another look on this topic soon.

> We'd also like to avoid introducing global variables.

Agree.

Sean-Der:
@hugoArregui @enobufs 

This is a 👍 from me! Since we don't introduce any risk for users (and still enable testing) I would love to roll ahead with this :)

agent.go
```diff
@@ -242,11 +249,16 @@ func NewAgent(config *AgentConfig) (*Agent, error) {
...
 		log:         loggerFactory.NewLogger("ice"),
+		net:         config.Net,
 
 		forceCandidateContact: make(chan bool, 1),
 	}
 	a.haveStarted.Store(false)
 
+	if a.net == nil {
+		a.net = vnet.NewNet(nil)
+	}
```
enobufs: Let's add a warning log if a.net is not nil.
See: https://github.com/pion/transport/wiki/vnet:-Usage-Policy#log-warning-from-each-pion-module-if-vnet-is-enabled

enobufs:
@hugo Question.   I believe pion/ice uses net.Interfaces() to gather local candidates. Are you planning to convert it to vnet in another PR?

Sean-Der:
https://github.com/pion/ice/blob/7e5b1e89c48a4353e52b98b8c04293f3c3a39d5f/gather.go#L19 is the line @enobufs is talking about

@hugoArregui @enobufs maybe we can add a lint for `net` import? That way when new devs add code we don't have to worry about making sure they miss something manually.

hugoArregui:
> @hugoArregui @enobufs maybe we can add a lint for `net` import? That way when new devs add code we don't have to worry about making sure they miss something manually.

To do so we will need to add type aliases in vnet for the common net things like IP

enobufs:
> > @hugoArregui @enobufs maybe we can add a lint for `net` import? That way when new devs add code we don't have to worry about making sure they miss something manually.
> 
> To do so we will need to add type aliases in vnet for the common net things like IP

enobufs:
(oops I hit a wrong button)
Adding linting to disallow using particular function of net package may help.
So, can it detect the following?

* GOOD: `.+`.net.ListenPacket*
* BAD: `\s+`net.ListenPacket* => lint error

agent.go
hugoArregui: done!

hugoArregui:
Sorry I forgot about Interfaces! it's done now!

enobufs:
To offload @hugoArregui's work, I offered to take over this PR to finish up.
@Sean-Der, if you are ok, I will merge.

enobufs:
Continuing on tracking srflx-relay candidate issue.
So far I have found that the local relay candidate receives a packet from a new remote transport addr (should generate a prflx), but it does not send a ping to the new prflx candidate.
  
Screen Shot 2019-07-06 at 4 13 24 PM

enobufs:
I found another problem in pion/turnc. It is matching remote address with PeerAddress (defined in gortc/turn) which includes port number.

https://github.com/pion/turnc/blob/v0.0.6/client.go#L149

"Permission" should be per protocol+IP address only as RFC suggests. Otherwise, it can not find a valid pair when the other end has a symmetric NAT.

From: https://tools.ietf.org/html/rfc5766#section-2.3
```
To ease concerns amongst enterprise IT administrators that TURN could be used to bypass corporate firewall security, TURN includes the notion of permissions. TURN permissions mimic the address-restricted filtering mechanism of NATs that comply with [RFC4787].
```

enobufs:
@ernado, @Sean-Der  I am a bit overwhelmed by the complexity in pion/turnc and not sure how to solve this problem.  If you could give me a suggestion, that would be great.

As I mentioned in my previous message, I'd like to receive a STUN (connectivity check) message from a remote. I have already fixed the pion/turn server (at pion/turn@663b154) so that the permission is made just for IP (ignores port number) as RFC suggests.

The pion/turnc tho, the permission object and remote transport address has a 1:1 relationship, where it nees to have 1:n relationship. For instance, remote addresss 28.1.1.1:1234 and 28.1.1.1:5678 may share the "permission".

enobufs:
At this point, I am pretty sure our implementation of current pion/turnc is incorrect.
I am going to try to apply these changes:
* Rename current "Permission" to "Channel"
* Introduce a new permission map (map[string]*permission)
    - key is <protocol>:<peer-ip-address>, e.g. "udp:28.1.1.1"
    - the map is referenced by all channels to check permissions

I will file an issue in pion/turn once details of my proposal become clear.

Sean-Der:
@enobufs 

If you think it would be less work to rewrite, then refactor I am totally in support! In ~2 days I will have a better schedule and we talk next steps. sorry about the delay!

enobufs:
> @enobufs
> 
> If you think it would be less work to rewrite, then refactor I am totally in support! In ~2 days I will have a better schedule and we talk next steps. sorry about the delay!

Thanks!  Current pion/turnc also uses net.DialUDP/net.Conn instead of net.ListenPacket/net.PacketConn, which makes it impossible to use the existing UDPConn.  We will need to make a major change for that too.  Let me refactor in pion/turnc#11.

Sean-Der:
Nice! Yea whatever you think is best @enobufs 

IMO we need to have ownership over each individual part, if @ernado is up for fixing this stuff that is the best case scenario! Otherwise we probably need to rewrite/rework things so we know it first hand.

I think having no dependencies/everything under `pion/` has been one of the best decisions we have made so far. Fantastic how fast we can move with that :)

enobufs:
@Sean-Der @hugoArregui 
Good news! I have integrated the new pion/turn (refactor/client branch) today.
Now I got successful relay-srflx candidate pair!!
```
ice TRACE: 23:46:41.632977 agent.go:474: Set selected candidate pair: prio 72057593450725376 (local, prio 16777215) relay 1.2.3.4:5495 related 0.0.0.0:5237 <-> prflx 28.1.1.1:49157 related :0 (remote, prio 1862270975)
```

I have to solve one race error from vnet - but that would be it. 🤞 
CandidateRelay with the new TURN client looks very simple. I was able to use the default candidateBase.start()/.recvLoop()/ as-is. You'd love it!!

Almost there... :)

enobufs:
I guess it doesn't matter anymore because we broke API last time too. Let's sort this thing out when v2.1.0 rolls out...  until then keep pion/turn version as 1.x (but treat it as 0.x in semver sense).

Issue #47, opened 2019-05-23
Implement TURN


Sean-Der:
This is a covering ticket for all the things we need to do.

### `pion/webrtc`
* [x] Pass TURN configuration from `pion/webrtc` to `pion/ice`
* [x] Add support for `ICETransportPolicy` so we can force TURN

---

### `pion/ice`
* [x] TURN MVP. Add basic TURN support via `pion/turnc`
* [x] Update checklist/nomination logic so TURN is last (currently we just take first success)
* [ ] Add TURN server to `travis-ci`, implement a version and confirm it can't regress

hugoArregui:
#42 may already solve the 3rd point

Sean-Der:
@hugoArregui I am going to work on getting TURN/STUN tested in travis-ci. This is probably going to take me 2-3 days.

Would you be up for getting your checklist stuff re-merged? Sorry for all the breakage, I would love to work together closely and get both your changes in.

Sean-Der:
@enobufs Any other next steps you can think of?  I feel pretty good about where we are at TURN wise!

enobufs:
@Sean-Der @ernado @hugoArregui MANY Thanks to all of you for making this happen! I am sure many pion users will be very happy about the TURN support! I will be updating my app and test the new feature. My time this week is very limited because I have to move into my new apartment this weekend tho, I should be able to get some feedback in the week of June 2)

I guess the next step is to further improve ICE including @hugoArregui was planning to do (checklist stuff). Now we have relay candidate, a proper priority control in candidate nomination (and quality assurance) would be desired. Now we have another expert @ernado - awesome! I'm very hopeful that pion would yield very competitive connectivity over other webrtc libraries!

hugoArregui:
@Sean-Der sure! Let's ensure we are on the same page regarding the implementation though, the thing about the heuristic I wrote in the PR is now we have trickle.. so I'm not sure if it holds.. maybe I can add a "mininumCheckWait" (sorry, awful name) to wait a bit for candidates?

PR #48, opened 2019-05-24
Make Candidate an interface

Sean-Der:
This change will allow us to have custom logic and members
per interface type. Relay candidates will have a completely different
read loop, and candidate specific state.

Relates to #47

### DIFF
--- a/candidate.go
+++ b/candidate.go
@@ -1,12 +1,8 @@
 package ice
 
 import (
-	"fmt"
 	"net"
-	"sync"
 	"time"
-
-	"github.com/pion/stun"
 )
 
 const (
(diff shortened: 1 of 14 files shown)


Sean-Der:
This diff is mostly me moving files around. I would say the big thing I would like to know is if people agree with how this is implemented.

Most candidates share all the same logic, but TURN needs to have its own `readLoop` and will have state not shared by other candidates.

candidate.go
```diff
@@ -20,260 +16,26 @@ const (
...
-func (c *Candidate) addr() net.Addr {
-	return &net.UDPAddr{
-		IP:   c.IP,
-		Port: c.Port,
-	}
+type Candidate interface {
+	start(a *Agent, conn net.PacketConn)
+	addr() net.Addr
+
+	setLastSent(t time.Time)
```
hugoArregui: probably you can remove this from the interface

candidate.go
```diff
@@ -20,260 +16,26 @@ const (
...
+	seen(outbound bool)
+	LastSent() time.Time
+	setLastReceived(t time.Time)
```
hugoArregui: probably you can remove this from the interface

hugoArregui:
I have no strong opinion about this change mainly because I don't know what's involved for the relay candidates. If the only change is the readLoop and the state maybe you could consider abstracting that out? (something like the thing we did for the agent selector)

Sean-Der:
@hugoArregui thanks for the review (speaking of ICE reviews we need to land all your work!)

As long as you think the code is still pleasant  to work with I would like to keep going with it. The relay change will involve

* Custom read behavior
* Custom write behavior
* Custom behavior when creating new candidates (for each candidate of other type we need to create a new permission on our TURN server)

candidate.go
Sean-Der: Fixed!

candidate.go
Sean-Der: Fixed!

enobufs:
Later might be better than never...  I support making Candidate an interface.

PR #52, opened 2019-05-25
WIP Turn Implementation

Sean-Der:
* This disables all other candidates besides relay
* We hardcode networkTypes in a few places

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -33,6 +33,10 @@ const (
 	maxPendingBindingRequests = 50
 )
 
+var (
+	defaultCandidateTypes = []CandidateType{CandidateTypeHost, CandidateTypeServerReflexive, CandidateTypeRelay}
+)
+
 type candidatePairs []*candidatePair
 
 func (cp candidatePairs) Len() int      { return len(cp) }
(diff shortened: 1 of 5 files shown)


enobufs:
@Sean-Der TURN implementation so far look good to me although there are few things don't entirely sure about your intentions / plans:
* CandidateRelay.start() vs CandidateBase.start(): how would base's readLoop will be kicked?
* Test strategy for TURN: (I'd wish we already had vnet now!!)

I wonder if we utilize some public TURN servers at least for a quick testing...
See: https://gist.github.com/sagivo/3a4b2f2c7ac6e1b5267c2f1f59ac6c6b
You could test those servers with this browser tool: https://webrtc.github.io/samples/src/content/peerconnection/trickle-ice/
This one seems to be working...:
```
{
    url: 'turn:numb.viagenie.ca',
    credential: 'muazkh',
    username: 'webrtc@live.com'
},
```

Sean-Der:
@enobufs Sorry on mobile!

* Relay has to do a readLoop per remote candidate. So we cant start a single loop (like base) I forgot the function name but it starts a read in a routine 

* for testing I am going to use pion/turn. I am going to do the basics in this PR (after Trickle lands) but then do full integration testing in a follow up PR. I can do STUN also (will disable host candidate gathering via AllowedCandidates)

Sean-Der:
@hugoArregui @enobufs ok this is up for merge!

If neither of you see anything awful I would like to merge this (and then add integration testing for TURN/STUN) in another commit. I want to get this in mainline so @enobufs can be unblocked :)

PR #53, opened 2019-05-25
Add trickle support

### LINKED ISSUE #51: Slowdown because of server reflexive address allocation
### What happened?
Between v0.2.7 and v0.2.8 there was a change in the way server reflexive addresses are allocated. Rather than collecting only one (default) address, the gathering process now actually goes through all of the local interfaces which takes a significant load time.

Here is the part that causes this:
https://github.com/pion/ice/blob/v0.2.8/gather.go#L131

### Proposal
Create an asynchronous process that dispatches a goroutine to collect the server reflexive addresses but only waits for one (default one) to come back before continuing with the execution.

### Requirements

- [x] OnIceCandidate must be called when new candidate is discovered
- [ ] there must be a method for adding new remote candidates (not sure of the method name)
- [x] there should be an option for disabling trickle alltogether and waiting for all candidates

trivigy:
#### Reference issue
Resolves #51

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -91,24 +91,20 @@ func listenUDP(portMax, portMin int, network string, laddr *net.UDPAddr) (*net.U
 }
 
 // GatherCandidates initiates the trickle based gathering process.
-func (a *Agent) GatherCandidates(urls []*URL, networkTypes []NetworkType) error {
-	return a.run(func(agent *Agent) {
+func (a *Agent) GatherCandidates() error {
 		if a.gatheringState == GatheringStateGathering {
 			a.log.Warnf("Attempting to gather candidates during gathering state\n")
 			return
 		}
 
-		go a.gatherCandidates(&AgentConfig{
-			Urls:         urls,
-			NetworkTypes: networkTypes,
-		})
+		go a.gatherCandidates()
 	})
 }
 
-func (a *Agent) gatherCandidates(config *AgentConfig) {
+func (a *Agent) gatherCandidates() {
 	a.gatheringState = GatheringStateGathering
-	a.gatherCandidatesLocal(config.NetworkTypes)
-	a.gatherCandidatesSrflx(config.Urls, config.NetworkTypes)
+	a.gatherCandidatesLocal(a.networkTypes)
+	a.gatherCandidatesSrflx(a.urls, a.networkTypes)
 	if err := a.run(func(agent *Agent) {
 		if a.onCandidateHdlr != nil {
 			go a.onCandidateHdlr(nil)
--- a/gather.go
+++ b/gather.go
@@ -92,17 +92,38 @@ func listenUDP(portMax, portMin int, network string, laddr *net.UDPAddr) (*net.U
 
 // GatherCandidates initiates the trickle based gathering process.
 func (a *Agent) GatherCandidates() error {
+	gatherErrChan := make(chan error, 1)
+
+	runErr := a.run(func(agent *Agent) {
 		if a.gatheringState == GatheringStateGathering {
-			a.log.Warnf("Attempting to gather candidates during gathering state\n")
+			gatherErrChan <- ErrMultipleGatherAttempted
+			return
+		} else if a.onCandidateHdlr == nil {
+			gatherErrChan <- ErrNoOnCandidateHandler
 			return
 		}
 
 		go a.gatherCandidates()
+
+		gatherErrChan <- nil
 	})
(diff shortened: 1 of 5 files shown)


gather.go
```diff
@@ -112,11 +125,17 @@ func gatherCandidatesLocal(a *Agent, networkTypes []NetworkType) {
 			a.localCandidates[c.NetworkType()] = set
 
 			c.start(a, conn)
+
+			a.mutex.Lock()
```
Sean-Der: Is it possible to use `a.run` instead of taking the lock?

gather.go
```diff
@@ -90,7 +90,38 @@ func listenUDP(portMax, portMin int, network string, laddr *net.UDPAddr) (*net.U
...
+
+		go a.gatherCandidates(&AgentConfig{
+			Urls:         urls,
+			NetworkTypes: networkTypes,
+		})
+	})
+}
+
+func (a *Agent) gatherCandidates(config *AgentConfig) {
+	a.gatheringState = GatheringStateGathering
```
Sean-Der: I think we need to lock while setting this.

IMO I think it would be nice to have this as an atomic. Just add a getter/setter and then we don't need to worry :)

Sean-Der:
@trivigy Amazing, I love this! So exciting this is happening :) @hugoArregui or @masterada minding doing one more over. ICE is so sensitive to changes, always effects `webrtc` in weird ways so worth another look.

* Has a toggle (no user impact)
* Has tests
* Doesn't break the API

Doesn't get better then that :)

gather.go
```diff
@@ -90,7 +90,38 @@ func listenUDP(portMax, portMin int, network string, laddr *net.UDPAddr) (*net.U
 	return nil, ErrPort
 }
 
-func gatherCandidatesLocal(a *Agent, networkTypes []NetworkType) {
+// GatherCandidates initiates the trickle based gathering process.
+func (a *Agent) GatherCandidates(urls []*URL, networkTypes []NetworkType) error {
```
Sean-Der: Should we do our check return `ErrNoOnCandidateHandler` here?

trivigy: The issue here is that this stuff is happening behind this `a.run` task which happens synchronously but by a different goroutine. I originally added ErrNoOnCandidateHandler and was returning it here but because I removed the locks, and put `a.run` I am not sure how to do it anymore.

gather.go
trivigy: I can close this stuff behind `a.run`. You good with that?

Sean-Der: Yea that is good with me!

gather.go
Sean-Der: https://github.com/pion/ice/blob/master/candidate_base.go#L109

That is the best we can do for now. I think we want to drop the `a.run` pattern though, someone did a profile and told me it was a major hot spot for them.

masterada:
@trivigy looks great, 3 little things:
- Please add a method AddIceCandidate so we can support the other half of trickle as well (when the remote peer sends new candidates). It should simply add to the list of remote candidates. https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addIceCandidate
- There is an ice-option for indicating the agent supports trickle. Please see: https://tools.ietf.org/html/draft-ietf-mmusic-trickle-ice-02#section-4 and https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/canTrickleIceCandidates. I think canTrickleIceCandidates is not really important to implement, but it would be nice to add the trickle attribute to offers/answers so the remote can know that trickle ice is supported.
- This will seriously conflict with the checklist pr: https://github.com/pion/ice/pull/42 . Currently every 2 seconds (by default), all candidates pairs are checked for connectivity. In other words a.localCandidates are paired with a.removeCandidates and each pair is checked. With the checklist pr a.localCandidates and a.remoteCandidates are only paired in the beginning, so new candidates will be ignored. Although I think this should be merged first and this issue solved in the checklist pr.

Sean-Der:
Thanks for the review @masterada !!

* We have `AddRemoteCandidate`
* Nice had no idea about this! Will make sure to add to `pion/webrtc`

masterada:
>     * We have `AddRemoteCandidate`
nice, missed that

gather.go
Sean-Der: Fixed!

gather.go
Sean-Der: Fixed!

PR #56, opened 2019-05-29
Modify gathering process to be completely parallel

### LINKED ISSUE #55: Improve gathering process with parallelization
This is just to track the work related to making the gathering process completely parallel in order to improve the gathering speed.

trivigy:
resolves #55

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -3,6 +3,7 @@ package ice
 import (
 	"fmt"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/pion/stun"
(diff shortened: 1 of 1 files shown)


trivigy:
@masterada This is the parallelization you mentioned. Definitely improves the gathering speed.

gather.go
```diff
@@ -146,48 +147,64 @@ func (a *Agent) gatherCandidates() {
...
+	var wg sync.WaitGroup
+	wg.Add(len(localIPs) * len(supportedNetworks))
 	for _, ip := range localIPs {
 		for _, network := range supportedNetworks {
-			conn, err := listenUDP(int(a.portmax), int(a.portmin), network, &net.UDPAddr{IP: ip, Port: 0})
-			if err != nil {
-				a.log.Warnf("could not listen %s %s\n", network, ip)
-				continue
-			}
+			go func(network string, ip net.IP) {
```
Sean-Der: Possible to put the `wg.Done()` in a deferred?

trivigy: done.

gather.go
```diff
@@ -146,46 +147,62 @@ func (a *Agent) gatherCandidates() {
...
 
-			if err := a.run(func(agent *Agent) {
-				set := a.localCandidates[c.NetworkType()]
-				set = append(set, c)
-				a.localCandidates[c.NetworkType()] = set
-			}); err != nil {
-				a.log.Warnf("Failed to append to localCandidates: %v\n", err)
-				continue
-			}
+				if err := a.run(func(agent *Agent) {
```
masterada: this is async, wg.Done can be called before the candidate is added, i suggest another wg (or maybe sync.Condition) here

masterada:
nice :)

gather.go
trivigy: sync.Cond is likely to go away in go 2. I would recommend against that.
https://github.com/golang/go/issues/21165

trivigy: p.s. `a.run` is not really async. It uses an unbuffered channel for the tasks which means only one task is running at a time. In this case, every `a.run` waits for it's turn to write into the channel. Only after that point it runs separately. What happens if you have another task running that is stalling?! Kinda challenging design with this `a.run` but it works.

trivigy: @masterada If you still feel strongly about this specific comment let me know. I will try to figure out how to address it.

masterada: nvm, you set `a.gatheringState = GatheringStateComplete` in a.run as well, so it definitely runs after all candidates are gathered (sorry, was looking on it on mobile :) )

gather.go
```diff
@@ -146,46 +147,62 @@ func (a *Agent) gatherCandidates() {
...
-			conn, err := listenUDP(int(a.portmax), int(a.portmin), network, &net.UDPAddr{IP: ip, Port: 0})
-			if err != nil {
-				a.log.Warnf("could not listen %s %s\n", network, ip)
-				continue
-			}
+			go func(network string, ip net.IP) {
+				conn, err := listenUDP(int(a.portmax), int(a.portmin), network, &net.UDPAddr{IP: ip, Port: 0})
+				if err != nil {
+					a.log.Warnf("could not listen %s %s\n", network, ip)
+					wg.Done()
```
masterada: defer wg.Done() (you call it before any return in this function)

trivigy: done.

gather.go
jinleileiking: It seems defer wg.wait

PR #57, opened 2019-05-29
Improve Gathering error handling

Sean-Der:
Gather only asserted that it wasn't GatheringStateGathering. This
meant that if gathering was complete it would allow multiple runs.


trivigy:
LGTM

PR #62, opened 2019-06-05
Only take the first result for ServerReflexive

### LINKED ISSUE #58: Make srflx gathering configurable
@masterada @trivigy 

We need to make `srflx` gathering only do the 'default' interface by default I think. 

While not spec-compliant I have gotten feedback from multiple people about speed regressions. This wouldn't be a problem if we had Trickle ICE from day one, but lots of people have setup their signaling already.

I say we make configurable, and maybe when we do `/v3` we can roll back to the 'all interface' behavior.

Sean-Der:
Before we would iterate all interfaces and block until all had
completed. Instead return when we get the first candidate.

Resolves #58

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -10,6 +10,10 @@ import (
 	"github.com/pion/turnc"
 )
 
+const (
+	stunGatherTimeout = time.Second * 5
+)
+
 func localInterfaces(networkTypes []NetworkType) (ips []net.IP) {
 	ifaces, err := net.Interfaces()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


gather.go
```diff
@@ -244,28 +252,35 @@ func (a *Agent) gatherCandidatesSrflx(urls []*URL, networkTypes []NetworkType) {
...
 						return
 					}
-
-					if err := a.run(func(agent *Agent) {
-						a.addCandidate(c)
-					}); err != nil {
-						a.log.Warnf("Failed to append to localCandidates: %v\n", err)
-						return
+					select {
+					case serverReflexiveCandidate <- gatherResult{c, conn}:
```
trivigy: How do you know that the first is the default? I think when it returns local interfaces, they are just sorted by interface name. At least that is how `ifconfig` does it.

Sean-Der: There doesn't seem to be a portable way to get the `default` route so couldn't go that route.

Instead for now I say we just take the Srflx candidate that works. If people want we can give them a flag to `wait for all` 

But this way we still respect the port restriction, and we can get a candidate.

trivigy:
I made a way to only do default. I am jumping on a train to head home. Will try recreating it on the train. Give me 30min to share that code before merging.

PR #66, opened 2019-06-07
Add E2E testing for Srflx and Relay candidates

Sean-Der:
Using pion/turn add tests for both candidates types

Resolves #47

PR #68, opened 2019-06-07
Add turn client Close support

ernado:
Fix the goroutine leak.


Sean-Der:
Cherry-picked into #66

PR #71, opened 2019-06-20
Initial stats API

hugoArregui:
Related to https://github.com/pion/webrtc/pull/721

### DIFF
--- a/util.go
+++ b/util.go
@@ -1,6 +1,7 @@
 package ice
 
 import (
+	"fmt"
 	"math/rand"
 	"net"
 	"sync/atomic"
@@ -72,4 +73,19 @@ func addrEqual(a, b net.Addr) bool {
 	}
 
 	return aType == bType && aIP.Equal(bIP) && aPort == bPort
+}
+
+func generateCandidateID() (string, error) {
+	return generateRandString("candidate:", "")
+}
+
+func generateRandString(prefix, sufix string) (string, error) {
+	b := make([]byte, 16)
+	_, err := rand.Read(b) //nolint
+
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("%s%X-%X-%X-%X-%X%s", prefix, b[0:4], b[4:6], b[6:8], b[8:10], b[10:], sufix), nil
 }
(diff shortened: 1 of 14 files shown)


hugoArregui:
@maxhawkins @Sean-Der will you guys be ok by merging this as it is and create an issue to track the missing stats? (the ones commented in agent.go). I don't want to create a huge PR nor to have to a PR open for long and have to rebase all the time.

enobufs:
Can't wait to see this landing!! stats, stats, Stats!

hugoArregui:
@enobufs I know you wanted to use this to get candidate pairs, will your use case be covered by the info I added in this PR?

util.go
```diff
@@ -73,3 +74,7 @@ func addrEqual(a, b net.Addr) bool {
 
 	return aType == bType && aIP.Equal(bIP) && aPort == bPort
 }
+
+func rndCandidateID() string {
+	return fmt.Sprintf("candidate:%d", time.Now().UnixNano())
```
maxhawkins: What if this is called concurrently? Maybe we should use a UUID?

maxhawkins:
Have you given any thought to how this would be surfaced in PeerConnection? We should probably have a GetStats() method to match the browser implementation.

enobufs:
> @enobufs I know you wanted to use this to get candidate pairs, will your use case be covered by the info I added in this PR?

It looks exactly what I would like to have for my diagnostic purpose!

util.go
hugoArregui: yes, that would work too, I'm using nano so I think a clash is very unlikely, but we could use a uuid library too

hugoArregui:
> Have you given any thought to how this would be surfaced in PeerConnection? We should probably have a GetStats() method to match the browser implementation.

Yes, I implemented a GetStats as well https://github.com/pion/webrtc/pull/721/files#diff-d2ee665bb8c62c914fee5e9c743be25cR1836

util.go
Sean-Der: @hugoArregui we could use https://github.com/pion/ice/blob/master/mdns.go#L31

If you want to move it into `util.go` and have the caller append `.local`

hugoArregui: good call! done

PR #73, opened 2019-06-22
Use config objects to create candidates + optional CandidateID

hugoArregui:
Use config objects to create candidates + add optional candidateID

The CandidateID config is useful when testing, otherwise two objects won't match because of the ID

### DIFF
--- a/candidate_host.go
+++ b/candidate_host.go
@@ -12,26 +12,40 @@ type CandidateHost struct {
 	network string
 }
 
+// CandidateHostConfig is the config required to create a new CandidateHost
+type CandidateHostConfig struct {
+	CandidateID string
+	Network     string
+	Address     string
+	Port        int
+	Component   uint16
+}
+
 // NewCandidateHost creates a new host candidate
-func NewCandidateHost(network string, address string, port int, component uint16) (*CandidateHost, error) {
-	candidateID, err := generateCandidateID()
-	if err != nil {
-		return nil, err
+func NewCandidateHost(config *CandidateHostConfig) (*CandidateHost, error) {
+	candidateID := config.CandidateID
+
+	if candidateID == "" {
+		var err error
+		candidateID, err = generateCandidateID()
+		if err != nil {
+			return nil, err
+		}
 	}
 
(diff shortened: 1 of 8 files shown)


hugoArregui:
@Sean-Der can we ignore the lint error somehow?

candidate_host.go
```diff
@@ -12,26 +12,42 @@ type CandidateHost struct {
...
 // NewCandidateHost creates a new host candidate
-func NewCandidateHost(network string, address string, port int, component uint16) (*CandidateHost, error) {
-	candidateID, err := generateCandidateID()
-	if err != nil {
-		return nil, err
+func NewCandidateHost(config *CandidateHostConfig) (*CandidateHost, error) {
+	var candidateID string
+	var err error
+
+	if config.CandidateID != nil {
```
Sean-Der: Instead of a `*` I would use an empty string to force generation (is there ever a case where we want an empty string ID?)

Sean-Der:
@hugoArregui yes we can ignore the lint error! I think you can just `//nolint` next to the line (I don't know what happens on multi-line)

couple other ideas that might work!
* for testing we can maybe just set the IDs ourselves after creation? Then we don't have to change the API at all.
* Just make the ID mandatory first argument, and don't do a struct.
* Have one config for Peer/Server reflexive candidates.

hugoArregui:
> * for testing we can maybe just set the IDs ourselves after creation? Then we don't have to change the API at all.

The problem are not this tests but the webrtc tests in which a candidate is compared to manually crafted candidate, so if two ids are internally generated the would conflict. Also, I think it makes sense to override our name generation if you are using ICE directly (I mean not as part of pion/webrtc).

> * Just make the ID mandatory first argument, and don't do a struct. 
> * Have one config for Peer/Server reflexive candidates.

I can do that as well, what I was trying to do is:

a) avoid breaking changes in the future. The thing is, if someone is using ICE directly everything we change in the candidate creation is a breaking change. Having a config object would prevent it as much as possible because we can add new fields with default values.
b) readability

PR #77, opened 2019-07-12
Leave it to the caller to reject invalid stun URL

mchlrhw:
#### Description
URLs in the stun and stuns schemes should not have a query, but it is useful to leave it up to the caller to decide how to handle invalid URLs, so now we return both the error and the parsed URL.

#### Reference issue
Related to https://github.com/pion/webrtc/issues/735


mjmac:
As I noted in the parent issue, I'm not sure that this is the best approach. I'll defer to @Sean-Der though, as I haven't been involved in the project for a while.

mchlrhw:
Closing in favour of alternative approach.

PR #81, opened 2019-07-17
Update turn to v1.3.2

enobufs:
This is related to pion/webrtc#740

The test with the PR (pion/webrtc#740) was failing because pion/turn had an unportable code which was caused the build failure for wasm.

This PR includes pion/turn@v1.3.2 which removed the offending code.

Also, there was a bug in CandidateRelay.  TURN client's local conn (socket) was not closed on Agent.Close(), (was closing relayConn twice instead) and was causing "unexpected goroutine error".


enobufs:
@Sean-Der Now, what is failing is `TestMulticastDNSOnlyConnection` again.
pion/webrtc#740 is passing already with this branch - so I'm confident this time.  Almost there!

Issue #75, opened 2019-06-27
TestMulticastDNSOnlyConnection frequently timeout in travis


hugoArregui:
```
=== RUN   TestMulticastDNSOnlyConnection
ice ERROR: 2019/06/27 12:08:45 error processing checkCandidatesTimeout handler the agent is closed
goroutine profile: total 43
13 @ 0x4584ff 0x453a4a 0x453036 0x4b3f95 0x4b6745 0x4b6721 0x5f8df0 0x643091 0x64acc3 0x64aca4 0x6520f6 0x486771
#	0x453035	internal/poll.runtime_pollWait+0x55			/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/runtime/netpoll.go:182
#	0x4b3f94	internal/poll.(*pollDesc).wait+0xe4			/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/internal/poll/fd_poll_runtime.go:87
#	0x4b6744	internal/poll.(*pollDesc).waitRead+0x144		/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/internal/poll/fd_poll_runtime.go:92
#	0x4b6720	internal/poll.(*FD).RawRead+0x120			/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/internal/poll/fd_unix.go:534
#	0x5f8def	net.(*rawConn).Read+0x6f				/home/travis/.gimme/versions/go1.12.6.linux.amd64/src/net/rawconn.go:43
#	0x643090	golang.org/x/net/internal/socket.(*Conn).recvMsg+0x3d0	/home/travis/gopath/pkg/mod/golang.org/x/net@v0.0.0-20190619014844-b5b0513f8c1b/internal/socket/rawconn_msg.go:31
#	0x64acc2	golang.org/x/net/internal/socket.(*Conn).RecvMsg+0x212	/home/travis/gopath/pkg/mod/golang.org/x/net@v0.0.0-20190619014844-b5b0513f8c1b/internal/socket/socket.go:255
#	0x64aca3	golang.org/x/net/ipv4.(*payloadHandler).ReadFrom+0x1f3	/home/travis/gopath/pkg/mod/golang.org/x/net@v0.0.0-20190619014844-b5b0513f8c1b/ipv4/payload_cmsg.go:31
[...]

PR #83, opened 2019-07-18
More error handling around interface iteration

Sean-Der:
If we can't iterate host interfaces give a warning to the user.
Also don't bail early if a single interface fails, other interfaces
might still be in a good condition.

Relates to pion/ice#75


Sean-Der:
@enobufs @hugoArregui this also bump mdns (which contains the real bug fix)

https://github.com/pion/mdns/pull/2

PR #89, opened 2019-08-12
avoid agent.run for getSelectedPair

hugoArregui:
Procted the selected candidate pair by a RWMutex so we don't need to relay in the agent taskloop in order to use this method

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -127,9 +127,12 @@ type Agent struct {
 	remotePwd        string
 	remoteCandidates map[NetworkType][]Candidate
 
-	checklist    []*candidatePair
-	selector     pairCandidateSelector
-	selectedPair *candidatePair
+	checklist []*candidatePair
+	selector  pairCandidateSelector
+
+	selectedPairMutex sync.RWMutex
+	selectedPair      *candidatePair
+
 	urls         []*URL
 	networkTypes []NetworkType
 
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -629,15 +640,19 @@ func (a *Agent) validateSelectedPair() bool {
 // if no packet has been sent on that pair in the last keepaliveInterval
 // Note: the caller should hold the agent lock.
 func (a *Agent) checkKeepalive() {
-	if a.selectedPair == nil {
+	a.selectedPairMutex.RLock()
```
Sean-Der: Use `getSelectedPair` instead?

agent.go
```diff
@@ -610,14 +615,20 @@ func (a *Agent) taskLoop() {
 // validateSelectedPair checks if the selected pair is (still) valid
 // Note: the caller should hold the agent lock.
 func (a *Agent) validateSelectedPair() bool {
-	if a.selectedPair == nil {
+	a.selectedPairMutex.RLock()
```
Sean-Der: use `getSelectedPair` instead?

Sean-Der:
@hugoArregui we should probably get rid of the `run` construct all together. If you see any another performance wins/chance to do it I would say go for it!

hugoArregui:
> @hugoArregui we should probably get rid of the `run` construct all together. If you see any another performance wins/chance to do it I would say go for it!

Agree!

PR #91, opened 2019-08-14
Improve candidate base locking mechanism

hugoArregui:
- Use atomics for lastReceived and lastSent fields
- the lock around resolvedAddr and conn were not being used, so I removed it

### DIFF
--- a/candidate_base.go
+++ b/candidate_base.go
@@ -3,7 +3,7 @@ package ice
 import (
 	"fmt"
 	"net"
-	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/pion/logging"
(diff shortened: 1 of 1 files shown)


Sean-Der:
@hugoArregui I don't know how much of a performance change it will be, but could you try making these atomics?

It would be great if we only use mutex in the case of `Block until this thing is ready` but for stuff we want to sample and `nil value is ok` I would love to use atomics. It just makes the code easier to reason about/will make debugging around waiting for mutexes easier.

Push back if you think this is a bad idea!

Sean-Der:
@hugoArregui one thing we need to be careful about is that atomics have alignment issues on x86. I will work on setting up a tester on travis-ci today though!

Sean-Der:
@enobufs has a bunch of knowledge around this also. I will see what I can do with travis-ci and getting multiple architectures going

PR #94, opened 2019-08-29
Lite

seppo0010:


Sean-Der:
@seppo0010  Great PR, really excited that Pion is working out for your use cases!

I am curious reviewing this code, what value is forcing ICE lite providing you? Do you want to force `host` candidates only, or do you just not want outbound connectivity checks?

I am curious if we can find a better compromise. I think this is going to add a decent amount of complexity to the code. We basically need two more selectors `ice lite controlled` and `ice lite controlling` and I would prefer not to take any short cuts.

thanks!

seppo0010:
I do not want to have outbound connectivity checks. I'm running `pion` in a server I own on the internet with a known public IP and ports so I am certain all users will be able to connect, similarly to running a web server, therefore I know I don't need those checks.

Sean-Der:
@seppo0010 

I am not pushing back, just curious. Why do you not want outbound connectivity checks? Are you worried about bandwidth, scaling, security or something else?

I want to make Pion for you 100%, but I am just trying to make it work with the least lines of code possible :) ICE is probably our most fragile component so always worried to touch it.

I am totally up for going this way though. Just want to see if maybe we can make more minimal changes for it to work for you.

seppo0010:
I'm mostly worried about security, although not much. I work in a big company and I'm sure the infrastructure team does not appreciate machines performing random connectivity checks, specially to user-provided IP addresses and ports. That means that any user can provide an IP address and a port inside our private network and some package will be sent, pf course they cannot choose what will go in it and it will be delivered using UDP so it's probably just fine, but I guess [Server-Side Request Forgery](https://www.owasp.org/index.php/Server_Side_Request_Forgery) is the problem I'm trying to avoid.

I'm OK with not merging this change, I can maintain my own fork if my use case is not common enough to take the complexity cost in the project.

Sean-Der:
Oh no that is an excellent point! We should get this merged, I didn't consider this at all.

My worst case scenario is having you fork/not use upstream! So I really want to get you using master, I am sure all your use cases are being hit by others. They are just not communicating it though.
* We lose out on you testing the latest changes and giving feedback
* I really love having smart people involved in the community :)
* Lots of other stuff I am forgetting!

selection.go
```diff
@@ -298,3 +298,16 @@ func (s *controlledSelector) HandleBindingRequest(m *stun.Message, local, remote
...
 }
+
+type liteSelector struct {
+	pairCandidateSelector
+}
+
+// A lite selector should not contact candidates
+func (s *liteSelector) ContactCandidates() {
+	if _, ok := s.pairCandidateSelector.(*controllingSelector); ok {
+		// TODO: implement lite controlling agent. For now falling back to full agent.
```
Sean-Der: Can you create a github issue and tag it in the code base and just make the comment `pion/ice#$ISSUE_ID` otherwise we will never come back to it :)

I added `godox` to `pion/webrtc` but not all the dependencies yet!

agent.go
```diff
@@ -389,10 +409,13 @@ func NewAgent(config *AgentConfig) (*Agent, error) {
 		a.taskLoopInterval = config.taskLoopInterval
 	}
 
-	if config.CandidateTypes == nil || len(config.CandidateTypes) == 0 {
```
Sean-Der: We still want to use the user provided `CandidateTypes` if they passed in. I just want to do an assert when `lite` is chosen. I think the user should be explicit when using `lite` so they fully understand what is happening. We can loosen this restriction later, but I just don't want inputs to conflict.

```
if a.lite && (len(a.candidateTypes) != 1 || a.candidateTypes[0] != CandidateTypeHost) {
  return nil ErrLiteUsingNonHostCandidates
}
```

agent.go
```diff
@@ -404,6 +427,19 @@ func NewAgent(config *AgentConfig) (*Agent, error) {
 	return a, nil
 }
 
+func (a *Agent) getCandidateTypesForConfig(config *AgentConfig) ([]CandidateType, error) {
```
Sean-Der: We can remove. I would make users be explicit, I just want to keep this as simple as possible.

Sean-Der:
@seppo0010 ok if you don't mind rebasing and addressing minor changes I am happy to merge! Then we can land stuff on the other side and you can be unblocked :)

Sorry for moving slow, will try to get this through faster now.

seppo0010:
Hi @Sean-Der. I'm confused about the candidates suggestion.

Currently it is respecting the user input if provided (unless it is lite and the decision goes against the spec!) and it is using different defaults for the full agent and the lite agent. What would you like to see changed? This provides four different results based on the received configuration so I extracted it as a method for clarity, but I think the behaviour is the one you are asking.

Sean-Der:
This landed with https://github.com/pion/ice/commit/953f36f07fa902245756924e5a9ae2499b829ebf thank you @seppo0010!

PR #99, opened 2019-09-07
Clean up switch conditions

ZizhengTai:
#### Description

Remove some unnecessary repetitions.


Sean-Der:
Nice catch

This seems like a good change to get into linters! No idea if this would be easy, but could have a high impact. Makes the code a lot easier to reason (only have to worry about one piece of state)

PR #101, opened 2019-09-15
Add ability to filter interfaces

AeroNotix:
#### Description

Adds ability to filter specific interfaces via the use of a call back which is passed the name of the interface.

* An accompanying PR will be sent to pion/webrtc in order to expose
this through the SettingsEngine

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -150,6 +150,8 @@ type Agent struct {
 	log           logging.LeveledLogger
 
 	net *vnet.Net
+
+	interfaceFilter func(string) bool
 }
 
 func (a *Agent) ok() error {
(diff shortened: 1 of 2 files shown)


Sean-Der:
Hey @AeroNotix LGTM!

Mind adding some tests to assert the filtering actually happens/nil filters don't cause a crash etc..

AeroNotix:
> Hey @AeroNotix LGTM!
> 
> Mind adding some tests to assert the filtering actually happens/nil filters don't cause a crash etc..

Sure thing!

AeroNotix:
@Sean-Der any tips on how to test this?

Sean-Der:
Hey @AeroNotix 

I would check out `gather_vnet_test.go` this is used to simulate networks. I would write tests that assert.

* Interface name gets passed through properly
* Excluding interfaces properly works. (You can query an agent for local interfaces I believe)

That should be a good start! We can always come back if we hit bugs in the future

Sean-Der:
Opening a new PR with tests

PR #103, opened 2019-09-20
Support 1:1 NAT option

enobufs:
Relates pion/webrtc#835
Please see pion/webrtc#835 for details.


enobufs:
Blocked by https://github.com/pion/transport/pull/38

enobufs:
I have just added agent config checks to detect incompatible cases as planned in the following chart (along with tests for those):
https://docs.google.com/spreadsheets/d/1doEPGTC5WHaN5N-9XcsF3I_-yB8rp04Jkv0HLHxkbtw

Sean-Der:
@enobufs that is cool!

You should propose to W3C to use those values, maybe they can enable/disable mDNS using that key as well.

Sean-Der:
that would make a lot of people happy for sure

enobufs:
"NoHost" is defined in libwebrtc, but not exposed:
```c
/**
 * Represents the ice transport policy. This exposes the same states in C++,
 * which include one more state than what exists in the W3C spec.
 */
typedef NS_ENUM(NSInteger, RTCIceTransportPolicy) {
  RTCIceTransportPolicyNone,
  RTCIceTransportPolicyRelay,
  RTCIceTransportPolicyNoHost,
  RTCIceTransportPolicyAll
};
```

external_ip_mapper.go
```diff
@@ -0,0 +1,129 @@
...
+				m.ipv4Map[locIP.String()] = extIP
+			} else {
+				if isLocIPv4 {
+					return nil, ErrInvalidNAT1To1IPMapping
+				}
+				// check if dup of local IP
+				if _, ok := m.ipv6Map[locIP.String()]; ok {
+					return nil, ErrInvalidNAT1To1IPMapping
+				}
+				m.isLocIPv6Specified = true
```
hugoArregui: super silly nitpick: do you perhaps want to move this at the beggining of the block so it's simetric to the true block?

external_ip_mapper.go
```diff
@@ -0,0 +1,129 @@
...
+		ipPair := strings.Split(extIPStr, "/")
+		if len(ipPair) == 0 || len(ipPair) > 2 {
+			return nil, ErrInvalidNAT1To1IPMapping
+		}
+
+		extIP, isExtIPv4, err := validateIPString(ipPair[0])
+		if err != nil {
+			return nil, err
+		}
+		if len(ipPair) == 1 {
```
hugoArregui: sorry if this is not valid, I should test this myself but I want to give you a review before going to bed:

Let's say the first mapping is a single ip `1.2.3.4` and the second one is a pair `1.2.3.5/10.1.5.26`. What your ipv4Map will contain, if I'm reading the code correctly, is: 

```
{ 
   "*": "1.2.3.4",
   "10.1.5.26": "1.2.3.5"
}
```

and `isLocIPv4Specified` will be true. 

Now, something is not right if that's the case, because the `findExternalIP` will ignore `*` if `isLocIPv4Specified` is true

external_ip_mapper.go
enobufs: Yes. Actually, the else block is correct. I will modify the if block. Thanks!

external_ip_mapper.go
enobufs: You are such an awesome reviewer! I did not have the test for that case, so I added one, which ended up with the exact result you predicted! And that is not what I expected. I will fix that.

hugoArregui: Glad it helped!, I was thinking about how can we simplify the code in order to catch this kind of errors and I came up with a little refactor: https://github.com/pion/ice/commit/3992531489c3aaead524f4145bfab86ce44c7f7d let me know what you think (feel free to pick just some pieces or nothing at all). Anyway I will aprove the PR right now

enobufs: Oh, I love it. Thank you!  I merged your commit, then did a bit of naming and trivial code optimization.

enobufs:
Build error with commit message?? What's wrong?
```
IP mapper refactoring

Relates to pion/webrtc#385
```

Let me rebase master...

hugoArregui:
Oh, probably my commit message is breaking the message linter, I hate that linter

enobufs:
@Sean-Der @seppo0010 are you happy the implementation so far?
I will soon work on pion/webrtc to expose this option to API!

external_ip_mapper.go
```diff
@@ -0,0 +1,143 @@
...
+
+	extIP, ok := m.ipMap[locIP.String()]
+	if !ok {
+		return nil, ErrExternalMappedIPNotFound
+	}
+
+	return extIP, nil
+}
+
+type externalIPMapper struct {
```
seppo0010: This is not my area of knowledge, but I don't think this data modeling is correct.

My main concern is why do we have an ipMapping for ipv4 and one for ipv6. The keys of both mappings should never collide, and I think it should be possible to have an external ipv4 address mapping to a local ipv6 one while it cannot happen now. If we have one mapping though we cannot have a _sole_ ipv4 and a _sole_ ipv6 address. And now I'm questioning... should we have a _sole_ ip address, or even one ipv4 and one ipv6 address at most? I think it should be possible to specify multiple ip addresses as candidates (e.g.: in a client/server scenario multiple servers are available and the client can choose the fastest one). Maybe I'm overthinking it. Feel free to ignore this comment or ping me if you want some clarification.

enobufs: Yea, I had one map for both ip families at first. Then I realized I needed to make sure (when local IP is not specified), the mapping should allow up to one IPv4 and one IPv6, and the code turned out much cleaner with the separate map... but that is all. (just an implementation detail)

PR #111, opened 2019-09-29
fix: Controlled selectors now timeout with CandidateSelectionTimeout

AeroNotix:
I noticed some peer connections weren't closing when the peer was a controlled selector and its state was in the `checking` state (i.e. waiting to determine good candidate pairs). With this PR a controlled selector will timeout its connectivity checks after `CandidateSelectionTimeout` has passed.

### DIFF
--- a/selection.go
+++ b/selection.go
@@ -206,7 +206,7 @@ func (s *controlledSelector) ContactCandidates() {
 			s.agent.checkKeepalive()
 		}
 	} else {
-		if time.Since(s.startTime) > s.agent.connectionTimeout {
+		if time.Since(s.startTime) > s.agent.candidateSelectionTimeout {
 			s.log.Trace("check timeout reached and no valid candidate pair found, marking connection as failed")
 			s.agent.updateConnectionState(ConnectionStateFailed)
 		} else {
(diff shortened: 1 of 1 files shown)


selection.go
```diff
@@ -203,8 +206,13 @@ func (s *controlledSelector) ContactCandidates() {
 			s.agent.checkKeepalive()
 		}
 	} else {
-		s.log.Trace("pinging all candidates")
-		s.agent.pingAllCandidates()
+		if time.Since(s.startTime) > s.agent.connectionTimeout {
```
hugoArregui: I think we should use s.agent.candidateSelectionTimeout instead

other than that, LGTM!

AeroNotix: Pushed that change. :+1:

hugoArregui:
@AeroNotix this is approved, you need to squash the commits and set a commit message that works for the linter:

```
* Separate subject from body with a blank line
* Limit the subject line to 50 characters
* Capitalize the subject line
* Do not end the subject line with a period
* Wrap the body at 72 characters
```

(you can test it locally by running .github/lint-commit-message.sh)

AeroNotix:
Will do Hugo!

Sean-Der:
Landed with 377e52e387c26d40dc895519c9b4bba8ebe07f3e

Thanks @AeroNotix! Just unblocking for TCP, thanks for fixing this :)

PR #112, opened 2019-09-30
feat: semantic release support

hugoArregui:
semantic release support

### DIFF
--- a/.github/lint-commit-message.sh
+++ b/.github/lint-commit-message.sh
@@ -1,43 +1,19 @@
 #!/usr/bin/env bash
 set -e
 
-display_commit_message_error() {
-cat << EndOfMessage
-$1
-
--------------------------------------------------
-The preceding commit message is invalid
-it failed '$2' of the following checks
-
-* Separate subject from body with a blank line
-* Limit the subject line to 50 characters
-* Capitalize the subject line
-* Do not end the subject line with a period
-* Wrap the body at 72 characters
-EndOfMessage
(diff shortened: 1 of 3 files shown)


Sean-Der:
LGTM! I am totally up for changing to semantic versioning.

https://www.npmjs.com/package/commitplease looks like what angularjs uses. Looking through their PRs people constantly spam them with commits breaking their rules:)

hugoArregui:
@Sean-Der yes, but it requires node :disappointed: not sure if we should install node just for that (I have no strong opinion on the matter,)

Sean-Der:
> @Sean-Der yes, but it requires node 😞 not sure if we should install node just for that (I have no strong opinion on the matter,)

Agree. Would rather stickto Go (or just shell script!) it looks like the rules are a little more complicated than the checks we have now though.

I am happy as long as subject/body are split up and we have a length limit. This will be really great for generating a changelog, nice work!

```
<type>(<scope>): <subject>
<BLANK LINE>
<body>
<BLANK LINE>
<footer>
```

hugoArregui:
@Sean-Der body is optional though, is that ok?

hugoArregui:
I stil need to test the release script but I need to do it in master directly, if you are ok with everything can you approve @Sean-Der ?

Sean-Der:
@hugoArregui LGTM! When I get a chance I will improve the checks (assert empty lines and check line lengths)

PR #119, opened 2019-10-15
Gather STUN/TURN URLs concurrently

### LINKED ISSUE #118: Gather TURN/STUN candidates async
Currently we gather them one at a time. 

However if one TURN server is slow it can block gathering all the other ones.

@AeroNotix can offer more details/has context also

Sean-Der:
Gather candidates of each type concurrently, this way a single slow
TURN/STUN server doesn't block the others.

We should make the entire gather process more concurrent and spawn each
gather in a separate goroutine eventually

Resolves #118

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -885,25 +885,32 @@ func (a *Agent) addRemoteCandidate(c Candidate) {
 	a.requestConnectivityCheck()
 }
 
-// addCandidate assumes you are holding the lock (must be execute using a.run)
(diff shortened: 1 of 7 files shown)


Sean-Der:
@AeroNotix TURN URLs will now gather concurrently, this will probably still have issue because STUN/TURN block each other

Mind checking this out and see if it unblocks you in the short term. If not I can do the full rewrite tomorrow!

PR #121, opened 2019-10-15
pending binding requests are removed from the queue after 500ms if the queue is full

jjmaldonis:


agent.go
```diff
@@ -52,14 +52,18 @@ const (
 	maxBufferSize = 1000 * 1000 // 1MB
 
 	// the number of outbound binding requests we cache
-	maxPendingBindingRequests = 50
+	maxPendingBindingRequests = 500
```
Sean-Der: We can drop this entirely.

agent.go
```diff
@@ -953,14 +954,15 @@ func (a *Agent) findRemoteCandidate(networkType NetworkType, addr net.Addr) Cand
 func (a *Agent) sendBindingRequest(m *stun.Message, local, remote Candidate) {
 	a.log.Tracef("ping STUN from %s to %s\n", local.String(), remote.String())
 
-	if overflow := len(a.pendingBindingRequests) - (maxPendingBindingRequests - 1); overflow > 0 {
-		a.log.Debugf("Discarded %d pending binding requests, pendingBindingRequests is full", overflow)
-		a.pendingBindingRequests = a.pendingBindingRequests[overflow:]
+	nRemoved := a.invalidatePendingBindingRequests()
+	if nRemoved > 0 {
+		a.log.Tracef("Discarded %d binding requests because they expired", nRemoved)
```
jjmaldonis: Do you want this as a Trace message? It prints how many binding requests were removed due to expiration.

agent.go
```diff
@@ -985,9 +987,30 @@ func (a *Agent) sendBindingSuccess(m *stun.Message, local, remote Candidate) {
...
+		a.pendingBindingRequests[i] = bindingRequest{}
+	}
+	a.pendingBindingRequests = temp
+	return nRemoved
+}
+
 // Assert that the passed TransactionID is in our pendingBindingRequests and returns the destination
 // If the bindingRequest was valid remove it from our pending cache
 func (a *Agent) handleInboundBindingSuccess(id [stun.TransactionIDSize]byte) (bool, *bindingRequest) {
+	a.invalidatePendingBindingRequests()
```
jjmaldonis: `Agent.handleInboundBindingSuccess` now runs two for-loops over the same list, which is not ideal. The benefit is that the functionality for invalidating pending requests and identifying the valid request is separated. Let me know if you want to move the invalidation logic into the `handleInboundBindingSuccess` method.

agent.go
Sean-Der: I think this is good!

agent.go
Sean-Der: I think this is fine! We can measure and rewrite if it is a problem

agent.go
```diff
@@ -985,9 +987,30 @@ func (a *Agent) sendBindingSuccess(m *stun.Message, local, remote Candidate) {
...
+		if time.Since(bindingRequest.timestamp) < maxBindingRequestTimeout {
+			temp = append(temp, bindingRequest)
+		} else {
+			nRemoved++
+		}
+	}
+	// Garbage collect removed elements
+	for i := len(temp); i < len(a.pendingBindingRequests); i++ {
+		a.pendingBindingRequests[i] = bindingRequest{}
+	}
```
Sean-Der: Will GC never happen if we do this? Is this a language or implementation detail? 

We don't do this in other places so I would push back against this (unless you measured and it is a real problem!)

agent.go
```diff
@@ -985,9 +987,30 @@ func (a *Agent) sendBindingSuccess(m *stun.Message, local, remote Candidate) {
...
+	a.pendingBindingRequests = temp
+	return nRemoved
```
Sean-Der: Instead of `nRemoved` what do you think of doing `preInvalidateLength - len(a.pendingBindingRequests )`?

Just nice to have a little less state, I read this function a few times to make sure I fully understood everything that was happening. If you disagree please push back, I don't feel strongly :)

Sean-Der:
Merged with 34ef65fcea0e14a864300de4ab715e700830c4cc thanks again @jjmaldonis :)

PR #125, opened 2019-10-31
Update to pion/transport@v0.8.10

Sean-Der:
Tests now assert for routine leaks on startup as well


agent.go
```diff
@@ -331,6 +331,13 @@ func NewAgent(config *AgentConfig) (*Agent, error) {
 	if err != nil {
 		return nil, err
 	}
+	closeMDNSConn := func() {
```
Sean-Der: Before we leaked mDNS connections when errors happened, nice little find from improved tooling :)

PR #127, opened 2019-11-07
Configuration support for local agent username fragment and password

0x19:
#### Description

Adding configuration support for local username fragment and password, making sure we are validating values based on the RFC standard. Wrote unit test case to cover backward compatibility and test RFC standard validation.

#### Unit Test Case Output

```
go test -run TestAgentCredentials
ice ERROR: 2019/11/07 03:50:38 failed to accept local username fragment as it is less than 24 bits long
ice ERROR: 2019/11/07 03:50:38 failed to accept local password as it is less than 128 bits long
PASS
ok  	github.com/pion/ice	0.032s
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Aaron France](https://github.com/AeroNotix)
 * [Chao Yuan](https://github.com/yuanchao0310)
 * [Jason Maldonis](https://github.com/jjmaldonis)
+* [Nevio Vesic](https://github.com/0x19)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/agent.go
+++ b/agent.go
@@ -185,6 +185,13 @@ type AgentConfig struct {
 	PortMin uint16
 	PortMax uint16
 
+	// LocalUfrag and LocalPwd values used to perform connectivity
+	// checks.  The values MUST be unguessable, with at least 128 bits of
+	// random number generator output used to generate the password, and
+	// at least 24 bits of output to generate the username fragment.
+	LocalUfrag string
+	LocalPwd   string
+
 	// Trickle specifies whether or not ice agent should trickle candidates or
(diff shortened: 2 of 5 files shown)


Sean-Der:
Setting the local uFrag/uPwd instead of randomly generating 👍 from me!

Is it possible to use https://github.com/pion/ice/blob/master/transport.go#L13-L23 Dial/Accept to set the remote?

Sean-Der:
I am all for returning an error if the local ufrag/upwd is invalid!

However, we probably can't block on the remote. Even though it does diverge I still would like to be able to connect to non-compliant agents.

0x19:
Thanks!

Yes, it does :) 

I tried to test it numerous times while building a solution for the company. 

What we do is use ice directly to make a peer discovery program, and when doing a dial/accept, passed username and password failed to be accepted for remote and local. 

The reason for it is the NewAgent creation process where we hand code random bit creation, and when check occurs, comparison of what is in the struct and what is passed in as Dial/Accept arguments fail and we get rejection message.

0x19:
Let me make sure remote frag/pwd is set only if passed along. Great catch!

0x19:
Made sure remote user frag and pwd are set only if configuration values are set. Otherwise fails to the default, empty strings.

Sean-Der:
@0x19 how do you start the agent now? Do you use `Dial`/`Accept` or something else?

That allows you to pass the remote details, why would you rather pass it via the `AgentConfig`? https://github.com/pion/ice/blob/master/agent.go#L564

Sorry I might be missing something, it has been a while since I have worked on ICE stuff.

0x19:
Will create a greater unit test case tomorrow to explain what is happening but you might have the point with fully removing remote. :D

0x19:
@Sean-Der you were right. Tested it and in fact, it has no meaning to be in the configuration as remote frag/pwd will be overwritten by the Dial() itself.

Removed it and left only local frag/pwd available to be changed through the configuration.

Thanks for pointing that out!

It's now ready to be reviewed :)

PR #132, opened 2019-12-11
Check for nil when loading atomic values

dihamilton:
#### Description

This is required to prevent an error when values are loaded without first being set.

I'm not sure about follow on effects so it would be worth someone with more experience taking a look.

The affected areas are: https://github.com/pion/ice/blob/57e7b93a4bac70ee2fd3b38cc2fdf3c6d586ed27/agent.go#L777
and
https://github.com/pion/ice/blob/57e7b93a4bac70ee2fd3b38cc2fdf3c6d586ed27/agent.go#L798
where I think calling PingCandidate if lastSent has not been set is probably fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Chao Yuan](https://github.com/yuanchao0310)
 * [Jason Maldonis](https://github.com/jjmaldonis)
 * [Nevio Vesic](https://github.com/0x19)
+* [David Hamilton](https://github.com/dihamilton)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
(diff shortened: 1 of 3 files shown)


Sean-Der:
@dihamilton nice work :) Congrats on the first contribution 

I think both are fine! The one on `777` is check for timeouts, so having a zero value makes sense.

The one on `798` also makes sense it will ping the candidate (and then update the value) so it will not be spammy.

Sean-Der:
Merged! 11a6fedef9dc84e89ee767b8e0ae501f9c447240 and  5870fb0c99371982fa77dd394e6d51fcb62a372c

I had to run `gofmt` on the second commit to make it pass `golangci-lint`

dihamilton:
@Sean-Der thanks for the gofmt change! I'll add that to my tooling on save.

PR #144, opened 2020-01-13
Fix TestRelayOnlyConnection on debian

at-wat:
Debian based linux uses 127.0.1.1 as a local loopback address
in some condition of the install.

ref: https://github.com/pion/ice/pull/142


at-wat:
passed on bionic: https://travis-ci.org/pion/ice/builds/636329751

PR #159, opened 2020-02-24
Implement TLS and DTLS Gathering

### LINKED ISSUE #133: Handle `turns` URLS
Currently we only do UDP+TCP. We need to add DTLS and TLS

Sean-Der:
Also add tests to cover.

Resolves #133


Sean-Der:
No rush I am going to bed!

PR #169, opened 2020-04-04
Fix async processing

at-wat:
- Fix asynchronous task processing
  - Make callback handler atomic/Value.
  - Launch candidate callback loop in gatherCandidates().
- Asynchronously add remote candidate
  - As same as mDNS candidate.

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -55,12 +55,12 @@ func (a *Agent) GatherCandidates() error {
 		if a.gatheringState != GatheringStateNew {
 			gatherErrChan <- ErrMultipleGatherAttempted
 			return
-		} else if a.onCandidateHdlr == nil {
+		} else if a.onCandidateHdlr.Load() == nil {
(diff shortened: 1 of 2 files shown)


gather.go
```diff
@@ -71,45 +71,67 @@ func (a *Agent) GatherCandidates() error {
...
+				if err := a.gatherCandidatesRelay(a.urls); err != nil {
+					a.log.Errorf("Failed to gather relay candidates: %v\n", err)
+				}
 			}
 		}
-	}
-	if err := a.run(func(agent *Agent) {
-		if a.onCandidateHdlr != nil {
-			a.chanCandidateCallback <- func() { a.onCandidateHdlr(nil) }
+		if err := a.run(func(agent *Agent) {
```
Sean-Der, line 119: I would combine the `run` calls. This just exists to allow 'cancelling' of queued tasks when `Close()` is called on the agent. Since the tasks are similar I say squash them.

PR #184, opened 2020-05-13
Make mDNS opportunistic.

oridb:
It seems like mDNS was intended to be opportunistic, but right now,
if creating an mDNS client fails, the whole ICE connection fails.
This allows pion to work in more restricted network namespaces,
or on less supported environments like 9front.

This allows webwormhole work on Plan 9.


Sean-Der:
Fantastic! Can you just log the error instead?

The linter won’t let us just ignore.

oridb:
Done.

Sean-Der:
Merged and tagged! Thank you so much @oridb 

I am going to also tag a new version of `pion/webrtc`! Want me to open the PR to webwormhole?

Sean-Der:
Merged as 61849bec4743d1882132130805403d188ac4e345 I just added back setting `err`

PR #185, opened 2020-05-16
Agent.ResetPair() func instead connection timeout

petkunand:
#### Description
force reconnect

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -745,7 +745,7 @@ func (a *Agent) findPair(local, remote Candidate) *candidatePair {
 	return nil
 }
 
-// ResetPair connection
+// ResetPair initiate ice restart on client without send candidates
 func (a *Agent) ResetPair() {
 	a.setSelectedPair(nil)
 }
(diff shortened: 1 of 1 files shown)


Sean-Der:
Thanks for the contribution @petkunand 

For this functionality we need to implement ICE Restarts. ICE doesn't allow changing the selected pair otherwise. The RFC is [here](https://tools.ietf.org/html/rfc8445#section-9)

petkunand:
@Sean-Der 
ICE Restart possible if place to pion/webrtc/peerconnection.go this code:

// ResetPair connection
func (pc *PeerConnection) ResetPair() {
	if pc.iceGatherer != nil && pc.iceGatherer.agent != nil {
		pc.iceGatherer.agent.ResetPair()
	}
}

it worked in one of my projects

Sean-Der:
Interesting! I have no idea how/why this is working (but will investigate)

Is Pion the offer or answerer? It sounds like Google's ICE stack just allows multiple nominations/selections? 

Even if it does work we need to be careful, Google can change things at anytime and could break Pion. So we have to always make sure to be spec compliant/not program against their bugs.

petkunand:
pion offer

PR #200, opened 2020-06-22
Implement ICE Restarts

Sean-Der:
Add two new API calls `Restart(Pwd, Ufrag)` and `SetRemoteCredentials(Pwd, Ufrag)`

Calling `Restart` moves the ICE Agent back to Gathering. `SetRemoteCredentials` can be called and atomically updates the current remote credentials


Sean-Der:
The following two commits are the ones that matter the most. Thanks for reviewing everyone!

https://github.com/pion/ice/pull/200/commits/5c5a787bb0ef0e91291f41124a78868c3aa92b12 
https://github.com/pion/ice/pull/200/commits/d354ed7978e36d9cf317aad33ddad725336cbd2b

PR #201, opened 2020-06-24
Agent: More frequent candidate check in connecting state

nerd2:
#### Description

In the ICE connecting state, candidates may not be nominated until 500mS or 1S after ICE start time. But currently the task ticker only runs every 2 seconds by default, leading to slower than necessary ICE completion.

Run the task check ticker more frequently when we don't have chosen pairs.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -38,8 +38,7 @@ type Agent struct {
 	onConnected     chan struct{}
 	onConnectedOnce sync.Once
 
-	connectivityTicker *time.Ticker
-	// force candidate to be contacted immediately (instead of waiting for connectivityTicker)
+	// force candidate to be contacted immediately (instead of waiting for task ticker)
 	forceCandidateContact chan bool
 
 	tieBreaker uint64
@@ -82,8 +81,8 @@ type Agent struct {
 	// 0 means never
 	keepaliveInterval time.Duration
 
-	// How after should we run our internal taskLoop
-	taskLoopInterval time.Duration
+	// How often should we run our internal taskLoop to check for state changes when connecting
+	checkInterval time.Duration
 
 	localUfrag      string
 	localPwd        string
@@ -362,9 +361,7 @@ func (a *Agent) startConnectivityChecks(isControlling bool, remoteUfrag, remoteP
 
 		agent.updateConnectionState(ConnectionStateChecking)
 
-		// TODO this should be dynamic, and grow when the connection is stable
 		a.requestConnectivityCheck()
-		agent.connectivityTicker = time.NewTicker(a.taskLoopInterval)
 		go a.connectivityChecks()
 	}, nil)
 }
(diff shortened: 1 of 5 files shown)


Sean-Der:
Hey @nerd2 

Sorry I just added `ICE Restarts` and removed non-trickle ICE, so have caused merged conflicts. I do think this change make sense! 

Would you be able to check `a.connectionState` instead maybe? I think that would be more clear to the reader. Do you have a simple test that can demonstrate the speed up that comes from this? Would be nice if we could find a simple test to make sure this doesn't regress.

thanks!

nerd2:
Updated to handle the conflict. No idea how to write a test for this, you don't use any sort of time mocking anywhere do you?

agent.go
```diff
@@ -405,12 +402,22 @@ func (a *Agent) connectivityChecks() {
 	}
 
 	for {
+		interval := a.taskLoopInterval
+		switch lastConnectionState {
+		case ConnectionStateNew, ConnectionStateChecking: // While connecting, check candidates more frequently
+			interval /= 10
```
at-wat, line 408: I think it's better to add a new config parameter to specify ping interval during checking.
Then, interval can be choosed like
```go
var interval time.Duration
switch lastConnectionState {
case ConnectionStateNew, ConnectionStateChecking: // While connecting, check candidates more frequently
	interval = a.checkInterval
case ConnectionStateConnected, ConnectionStateDisconnected:
	interval = a.keepaliveInterval
default:
	// ConnectionStateFailed
	return
}
```

agent_config.go
```diff
@@ -100,6 +103,11 @@ type AgentConfig struct {
 	// keepAlive behavior should be modified with KeepaliveInterval and ConnectionTimeout
 	taskLoopInterval time.Duration
 
+	// taskLoopInterval controls how often our internal task loop runs when in the connecting phase,
+	// this would normally be more frequent thank the normal "taskLoopInterval", to allow
+	// quick state changes and connection establishment
+	TaskLoopConnectingInterval time.Duration
```
at-wat, line 109: IMO, `TaskLoop` would be better not to be appeared in public API as it is from internal implementation. (It can also be implemented by `time.Timer` without having a loop.)
It may be something like `ConnectingPingInterval`?

agent_config.go
```diff
@@ -100,6 +103,11 @@ type AgentConfig struct {
 	// keepAlive behavior should be modified with KeepaliveInterval and ConnectionTimeout
 	taskLoopInterval time.Duration
 
+	// taskLoopInterval controls how often our internal task loop runs when in the connecting phase,
```
at-wat, line 106: Please update `taskLoopInterval` to the field name for godoc.

agent_config.go
```diff
@@ -100,6 +103,11 @@ type AgentConfig struct {
...
+	// this would normally be more frequent thank the normal "taskLoopInterval", to allow
```
at-wat, line 107: The code snippet I posted at https://github.com/pion/ice/pull/201#discussion_r446602110 make it possible to remove `taskLoopInterval` and users no longer have to consider `be more frequent thank the normal "taskLoopInterval"`.

at-wat: It may be a different PR, and I'll do it if you are Ok.

agent_config.go
```diff
@@ -95,10 +95,9 @@ type AgentConfig struct {
 
 	LoggerFactory logging.LoggerFactory
 
-	// taskLoopInterval controls how often our internal task loop runs, this
-	// task loop handles things like sending keepAlives. This is only value for testing
-	// keepAlive behavior should be modified with KeepaliveInterval and ConnectionTimeout
-	taskLoopInterval time.Duration
+	// checkInterval controls how often our internal task loop runs when
+	// in the connecting state. Only useful for testing.
+	checkInterval time.Duration
```
at-wat, line 100: This maybe better to be exposed

at-wat:
let me rebase and merge

transport_test.go
```diff
@@ -41,7 +41,7 @@ func testTimeout(t *testing.T, c *Conn, timeout time.Duration) {
 
 	startedAt := time.Now()
 
-	for cnt := time.Duration(0); cnt <= timeout+defaultTaskLoopInterval; cnt += pollrate {
+	for cnt := time.Duration(0); cnt <= timeout+defaultCheckInterval; cnt += pollrate {
```
at-wat, line 44:
```suggestion
	for cnt := time.Duration(0); cnt <= timeout+defaultKeepaliveInterval; cnt += pollrate {
```

at-wat:
Thank you @nerd2. Rebased and merged as a4a53ab4433bbc210d7b4692d7c569097b71d0ec.

PR #205, opened 2020-06-26
Fix startConnectivityChecks multiple run guard

at-wat:
It should use mutex, otherwise it can enter twice before setting
haveStarted to true.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -56,7 +56,8 @@ type Agent struct {
 	mDNSName string
 	mDNSConn *mdns.Conn
 
-	haveStarted   atomic.Value
+	haveStarted   bool
+	muHaveStarted sync.Mutex
 	isControlling bool
 
 	maxBindingRequests uint16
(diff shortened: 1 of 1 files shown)


agent.go
```diff
@@ -327,14 +327,16 @@ func (a *Agent) startOnConnectionStateChangeRoutine() {
 }
 
 func (a *Agent) startConnectivityChecks(isControlling bool, remoteUfrag, remotePwd string) error {
-	if a.haveStarted.Load().(bool) {
+	a.muHaveStarted.Lock()
+	defer a.muHaveStarted.Unlock()
+	if a.haveStarted {
```
Sean-Der, line 332: Oops messed this up! Would you mind making this a swap instead? Maybe I am over enthuiastic, but I am really tired of debugging mutex/race issues :)

Then I would put the `SetRemoteCredentials` first. If you disagree happy to push forward :)

at-wat: Is it ok to `SetRemoteCredentials` before checking multiple call? I think it should keep old remote credentials which is already running.

Sean-Der: That's a good point, approved!

Issue #203, opened 2020-06-25
Some tests fails on windows due generateRandString uses time.Now().UnixNano()


funvit:
### Your environment.
  - Version: v0.7.15
  - Browser: -

### What did you do?

  go test -run  ^TestLocalCandidateStats  ./...  

### What did you expect?

test run ok

### What happened?

test fails with `invalid stats CandidateType`

---

After adding some more logging to test i'm found, that `generateRandString` generates same id for `hostLocal` and `srflxLocal`.

From stackoverflow post (#57285292):

> The system time has 100 ns precision, but it's only updated by the system timer interrupt every 15 ms or so.

IMHO if not expected to generate id as time-aware - better use completely random generation...

---

Example of more verbose logging:

```
--- FAIL: TestLocalCandidateStats (10.44s)
agent_test.go:974: localStats = [
    {Timestamp:2020-06-26 00:00:10.9252269 +0300 MSK m=+0.224718201 ID:candidate:A13A3AF2-05FE-F256-D5FA-AB6882F3F96A NetworkType:udp4 IP:192.168.1.1 Port:19216 CandidateType:host Priority:2130706431 URL: RelayProtocol:udp Deleted:false} 
    {Timestamp:2020-06-26 00:00:10.9252269 +0300 MSK m=+0.224718201 ID:candidate:A13A3AF2-05FE-F256-D5FA-AB6882F3F96A NetworkType:udp4 IP:192.168.1.1 Port:19217 CandidateType:srflx Priority:1694498815 URL: RelayProtocol:udp Deleted:false}
]
agent_test.go:975: host local stat id=candidate:A13A3AF2-05FE-F256-D5FA-AB6882F3F96A
agent_test.go:976: srflx local stat id=candidate:A13A3AF2-05FE-F256-D5FA-AB6882F3F96A
agent_test.go:993: invalid stats CandidateType, expected host got srflx
```

Sean-Der:
@funvit fantastic catch! If you open a PR, I will work on getting Windows testers in. I did it https://github.com/awslabs/amazon-kinesis-video-streams-webrtc-sdk-c/blob/master/.travis.yml#L170-L177 already


So this should be pretty close to what we want
```
    - name: "Windows Tester"
      os: windows
      script:
        - choco install golang
        - go test -v
```

funvit:
I wonder how to seed rand, due to comment in commit eab497aa25b8e4ebfc7f5830abd608fa31242a05:

> ...
[...]

PR #210, opened 2020-06-27
Fix random generators

at-wat:
Seeding random generator each time limits number of generated sequence
to 31-bits, and causes collision on low time accuracy environments.
Use global random generator seeded by crypto grade random.

### Reference issue
ref: #203


util.go
```diff
@@ -12,6 +14,17 @@ import (
 	"github.com/pion/transport/vnet"
 )
 
+func init() { // nolint:gochecknoinits
```
at-wat, line 17: To call this in the first use of random, any other call of `randSeq` and `generateRandString` should wait until calling `Seed()`, otherwise these parallel calls may return random generated by the default seed.
It makes code complex and may cause future bug, so I believe initializing in `init()` is the best solution.

util.go
```diff
@@ -12,6 +14,18 @@ import (
...
```
at-wat, line 17: To call this in the first use of random, any other call of `randSeq` and `generateRandString` should wait until calling `Seed()`, otherwise these parallel calls may return random generated by the default seed.
It makes code complex and may cause future bug, so I believe initializing in `init()` is the best solution.

util.go
```diff
@@ -12,6 +14,18 @@ import (
...
+	// Seeding random generator each time limits number of generated sequence to 31-bits,
+	// and causes collision on low time accuracy environments.
+	// Use global random generator seeded by crypto grade random.
+	var seed int64
+	if err := binary.Read(cryptorand.Reader, binary.LittleEndian, &seed); err != nil {
+		// crypto/rand is unavailable. Fallback to seed by time.
+		seed = time.Now().UnixNano()
+	}
+	rand.Seed(seed)
```
at-wat, line 26: To avoid touching `math/rand` global generator, we need to declare package global random generator in this package. But in this case, we need to be careful that all code using random must use this generator. If `math/rand` global generator is used, it will be seeded by default value all time. (might be potentially dangerous)

Sean-Der:
Would it be worth banning `math/rand` completely? Maybe a little heavy handed, but better safe then sorry.

util.go
Sean-Der: Can we just not use `math/rand` instead? Not a big fan of `init`

at-wat:
> Would it be worth banning `math/rand` completely?

I don't think so. `crypto/rand` consumes hardware entropy pool and using too much affect all other crypto applications on the same hardware.
Basically, we should select proper random generator for each usage.

In this code of unique ID generation, `crypto/rand` would be suitable, I think.

at-wat:
Wait, `randSeq` is used to generate agent credentials. It must be `crypto/random`.

at-wat:
For mDNS hostname, 

https://tools.ietf.org/id/draft-ietf-rtcweb-mdns-ice-candidates-02.html#gathering
> Generate a unique mDNS hostname. The unique name MUST consist of a version 4 UUID as defined in [RFC4122], followed by “.local”.

It must be UUIDv4 that is crypto grade 122bit random and contain UUID variant and version bits.
I think it's better to use UUID library like `github.com/google/uuid` which is designed and managed by crypto experts.

at-wat:
We currently use UUID like random value for it, but it is defined as
https://tools.ietf.org/html/rfc5245#section-15.1
> ```
>    candidate-attribute   = "candidate" ":" foundation SP component-id SP
>                            transport SP
>                            priority SP
>                            connection-address SP     ;from RFC 4566
>                            port         ;port from RFC 4566
>                            SP cand-type
>                            [SP rel-addr]
>                            [SP rel-port]
>                            *(SP extension-att-name SP
>                                 extension-att-value)
> 
>    foundation            = 1*32ice-char
>    component-id          = 1*5DIGIT
>    transport             = "UDP" / transport-extension
>    transport-extension   = token              ; from RFC 3261
>    priority              = 1*10DIGIT
>    cand-type             = "typ" SP candidate-types
>    candidate-types       = "host" / "srflx" / "prflx" / "relay" / token
>    rel-addr              = "raddr" SP connection-address
>    rel-port              = "rport" SP port
>    extension-att-name    = byte-string    ;from RFC 4566
>    extension-att-value   = byte-string
>    ice-char              = ALPHA / DIGIT / "+" / "/"
> ```

Random part must be 1 to 32 characters of ALPHA, DIGIT, `+`, `/`. Using UUID is out of the standard.

ICE candidate ID is used in plain text in SDP. It doesn't need crypto grade random.

Sean-Der:
> I don't think so. crypto/rand consumes hardware entropy pool and using too much affect all other crypto applications on the same instance.

Looking through your list the only thing I see in the hot path is the VP8 code. If a user is really blocked on generation they can choose to use a weaker version.  

I am worried that we will audit these now, but in the future this will get broken. I am not careful enough and with `goimports` sometimes I will swap between `rand` and `math` accidentally :/

> I think it's better to use UUID library like github.com/google/uuid which is designed and managed by crypto experts.
Works for me! As long as it doesn't drag in a bunch of dependencies/is stable I agree! I would almost consider this 1st party, things on `github.com/google` probably aren't gonna get pulled/break like things under personal repos.

at-wat:
https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate
> ## usernameFragment
> A DOMString containing a randomly-generated username fragment ("ice-ufrag") which ICE uses for message integrity along with a randomly-generated password ("ice-pwd"). You can use this string to verify generations of ICE generation; each generation of the same ICE process will use the same `usernameFragment`, even across ICE restarts.

at-wat:
> I am worried that we will audit these now, but in the future this will get broken. I am not careful enough and with `goimports` sometimes I will swap between `rand` and `math` accidentally :/

Hmm, at the point I studied information science, random generators should be properly selected for each usage to keep both security and performance.
I'm not very sure current crypto grade random generator can replace all random generators. Please let me look into the latest situations.

For now, how about adding a rule to add comment to every `math/rand` import?

Sean-Der:
> For now, how about adding a rule to add comment to every math/rand import?

@at-wat Yea I like that! I think that is the perfect mix of helping us not make mistakes, but giving the flexibility to do what we need

PR #211, opened 2020-06-27
Remove nominationRequestCount

Sean-Der:
Before then nominating a candidate pair we limited the
amount of requests we sent, and would set the connection
to failed.

This change removes that limit, and instead uses the Disconnected+Failed
timer to control this. I found this when debugging an unrelated thing,
and was suprised by the behavior. I didn't find anything in the spec
that mentioned this, and isn't a knob that users will understand how to
control easily.

### DIFF
--- a/selection.go
+++ b/selection.go
@@ -17,17 +17,15 @@ type pairCandidateSelector interface {
 }
 
 type controllingSelector struct {
-	startTime              time.Time
-	agent                  *Agent
-	nominatedPair          *candidatePair
-	nominationRequestCount uint16
-	log                    logging.LeveledLogger
+	startTime     time.Time
+	agent         *Agent
+	nominatedPair *candidatePair
+	log           logging.LeveledLogger
 }
 
 func (s *controllingSelector) Start() {
 	s.startTime = time.Now()
 	s.nominatedPair = nil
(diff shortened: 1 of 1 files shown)


Sean-Der:
@hugoArregui I recently added the `Failed` state to match WebRTC in the browser, and also added `ICE Restarts`.

Going to failed exists outside `selection.go` now, and is consistent across different agent types. If yo are in checking for Disconnected+Failed time it gives up (30 seconds)

I think with everything else we added this change isn't needed anymore? We correctly go to fail, and I think it is ok to sit in checking for a while on this?

Sean-Der:
@at-wat do you mind if I tag `v2.0.0` for `pion/webrtc` ? Or should we investigate any more issues?

at-wat:
@Sean-Der I think it's better to add a parameter for https://github.com/pion/ice/pull/201.
Or we can tag `-rc` until the API gets stable as we did at pion/dtls.

hugoArregui:
> @hugoArregui I recently added the `Failed` state to match WebRTC in the browser, and also added `ICE Restarts`.

I think this is ok, I thought this was part of the standard but I cannot find it so I'm probably confusing it with something else.

PR #214, opened 2020-06-29
Allow gathering srflx candidates from turn: urls

LanderN:
#### Description
This PR enables gathering srflx candidates from `turn:` urls.


Sean-Der:
@LanderN nice catch! I doubled check the RFC and realized my mistake :)

Would you mind adding a link to the RFC somewhere so people understand the change fully?

```
// When TURN servers are utilized, both types of
// candidates are obtained from the TURN server.
//
// https://tools.ietf.org/html/rfc5245#section-2.1
```

LanderN:
Sure thing @Sean-Der, would you add it to `gather.go`, or to the test? Both?

Sean-Der:
@LanderN I think adding to the test is good! If someone wants to undo your change that test will fail, and then they will see why we need this.

Nice work on this :)

PR #222, opened 2020-07-01
Make Transport return addresses of the current pair

cryptix:
Instead of always returning `nil` from `Conn.LocalAddr()` and `Conn.RemoteAddr()` pick the address the selected pair or nil if there is none.

I assume there is a reason for this but I couldn't find it anywhere. Primarily motivation for doing this was having correct addresses when passing them to the quic transport in pion/webrtc.

TODO:
* [x] remove `func getNetAddrFromCandidate`
I ran a few circles around this because I didn't see the unexported `addr() *net.UDPAddr`.

I feel like it should be easier to get this, maybe change `CandidateRelatedAddress` slightly? But since I'm fairly new to this code I wanted to check in first and not mangle it further.

updates pion/quic#31

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Ori Bernstein](https://eigenstate.org)
 * [Sam Lancia](https://github.com/nerd2)
 * [Lander Noterman](https://github.com/LanderN)
+* [Henry](https://github.com/cryptix)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/transport.go
+++ b/transport.go
@@ -115,14 +115,24 @@ func (c *Conn) Close() error {
 
 // TODO: Maybe just switch to using io.ReadWriteCloser?
 
-// LocalAddr is a stub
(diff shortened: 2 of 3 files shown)


Sean-Der:
:fire: this LGTM @cryptix I am all for adding it! 

It is fine to expose anything you need from this repo (since we are satisfying a spec). Things only get harder when we talk WebRTC.

* Would you mind fixing the commit and README.md linter `bash .github/assert-contributors.sh` and `bash .github/lint-commit-message.sh`. You can see everything we run in the `.travis.yml`

* Would you mind adding some simple tests? Just assert this returns `nil` before connected, and `non-nil` after 

Thanks for doing this, excited to have you involved :)

cryptix:
> I am all for adding it!

:tada: 

> Would you mind fixing the commit and README.md 

done.

> Just assert this returns nil before connected, and non-nil after

right, sure. i just need to find a good place for this. I tried it [here](https://github.com/pion/ice/blob/bd57bd814c01ac0534724b2677fcc1d91d7cfb11/transport_test.go#L271) but the connection was already established before the channels were closed... I'm sure I will find another place or whip something up.

Sean-Der:
Merged! Thanks @cryptix, I just added the tests myself. Will be merging stuff for the next few hours :)

PR #176, opened 2020-04-27
Fix IPv6 addresses gathering

buptczq:
Fix isSupportedIPv6 function, when a global unicast address passed to,
the function returns false.

PR #223, opened 2020-07-01
Fix IPv6 addresses gathering

cryptix:
This is just a rebase of @buptczq patch because it seems super useful.

(closes #176)


Sean-Der:
Nice! Thanks for doing this @cryptix

PR #226, opened 2020-07-04
Add support for passive ICE TCP candidates

### LINKED ISSUE #196: TCP Support?
## Summary

https://github.com/pion/webrtc/blob/bd6c5cbc71fc401eac3cd158c97586a79f92d34e/networktype.go#L13

TCP is not yet supported.

## Motivation

Feature completeness and spec compliance.

jeremija:
#### Description
Tested on Firefox and Chrome.

The current implementation creates global TCP listeners on all allowed interfaces when `AgentConfig#NetworkTypes` includes a network with type `tcp4` or `tcp6.

The config option `TCPListenPort` should be used to set a wanted port. When set to `0`, `TCP` listening willl be disabled. This could be changed to randomly select a port.

Note: the [RFC6544](https://tools.ietf.org/html/rfc6544) states there are three TCP candidate types:

- passive (this PR, controlled role)
- active (browsers currently support this, controlling role)
- simultaneous open (both active and passive, not sure what the role would be here since it's both controlled and controlling?)

TODO: check if priority should be lower for TCP Host type candidates.

#### Architecture

`tcpIPMux` keeps a global map of local TCP listeners in `tcpMuxes` variable. Alternatively, we could just have the users pass in a `net.Listener` and then we wouldn't need this class. The main reason `net.ListenTCP` is used at all is to try to keep using the existing config properties (`AgentConfig#NetworkTypes`).

`tcpIPMux` also contains a helper method, `RemoveUfrag`, which closes and removes all existing TCP connections related to a specific `Ufrag`.

`tcpMux` is created for every TCP listener. It accepts connections in a loop and reads the first packet. The first packet MUST be a STUN packet (binding request) with Ufrag or the connection will be closed.

The `Ufrag` is read and the initial packet with the related `net.Conn` is passed to a new or existing instance of `tcpPacketConn` associated with the `Ufrag`.

The `tcpPacketConn` struct implements `net.PacketConn` and can be used for reading from or writing to multiple TCP connections at once.

Since `TCP` protocol is a stream-oriented protocol, each packet must have a 2-byte prefix containing the packet length as described in [RFC 4571](https://tools.ietf.org/html/rfc4571). Also see [Section 10 of RFC 6544](https://tools.ietf.org/html/rfc6544#section-10).

#### Reference issue
Fixes #196

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -121,6 +121,7 @@ type Agent struct {
 	log           logging.LeveledLogger
 
 	net *vnet.Net
+	tcp *tcpIPMux
 
 	interfaceFilter func(string) bool
 
@@ -304,6 +305,12 @@ func NewAgent(config *AgentConfig) (*Agent, error) {
 
 		insecureSkipVerify: config.InsecureSkipVerify,
 	}
+
+	a.tcp = newTCPIPMux(tcpIPMuxParams{
+		ListenPort:     config.TCPListenPort,
+		Logger:         log,
+		ReadBufferSize: 8,
+	})
 
 	if a.net == nil {
 		a.net = vnet.NewNet(nil)
@@ -695,6 +702,15 @@ func (a *Agent) checkKeepalive() {
 
 // AddRemoteCandidate adds a new remote candidate
 func (a *Agent) AddRemoteCandidate(c Candidate) error {
+	// canot check for network yet because it might not be applied
+	// when mDNS hostame is used.
+	if c.TCPType() == TCPTypeActive {
+		// TCP Candidates with tcptype active will probe server passive ones, so
+		// no need to do anything with them.
+		a.log.Infof("Ignoring remote candidate with tcpType active: %s", c)
+		return nil
+	}
+
 	// If we have a mDNS Candidate lets fully resolve it before adding it locally
 	if c.Type() == CandidateTypeHost && strings.HasSuffix(c.Address(), ".local") {
 		if a.mDNSMode == MulticastDNSModeDisabled {
@@ -871,6 +887,7 @@ func (a *Agent) Close() error {
 
 	a.gatherCandidateCancel()
 	a.err.Store(ErrClosed)
+	a.tcp.RemoveUfrag(a.localUfrag)
 	close(a.done)
 
 	<-done
@@ -901,9 +918,9 @@ func (a *Agent) deleteAllCandidates() {
 }
 
 func (a *Agent) findRemoteCandidate(networkType NetworkType, addr net.Addr) Candidate {
-	ip, port, err := addrIPAndPort(addr)
-	if err != nil {
-		a.log.Warn(err.Error())
+	ip, port, _, ok := parseAddr(addr)
+	if !ok {
+		a.log.Warnf("Error parsing addr: %s", addr)
 		return nil
 	}
 
@@ -932,10 +949,17 @@ func (a *Agent) sendBindingRequest(m *stun.Message, local, remote Candidate) {
 
 func (a *Agent) sendBindingSuccess(m *stun.Message, local, remote Candidate) {
 	base := remote
+
+	ip, port, _, ok := parseAddr(base.addr())
+	if !ok {
+		a.log.Warnf("Error parsing addr: %s", base.addr())
+		return
+	}
+
 	if out, err := stun.Build(m, stun.BindingSuccess,
 		&stun.XORMappedAddress{
-			IP:   base.addr().IP,
-			Port: base.addr().Port,
+			IP:   ip,
+			Port: port,
 		},
 		stun.NewShortTermIntegrity(a.localPwd),
 		stun.Fingerprint,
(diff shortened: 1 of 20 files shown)


agent.go
```diff
@@ -121,6 +121,7 @@ type Agent struct {
 	log           logging.LeveledLogger
 
 	net *vnet.Net
+	tcp *tcpIPMux
```
Sean-Der, line 124: What do you think of making `tcpIPMux` always non-nil?

`if a.tcp != nil` kind of scares me. I can see new code getting added that misses that check and would crash in ways we would fail to test.

Sean-Der:
@jeremija amazing work! This is really huge for Pion, this is a big +1 that lots of other media servers and SFUs don't have :D

candidate_host.go
```diff
@@ -19,6 +19,7 @@ type CandidateHostConfig struct {
 	Address     string
 	Port        int
 	Component   uint16
+	TCPType     TCPType
```
Sean-Der, line 22: I worry that a reader would assume ALL `CandidateHost` instances must have a `TCPType`

Do you think it would be obnoxious to have a dedicated TCP Candidate type? Or just some other way to lay it out so users understand `udp OR tcp ... if tcp then one of these types`

tcp_ip_mux.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return m
+}
+
+func (m *tcpIPMux) Remove(key string) {
+	tcpMuxesMu.Lock()
+	defer tcpMuxesMu.Unlock()
+
+	if tcpMux, ok := tcpMuxes[key]; ok {
+		_ = tcpMux.Close()
```
Sean-Der, line 39: Can you log this error?

tcp_ip_mux.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+func (m *tcpIPMux) Listen(ip net.IP) (*tcpMux, error) {
+	tcpMuxesMu.Lock()
+	defer tcpMuxesMu.Unlock()
+
+	key := net.JoinHostPort(ip.String(), strconv.Itoa(m.params.ListenPort))
+
+	if tcpMuxes == nil {
```
Sean-Der, line 59: Can this go in `newTCPIPMux` ?

tcptype.go
```diff
@@ -0,0 +1,14 @@
+package ice
+
+// TCPType is the type of ICE TCP candidate as described in
+// ttps://tools.ietf.org/html/rfc6544#section-4.5
+type TCPType string
+
+const (
+	// TCPTypeActive is active TCP candidate, which initiates TCP connections.
+	TCPTypeActive TCPType = "active"
```
Sean-Der, line 9: I usually do `iota + 1`, the 'no default value' is nice for debugging!

tcp_packet_conn.go
```diff
@@ -0,0 +1,232 @@
...
+
+	recvChan chan streamingPacket
+
+	mu         sync.Mutex
+	wg         sync.WaitGroup
+	closedChan chan struct{}
+	closeOnce  sync.Once
+}
+
+var _ net.PacketConn = &tcpPacketConn{}
```
Sean-Der, line 27: Is this just to assert that it satisfies an interface?

jeremija: Yes

tcptype.go
jeremija: Updated to `int`s

tcp_ip_mux.go
jeremija: Yes, moved!

tcp_ip_mux.go
jeremija: Added logging

candidate_host.go
jeremija: I changed the type to `int` with a default value `TCPTypeUnspecified`.

agent.go
jeremija: Makes sense, changed!

tcp_packet_conn.go
Sean-Der: It should be ok to drop this (I think?) since it is all internal. We don't have to worry about accidentally breaking intreface. Or maybe if you want to keep I would put this in a `_test.go` 

We do this in RTCP, but I think it is kind of confusing.

jeremija: Removed!

Sean-Der:
@jeremija so this is my vote (and only if you agree). I don't want to burn you out and put this in any danger.

What do you think of merging this, but still making changes in a follow up PR?
* Moving this to `internal/tcp`
* Refactoring to use a API that is just ufrag+`PacketConn`

I approved and when it passes feel free to land :)

jeremija:
@Sean-Der thanks! So just to clarify, I think if we modified the code to allow passing of `net.Listener` as a config parameter instead of the port number, things might be a little simpler and we'd be able to remove the global map and the `tcp_ip_listener`. Then we'd only need a function to handle the `net.listener#Accept` and feed the connections to the relevant `tcpPacketConn` based on `ufrag`. WDYT?

Sean-Der:
@jeremija I say we roll with what we have now, and then explore how to get rid of all global state!

I would say a user creates the `ice.TCPMux` (which requires a `net.Conn`) and then passes it into a `webrtc.SettingEngine`. Then the 'global' state all needs to be handled by the user.

Sean-Der:
Ok I am even more excited for this now! I am working on a benchmark for how high we can scale Pion for video delivery (RTSP -> WebRTC bridge) and need to use TCP.

I am gonna spend a few days doing https://github.com/pion/srtp/issues/85, but tell me if there is anything I can do to help unblock you @jeremija :)

PR #227, opened 2020-07-05
Remove candidateSelectionTimeout

Sean-Der:
Remove candidateSelectionTimeout
    
    The code that uses this field was removed in
    e264665676a68107c582a480318880c08f2e19a5 but the field
    wasn't properly removed from the public API.

### DIFF
--- a/agent_config.go
+++ b/agent_config.go
@@ -20,9 +20,6 @@ const (
 	// defaultFailedTimeout is the default time till an Agent transitions to failed after disconnected
 	defaultFailedTimeout = 25 * time.Second
 
-	// timeout for candidate selection, after this time, the best candidate is used
-	defaultCandidateSelectionTimeout = 10 * time.Second
-
 	// wait time before nominating a host candidate
 	defaultHostAcceptanceMinWait = 0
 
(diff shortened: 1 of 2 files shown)


agent_config.go
```diff
@@ -157,12 +157,6 @@ func (config *AgentConfig) initWithDefaults(a *Agent) {
 		a.maxBindingRequests = *config.MaxBindingRequests
 	}
 
-	if config.CandidateSelectionTimeout == nil {
-		a.candidateSelectionTimeout = defaultCandidateSelectionTimeout
-	} else {
-		a.candidateSelectionTimeout = *config.CandidateSelectionTimeout
```
at-wat, line 163: Can `defaultCandidateSelectionTimeout` be remove?
`AgentConfig.CandidateSelectionTimeout` would be better to be commented as deprecated.

Sean-Der: Since we did the API break I think it is ok to kill it!

My concern is that it does nothing anymore :( so if a user is still using it they are going to be surprised it is a no-op I think.

at-wat: I agree to just remove it from v2 API.

at-wat: So,
https://github.com/pion/ice/blob/245c513fed1c946cb5b085fab0e424b28651cce3/agent_config.go#L107-L110 
can be removed.

Sean-Der: done!

Sean-Der:
@at-wat mind if I tag a new version? Want to do a new rc, or `v2.0.0` worthy?

at-wat:
I think it would be better to make a new rc, wait a little while (to ensure there is no other breaking change) and tag v2.0.0 to the same commit.

PR #239, opened 2020-07-11
Set Conn.SetDeadline on candidate close

at-wat:
To ensure recvLoop is unblocked.

### Reference issue
Possibly avoid https://github.com/pion/webrtc/issues/1311 (not sure)

### DIFF
--- a/candidate_base.go
+++ b/candidate_base.go
@@ -99,6 +99,10 @@ func (c *candidateBase) RelatedAddress() *CandidateRelatedAddress {
 
 // start runs the candidate using the provided connection
 func (c *candidateBase) start(a *Agent, conn net.PacketConn, initializedCh <-chan struct{}) {
+	if c.conn != nil {
+		c.agent().log.Warn("Can't start already started candidateBase")
+		return
+	}
 	c.currAgent = a
 	c.conn = conn
 	c.closeCh = make(chan struct{})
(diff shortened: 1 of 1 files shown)


Sean-Der:
Oof nice catch.

This should be effecting non-ice restart as well though right? We block until all candidates are closed, I am surprised we don't see this from other tests.

at-wat:
> This should be effecting non-ice restart as well though right? We block until all candidates are closed, I am surprised we don't see this from other tests.

@Sean-Der Basically, `Close()` should unblock all `Read()`, so this shouldn't change ICE behavior.
In https://github.com/pion/webrtc/issues/1311, blocked routine called `c.conn.Close()` just before blocking point.
```
#	0x931696	github.com/pion/ice/v2.(*candidateBase).close+0xf6		/home/travis/gopath/pkg/mod/github.com/pion/ice/v2@v2.0.0-rc.4/candidate_base.go:177
```
but `c.conn.ReadFrom` still blocks.
```
#	0x6d9f24	net.(*UDPConn).ReadFrom+0x94				/home/travis/.gimme/versions/go1.14.linux.amd64/src/net/udpsock.go:118
```

I think it's one of the below bugs
- bug in our code?
  - if `start()` was called twice, `c.closedCh` is overwritten and old `recvLoop` routine leaks.
- bug in Go net package?
- bug in Linux syscall?

PR #246, opened 2020-07-16
 Select candidate pair when network updated

cnderrauber:
#### Description
When webrtc's network changed(eg. from wifi to 4g)，it will gather ice candidate and send stun bind....., after process, webrtc will send a stun bind with use candidate, but the pion' ice will not set selected candidate pair because it's selectedPair is not nil. So it will not send data to the new pair, the media transport is interrupted.
#### Reference issue
Fixes #...

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Sam Lancia](https://github.com/nerd2)
 * [BUPTCZQ](https://github.com/buptczq)
 * [Henry](https://github.com/cryptix)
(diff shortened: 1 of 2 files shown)


Sean-Der:
Thanks for the CR @cnderrauber but unfortunately we can't accept this :(

[RFC 8445](https://tools.ietf.org/html/rfc8445) doesn't allow us to change selected pairs. 

```
   Once selected pairs have been produced for a data stream, an agent
   MUST send data on those pairs only.
```

The right way to handle this is through a [ice-restart](https://github.com/pion/webrtc/tree/master/examples/ice-restart)

Issue #245, opened 2020-07-15
Continued TCP-ICE Improvements


Sean-Der:
These are a list of all the things we can do to continue improving the `tcp-ice` implementation that landed.
* ~Move to `internal` (a small root package that isn't littered with protocol details is great)~
* [x] Take a `net.Listener` instead of having global state
* [x] ~Expose a `net.PacketConn` based API~ Expose a net.PacketConn based API, `Agent` 
* [x] Implement Active
* [ ] Implement simultaneous-open
* [ ] Explore if we can remove some concurrency from the API/reduce LoC
* [x] Fix ICE TCP Priority. Right now passive TCP host candidates will have higher priorities than `prflx`/`srflx` so TCP will be used even when UDP connectivity is possible

jeremija:
I think the spec is also pretty specific about some situations when the inbound connections need to be dropped. I think currently we'll accept any TCP connection with the right `ufrag`, but when receiving active (inbound) TCP connections, we should only accept the ones for which we already have candidates with the IP address (TCP port is irrelevant in this case because browsers will just set it to `9`).

Also we shouldn't drop the TCP connection during ICE restarts so that's another thing to think about.

jeremija:
We should also provide a mechanism that would allow us to figure out which network type is currently in use by an ICE Agent. For example, there's no need for having a Jitter Buffer when TCP is in use.

Perhaps there could be a function:

```go
func (a *Agent) SelectedNetworkType() (t NetworkType, ok bool` {
  pair := a.getSelectedPair()
  if pair == nil {
    return t, false
  }

  return pair.local.NetworkType(), true
}
```

And then we could expose this information in

```go
func (pc *PeerConnection) SelectedNetworkType() (t NetworkType, ok bool) {
  nt, ok := pc.iceTransport.gatherer.getAgent().SelectedNetworkType()
  if !ok {
    return t, ok
  }

  t, err := NewNetworkType(nt)
  return t, err != nil
}
```

jech:
> func (pc *PeerConnection) SelectedNetworkType() (t NetworkType, ok bool) {

How would the client know when this information becomes invalid and must be recomputed?  Can it only change when ICE transitions to the failed state, or can it change at any time?

[...]

PR #254, opened 2020-07-21
Remove global state for ICE TCP

### LINKED ISSUE #253: Remove global state for ICE TCP
Relates to #245.

jeremija:
This addresses a few points issue of #245:

 - Take a `net.Listener` instead of having global state
 - <strike>Expose a `net.PacketConn` based API, Agent</strike> Expose a `net.TCPMux` based API

Also, the unused `CloseChannel` was removed from `tcp_mux.go`.

Closes #253.


jeremija:
@Sean-Der I know we talked about how we should avoid `nil` values, but I couldn't figure out of a better solution now that `TCPMux` is optional. I'd be happy to hear if you have a solution that allows to never have a `nil` value here.

Sean-Der:
Absolutely amazing, this is so exciting to see a red diff :) couple ideas/points (push back if you disagree)

* I would make `TCPMux` an interface. That way we can mock it out in the future if we need.
* Maybe do a `invalidTCPMux` instead of a nil pointer?  Just so much easier to be resilient when stuff returns an err (instead of segv the whole process)

jeremija:
Sounds good, will try to do it over the weekend! An alternative would be to make the pointer receiver methods check whether the pointer is nil, but I like your solution better.

jeremija:
I think we will probably need a method that would allow us to close a TCP connection eventually. Something like:

```go
func (t *TCPMux) CloseConn(ufrag string, raddr net.Addr)
```

WDYT?

Sean-Der:
This is something a user will need to do?

Right now every gathered candidate gets closed internally, so `ufrag`s shouldn't pile up? (I think)

sorry will check on my laptop soon :)

jeremija:
No worries, I think I'm just getting a little confused :) So when I look at this statement from https://tools.ietf.org/html/rfc6544: 

> When ICE processing for a media stream completes, each agent SHOULD close all TCP connections (that were opened due to this ICE session) except the ones between the candidate pairs selected by ICE.

In the context of our host (listening tcp) candidate, this doesn't mean we have to manually close the remote TCP connections? Is this because these (potentially unselected) inbound TCP connections will actually be handled (closed) internally as remote candidates?

jeremija:
I have addressed both of your points:

> - [x] I would make TCPMux an interface. That way we can mock it out in the future if we need.
> - [x] Maybe do a invalidTCPMux instead of a nil pointer? Just so much easier to be resilient when stuff returns an err (instead of segv the whole process)

and I'm going to merge this since you already approved it. If you have any concerns about I'd be push and fix later. I'll tag this release and push an update to `pion/webrtc`. Excited to finally releasing this feature :)

PR #261, opened 2020-08-08
Do not prematurely unlock mutex in handleRecv

### LINKED ISSUE #260: Panic when closing ICE TCPPacketConn
See https://github.com/pion/webrtc/issues/1355

jeremija:
#### Description
Mutex being unlocked too soon caused a panic. See the issue for more details and a link to `pion/webrtc` issue.

Is there a better way of preventing a write to closed channel than this?

#### Reference issue
Fixes #260

### DIFF
--- a/tcp_packet_conn.go
+++ b/tcp_packet_conn.go
@@ -194,9 +194,10 @@ func (t *tcpPacketConn) removeConn(conn net.Conn) {
 func (t *tcpPacketConn) Close() error {
 	t.mu.Lock()
 
+	var shouldCloseRecvChan bool
 	t.closeOnce.Do(func() {
 		close(t.closedChan)
-		close(t.recvChan)
+		shouldCloseRecvChan = true
 	})
 
 	for _, conn := range t.conns {
@@ -207,6 +208,10 @@ func (t *tcpPacketConn) Close() error {
 	t.mu.Unlock()
 
 	t.wg.Wait()
+
+	if shouldCloseRecvChan {
+		close(t.recvChan)
+	}
 
 	return nil
 }


Sean-Der:
@jeremija The pattern we moved to in `pion/dtls` was only close a channel in the same context it is being written too. https://github.com/pion/dtls/blob/master/conn.go#L816-L823 juggling channels across multiple goroutines is really hard. 

I don't know the TCP code will enough, but happy to dig through and try to reason it out!

jeremija:
Thanks @Sean-Der! Yeah I have found that closing channels after being done with the writing is the best way to go as well. Not sure if that can be applied here. I'll try to wrap my head around this tonight or tomorrow - perhaps there is a better way to design this.

Sean-Der:
jech and at-wat are both super strong with this and probably better to ask questions!

Let me read the code real quick and ask some better questions. I think that will help with some rubber duck debugging :)

tcp_packet_conn.go
```diff
@@ -103,14 +103,13 @@ func (t *tcpPacketConn) startReading(conn net.Conn) {
 
 func (t *tcpPacketConn) handleRecv(pkt streamingPacket) {
 	t.mu.Lock()
+	defer t.mu.Unlock()
```
at-wat, line 106: I think `close(t.closedChan)` should be out of the lock on `Close()` to avoid deadlock if `handleRecv` takes lock during below select.
https://github.com/pion/ice/blob/8f4d021fc7e54cbeb9ac5633c80ef19b92e1a276/tcp_packet_conn.go#L195-L201

IMO, it would be better to close `recvChan` when leaving `startReading()` instead of closing it in `Close()`, and keep unlocking before select.

jeremija: @at-wat Thanks for your review! Just trying to see if we're on the same page: am I correct when I say that the deadlock you're talking about can occur when the following conditions are met:

1. goroutine A receives a packet, it enters `handleRecv`, but the `recvChan` is full, so it's stuck at `select`, and
2. goroutine B calls `Close` but is stuck because the mutex is still acquired in `handleRecv`?

> IMO, it would be better to close recvChan when leaving startReading() instead of closing it in Close(), and keep unlocking before select.

I'm not sure that's possible in this case: multiple connections can be added in `AddConn`, and it is not known when the last one was received. All of their packets are being written to the same receiving channel. If the `recvChan` was closed as soon as one `startReading()` ends, it would cause panics on `write to close channel` for other active connections, or the ones that are yet to be added.

jeremija: I guess we could check for the "closed" state in `AddConn` and prevent adding new connections after `Close` is called, and then close the `recvChan` as soon as all created reading goroutines have ended.

Sean-Der:
@at-wat @jeremija 

I think we could also tag a `v2.0.0` for ICE if you wanted! We could also start tagging beta's for `v3.0.0` for pion/webrtc :)

In ~week I will be more actively evolved again (in the mean time feel free to do w/e you think is right) . My lack of activity isn't because of lack of interest :(

PR #263, opened 2020-08-09
Make TCP ICE candidates have lower preference

### LINKED ISSUE #262: TCP ICE candidates should have lower preference than UDP
Currently their priority is higher thank UDP srflx / prflx, which results with TCP prflx -> TCP passive host candidates being used before UDP prflx / UDP srflx pairs.

Related to #245.

jeremija:
Closes #262.

Adds the changes in determining local preference for ICE TCP candidates according to RFC 6544, section 4.2.

https://tools.ietf.org/html/rfc6544#section-4.2

The only thing I'm not sure about is the value of `otherPref`, which is currently always set to `8191`, but the RFC says the following:

> If any two candidates have the same type-preference and direction-pref, they MUST have a unique other-pref.  With this specification, this usually only happens with multi-homed hosts, in which case other-pref is the preference for the particular IP address from which the candidate was obtained.  When there is only a single IP address, this value SHOULD be set to the maximum allowed value (8191).

### DIFF
--- a/candidate_base.go
+++ b/candidate_base.go
@@ -90,6 +90,72 @@ func (c *candidateBase) Component() uint16 {
 
 // LocalPreference returns the local preference for this candidate
 func (c *candidateBase) LocalPreference() uint16 {
+	if c.NetworkType().IsTCP() {
+		// RFC 6544, section 4.2
+		//
+		// In Section 4.1.2.1 of [RFC5245], a recommended formula for UDP ICE
+		// candidate prioritization is defined.  For TCP candidates, the same
+		// formula and candidate type preferences SHOULD be used, and the
+		// RECOMMENDED type preferences for the new candidate types defined in
+		// this document (see Section 5) are 105 for NAT-assisted candidates and
+		// 75 for UDP-tunneled candidates.
+		//
+		// (...)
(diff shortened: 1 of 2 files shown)


at-wat:
> The only thing I'm not sure about is the value of `otherPref`, which is currently always set to `8191`, but the RFC says the following:
> 
> > If any two candidates have the same type-preference and direction-pref, they MUST have a unique other-pref.  With this specification, this usually only happens with multi-homed hosts, in which case other-pref is the preference for the particular IP address from which the candidate was obtained.  When there is only a single IP address, this value SHOULD be set to the maximum allowed value (8191).

When a network interface `eth0` has two IP addresses e.g. `192.168.0.10` and  `192.168.0.11` (I have such configuration on my laptop for some testing usages), current pion implementation sets same preference value to the two candidates for these two addresses. It makes selection between two addresses completely random.
I think we should set `otherPref` of `8191` for `192.168.0.10` and  `8190` for `192.168.0.11` for example to make selection result deterministic.
It's out of this PR's scope and not very significant, I think it can be left as is.

jeremija:
Thanks for reviewing and explaining!

PR #270, opened 2020-09-02
Add ICE Candidate Marshal/Unmarshal

korymiller1489:
Refactor ICECandidate logic from `pion/webrtc` to `pion/ice`

Still need to discuss best way to implement logic cleanup in `pion/webrtc` to maintain backwards capability. For now, this PR only adds the logic to `pion/ice` so that it may be used standalone without including the entire webrtc package


Sean-Der:
I am not sure what the right way to do this, but this feels really messy.

We are trying to do the same thing across 3 packages. Maybe we should 
* delete everything ICE out of `pion/sdp`
* do parsing/serialization of a ICE Candidate in `pion/ice`
* `pion/webrtc` should pass everything into here.

korymiller1489:
This makes sense to me. I will try and go down this path. Thanks!

Sean-Der:
This is going to be a little more work, but not untenable.

We don't want a `ICECandidate` and `Candidate`. I think you want to extend the `Candidate` interface to have a `Marshal` and `Unmarshal` function? That means if you are just running a ICE Agent you can pass strings around and everything will work!

Then in `pion/webrtc` we will need to keep some code around to take `ice.Candidate` -> `webrtc.ICECandidate` and `ice.Candidate -> webrtc.ICECandidateInit` we are REALLY close to getting this done so don't lose hope on this. If you feel stuck please grab me and I can grab parts that are frustrating you.

korymiller1489:
Sounds good, I will give it another go

korymiller1489:
Not completely done yet, but I think this is the path you were wanting to take?

Sean-Der:
Yes! Perfect that is exactly it :) 

Then `pion/webrtc` can just use `Marshal` and `Unmarshal` and it has everything it needs. And you don't have to import `pion/webrtc` anymore! Sorry this is such a big task, but this cleanup has been a long time coming.

korymiller1489:
@Sean-Der I think this is ready to be pulled, still WIP but functional without breaking backwards compatibility

Sean-Der:
Merged with 166df5842e51785dba1352fa3ba9fea14deebf15

Nice work! I just brought over the lost error handling from `pion/sdp`

PR #287, opened 2020-10-08
Support proxy dialer

### LINKED ISSUE #284: ice connecting through corporate proxy
## Summary

Pion ICE cannot create a connection to turn server if peers are behind a Corporate Proxy.

## Motivation

When user configure a turn server in ICEServer configuration using  transport tcp, e.g. `turn:<my-server>:3478?transport=tcp`. The ICE is not able to resolve the TCP address or Dial the turn server if the peers are behind a corporate proxy.
The ICE package uses net.DialTCP which is not aware of the proxy settings hence the dial and connection fails.

User should be able to tell/configure the ICE package that he/she is behind a proxy.
The Ice package implements a Dial method, preferly using "golang.org/x/net/proxy" to dial the turn server via the proxy.

## Additional context

We can help implement this if you see above proposal as an acceptable solution.

assadobaid:
#### Description
When Running pion webRTC behind a http or https proxy e.g a corporate proxy Dialing to turn server fails. This fix enables user to implement golang.org/x/net/proxy Dial interface and provide it to the ICE package via settingEngine.

#### Reference issue
Fixes #284

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [JooYoung Lim](https://github.com/DevRockstarZ)
 * [Kory Miller](https://github.com/korymiller1489)
 * [ZHENK](https://github.com/scorpionknifes)
-* [assadobaid] (https://github.com/assadobaid)
+* [Assad Obaid](https://github.com/assadobaid)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [JooYoung Lim](https://github.com/DevRockstarZ)
 * [Kory Miller](https://github.com/korymiller1489)
 * [ZHENK](https://github.com/scorpionknifes)
+* [assadobaid] (https://github.com/assadobaid)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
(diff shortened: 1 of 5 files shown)


Sean-Der:
This LGTM, nice work! I will merge this tonight.

Sean-Der:
@assadobaid I am learning the proxy stuff, but how does it know to handle DTLS, TCP and TLS differently? Do we need to allow people to specify a proxy for each connection type?

Sean-Der:
Did you test this out yourself and works for you? Are you able to create DTLS/TCP/TLS allocations?

Sean-Der:
If you have the bandwidth I think we need to get tests up for this? Do you know how hard it would be to have a simple proxy?

assadobaid:
Thank you @Sean-Der for your comments #287 
We have tested the TCP and we do have in our backlog to implement and test TLS, it is part of our project. The implementation of the interface for TLS should not be much different. E.g (gorilla/websocket) websocket.DefaultDialer.Dial do a very similar solution. 

> Do we need to allow people to specify a proxy for each connection type?

Our idea is to have one generic method and from the networktype/protocall let user to decide what kind of dialer to implement and what kind of connection to make. I guess user knows the requirement for the connection type and implement the dialer based on it.
 If you think it would be better with different dialers for each case I can add it,
e.g.
`TCPDialer proxy.Dialer`
`TLSDialer proxy.Dialer`
 `DTLSDialer proxy.Dialer`

Another solution would be to only add for TCP and TLS but not for DTLS as many corporates proxies do not allow UDP data at all to pass through.

Sean-Der:
Ah fantastic, thanks so much for the quick response. I went and looked at the proxy API and it is super powerful (can do per host/custom behavior) so I get how it all works now.

 Ok no more questions from me, I think this is perfect for merging. Thanks for adding the test :)

Sean-Der:
Merged! 64a2cf0c473c3e860728f9aa5be6c9f64bb162fc

I rewrote the test a bit, just to assert that `Dial` was actually executed. Great work on this :)

PR #325, opened 2021-01-15
Fixed #197

### LINKED ISSUE #197: Crasher in Pion when logging at DEBUG or TRACE level. ICE Candidate Pair String()
### Your environment.
Version:
github.com/pion/rtp v1.5.5
github.com/pion/webrtc/v2 v2.2.15

Browser: N/A

Other Information -

### What did you do?
Turned on debug logging at the TRACE level.
ENV setting:
PION_LOG_TRACE=all

### What did you expect?
For the program to run as before, only with logging.

### What happened?
When the connection was shutdown by the remote side, the logger crashed Pion. 

It looks like the String() function in the candidatepair.go file assumes that 'p' will not be nil. But I think one part went away as the remote connection died. 

Screen Shot 2020-06-15 at 4 02 53 PM

There is other code that checks whether p is nil before continuing in that same file. Seems an easy fix.

--Tom

wawesomeNOGUI:
Added a check in case p has no value. I just returned the string "nil" if there is no value (e.g. connection loss).

#### Reference issue
Fixes #197

### DIFF
--- a/candidatepair.go
+++ b/candidatepair.go
@@ -26,8 +26,12 @@ type candidatePair struct {
 }
 
 func (p *candidatePair) String() string {
-	return fmt.Sprintf("prio %d (local, prio %d) %s <-> %s (remote, prio %d)",
+	if p != nil {
+	  return fmt.Sprintf("prio %d (local, prio %d) %s <-> %s (remote, prio %d)",
 		p.Priority(), p.local.Priority(), p.local, p.remote, p.remote.Priority())
+	}else{
+	  return "nil"
+	}
 }
 
 func (p *candidatePair) Equal(other *candidatePair) bool {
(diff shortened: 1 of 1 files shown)


candidatepair.go
```diff
@@ -26,8 +26,12 @@ type candidatePair struct {
 }
 
 func (p *candidatePair) String() string {
-	return fmt.Sprintf("prio %d (local, prio %d) %s <-> %s (remote, prio %d)",
+	if p != nil {
+	  return fmt.Sprintf("prio %d (local, prio %d) %s <-> %s (remote, prio %d)",
 		p.Priority(), p.local.Priority(), p.local, p.remote, p.remote.Priority())
+	}else{
+	  return "nil"
```
Sean-Der, line 33: woot, excited to see you burn through these :) 

* Would you mind adding a test? Just call `String()` on a `var nilCandidatePair *candidatePair` in a test.
* I would just return ""

wawesomeNOGUI: Cool! I edited to return "" if the interface is nil. I'll add in the call to the function to test it out!

wawesomeNOGUI: Oh whoops that's not an interface:D

wawesomeNOGUI: Did I add a test correctly?

Sean-Der:
Merged with c2e6f947d064e02436f5f6345619983baf018ce3 Thanks for doing this @wawesomeNOGUI 

I moved your `main` function into a test and added you to the README.

PR #328, opened 2021-01-24
Cancel Previous Gathering Routine On Restart

### LINKED ISSUE #280: ICE Agent can not be restarted when gathering
In recent versions of the Unnamed SFU, the client will restart ICE whenever the connection's ICE state goes into failed.  I'm occasionally getting the following error:

    ICE Agent can not be restarted when gathering

It happens somewhere in the code that handles incoming offers; I'd need to add more detailed logging in order to know which exact function triggers the error.  How do I avoid this problem?

wawesomeNOGUI:
#### Description

Updated ICE agent restart method to cancel the current gathering and then restart the candidate gathering process.

#### Reference issue

Fixes #280

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1196,8 +1196,10 @@ func (a *Agent) SetRemoteCredentials(remoteUfrag, remotePwd string) error {
 // Restart restarts the ICE Agent with the provided ufrag/pwd
 // If no ufrag/pwd is provided the Agent will generate one itself
 //
-// Restart must only be called when GatheringState is GatheringStateComplete
-// a user must then call GatherCandidates explicitly to start generating new ones
+// If there is a gatherer routine currently running, Restart will
+// cancel it.
+// After a Restart, the user must then call GatherCandidates explicitly
+// to start generating new ones.
 func (a *Agent) Restart(ufrag, pwd string) error {
 	if ufrag == "" {
 		var err error
@@ -1224,8 +1226,7 @@ func (a *Agent) Restart(ufrag, pwd string) error {
 	var err error
 	if runErr := a.run(a.context(), func(ctx context.Context, agent *Agent) {
 		if agent.gatheringState == GatheringStateGathering {
-			err = ErrRestartWhenGathering
-			return
+			agent.gatherCandidateCancel()
 		}
 
 		// Clear all agent needed to take back to fresh state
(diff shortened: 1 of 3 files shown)


wawesomeNOGUI:
I haven’t gotten to test this yet, I’ll test it in a bit

examples/ping-pong/main_RestartTest.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	remoteUfrag := <-remoteAuthChannel
+	remotePwd := <-remoteAuthChannel
+
+	if err = iceAgent.GatherCandidates(); err != nil {        //start gathering candidates
+		panic(err)                                        //will only find local because no STUN server
+	}
+
+	if err = iceAgent.Restart(localUfrag, localPwd); err != nil {
```
wawesomeNOGUI, line 140: I tested the update to Restart by just adding a few calls to it. I'm not sure if this test is perfect, but I have confirmed that the Restart function can be called while there is a gatherer running and the peers can still connect. (I just added: a `fmt.Println("cancled")` after calling `agent.gatherCandidateCancel` in my testing to confirm it ran.)

examples/ICERestart/main_RestartTest.go
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        remoteUfrag := <-remoteAuthChannel
+        remotePwd := <-remoteAuthChannel
+
+        if err = iceAgent.GatherCandidates(); err != nil { //start gathering candidates
+                panic(err) //will only find local because no STUN server
+        }
+
+        if err = iceAgent.Restart(localUfrag, localPwd); err != nil {
```
wawesomeNOGUI, line 140: I tested the update to Restart by just adding a few calls to it in the example/ping-pong/main.go. I'm not sure if this test is perfect, but I have confirmed that the Restart function can be called while there is a gatherer running and the peers can connect. (I just added: a fmt.Println("cancled") after calling agent.gatherCandidateCancel in agent.go in my testing to confirm it ran.)

wawesomeNOGUI: Actually, I just updated the agent_test.go instead, so we probably don't need this file.

stv0g (approved):
LGTM.

I've squashed and rebased the changes to the latest master.

PR #343, opened 2021-03-28
Fix ComponentRTCP value

Gaukas:
#### Description

The value of `ComponentRTCP` was set to 1 which has possibly resulted in the same priority for RTP and RTCP candidates and potentially more problems.

```
a=candidate:3031764619 1 udp 2130706431 192.168.24.132 60513 typ host
a=candidate:3031764619 2 udp 2130706431 192.168.24.132 60513 typ host
a=candidate:233762139 1 udp 2130706431 172.17.0.1 33110 typ host
a=candidate:233762139 2 udp 2130706431 172.17.0.1 33110 typ host
a=candidate:940760967 1 udp 1694498815 MY.PUBLIC.IP.ADDR 63447 typ srflx raddr 0.0.0.0 rport 49823
a=candidate:940760967 2 udp 1694498815 MY.PUBLIC.IP.ADDR 63447 typ srflx raddr 0.0.0.0 rport 49823
```

#### Note
I'm not sure if it is left intentionally or if I am having any misconceptions.

### DIFF
--- a/candidate.go
+++ b/candidate.go
@@ -7,13 +7,12 @@ import (
 )
 
 const (
-	receiveMTU             = 8192
-	defaultLocalPreference = 65535
-
 	// ComponentRTP indicates that the candidate is used for RTP
-	ComponentRTP uint16 = 1
+	ComponentRTP uint16 = 1 + iota
 	// ComponentRTCP indicates that the candidate is used for RTCP
 	ComponentRTCP
+	receiveMTU             = 8192
+	defaultLocalPreference = 65535
 )
 
 // Candidate represents an ICE candidate


Sean-Der:
Nice catch! It is wrong that we are emitting both RTP and RTCP. Since we do BUNDLE we should just do one.

This is what Chrome emits
```
a=candidate:2795255774 1 udp 2122260223 192.168.1.7 58135 typ host generation 0 network-id 1 network-cost 10
a=candidate:264484875 1 udp 1686052607 23.252.60.236 58135 typ srflx raddr 192.168.1.7 rport 58135 generation 0 network-id 1 network-cost 10
a=candidate:3894397742 1 tcp 1518280447 192.168.1.7 9 typ host tcptype active generation 0 network-id 1 network-cost 10

```

Gaukas:
Cool. I guess the RTCP value still needs to be patched somehow, also the behavior of the candidate generation. AFAIK the `data-channel` examples in `pion/webrtc` generate both RTP and RTCP. Please feel free to close this pull request!

Sean-Der:
@Gaukas you can remove the `RTCP` candidates entirely. We only need to generate each if we are doing non-BUNDLE.

Try removing RTCP candidates and Chrome/FireFox and Pion WebRTC should all still work!

Gaukas:
Yes, you're right. I found that when I was trying to create an instance with minimal SDP (trying to eliminate the signaling process). To be used in circumventing use cases...

PR #352, opened 2021-04-19
Fix #351 UDPMux read/write race condition

### LINKED ISSUE #351: UDPMux has race in Read/Write
We currently fill a `packetio.Buffer` with a list of address+message. In memory it looks like `|address|message|address|message|.... `

Currently it is possible for a Read to come in before a Write has completed, so the buffer looks like `|address|`.

We have a few options 
* make the Reader/Writer take a lock
* encode address+message in one vector
* use a packetio.Buffer per address

davidzhao:
Improvements to interface flexibility, and IPv6 compatibility

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -932,17 +932,29 @@ func (a *Agent) deleteAllCandidates() {
 	}
 }
 
-func (a *Agent) findRemoteCandidate(networkType NetworkType, addr net.Addr) Candidate {
+func (a *Agent) findRemoteCandidate(localNetworkType NetworkType, addr net.Addr) Candidate {
 	ip, port, _, ok := parseAddr(addr)
 	if !ok {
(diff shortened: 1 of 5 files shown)


davidzhao:
Hey @Sean-Der, take a look specifically at https://github.com/pion/ice/pull/352/files#diff-07afbb40d3dc2699797d28c0689ba544f67ca642d0d3efcff3c02a0bbb3604fdR952

It's an attempt at fixing some strange issues that I had ran into; where the outgoing/incoming ports through the mux didn't match remote candidates. I believe this is due to `findRemoteCandidate` assuming all remote candidates are IPV4. Let me know your thoughts.

Sean-Der:
Hey @davidzhao 

I pulled in all the Mux changes (great work!) 

I don't understand the IPv6 issue fully so I didn't take that yet. Let me read into that a bit more :)

PR #354, opened 2021-04-20
Generalize UDP and TCP Mux

### LINKED ISSUE #350: Generalize Mux
I think we can have UDP and TCP Mux be the same interface. This would allow users to just pass a list of Mux instances.

I think the pion/webrtc API is great as is though, less confusing for users.

Sean-Der:
Refactor UDPMux to match TCPMux patterns. The goal is to have a
collection of mux instances and no specific UDP/TCP code

Resolves #350

### DIFF
--- a/udp_mux.go
+++ b/udp_mux.go
@@ -2,11 +2,9 @@ package ice
 
 import (
 	"errors"
-	"fmt"
 	"io"
 	"net"
 	"os"
-	"strings"
 	"sync"
 
 	"github.com/pion/logging"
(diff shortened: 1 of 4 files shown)


udp_mux.go
```diff
@@ -151,9 +134,6 @@ func (m *UDPMuxDefault) Close() error {
 		m.mu.Lock()
 		defer m.mu.Unlock()
 
-		// close udp conn and prevent packets coming in
-		err = m.udpConn.Close()
```
davidzhao, line 155: I think we need to stop connWorker after Close is called. We shouldn't continue reading from it after it's closed.

udp_mux.go
```diff
@@ -214,7 +194,7 @@ func (m *UDPMuxDefault) connWorker() {
 	}()
 	buf := make([]byte, receiveMTU)
 	for {
-		n, addr, err := m.udpConn.ReadFrom(buf)
+		n, addr, err := m.params.UDPConn.ReadFrom(buf)
 		if err != nil {
```
davidzhao, line 198:
```suggestion
		if m.IsClosed() {
			return
		}
		if err != nil {
```

udp_mux.go
Sean-Der: Fixed!

udp_mux.go
Sean-Der: Took this + squashed. Thanks @davidzhao

davidzhao (approved):
looks good!

PR #361, opened 2021-04-22
Use os.IsTimeout instead of os.ErrDeadlineExceeded

### LINKED ISSUE #360: No longer compiles with Go 1.13
Ice no longer compiles with Go 1.13.  This is important for Galène, since some distributions still use 1.13.

```
$ /usr/lib/go-1.13/bin/go test ./...
# github.com/pion/ice/v2
../../../../pkg/mod/github.com/pion/ice/v2@v2.1.4/udp_mux.go:216:22: undefined: os.ErrDeadlineExceeded
```

jech:
This fixes compilation with Go 1.13 and 1.14.

Fixes #360.


davidzhao (approved):
sorry about the breakage. this looks good

PR #364, opened 2021-04-27
Fix UDPMux logger

cgojin:
If no logger is passed in, it will be created.

Error running [ice-single-port](https://github.com/pion/webrtc/tree/master/examples/ice-single-port) :
invalid memory address or nil pointer dereference


Sean-Der (approved):

Sean-Der:
Nice catch, thanks @cgojin :)

cgojin:
> Nice catch, thanks @cgojin :)
Like you friendly, Thanks. :)

PR #366, opened 2021-05-04
Obey Transport protocol

mloehnig:
#### Description
The purpose of this PR is to add a check for transportation protocol before attempting to connect via UDP to server. 

Using the latest version of `coturn` implementation with the `no-udp` flag specified and a strict firewall preventing UDP traffic one would observe that the ice candidate gathering takes 5 seconds. After some investigation I found that the `transport=tcp` flag was not being obeyed within the `gatherCandidatesSrflx` function. 

### Related issue
https://github.com/pion/ice/issues/367

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -350,7 +350,7 @@ func (a *Agent) gatherCandidatesSrflx(ctx context.Context, urls []*URL, networkT
 		for i := range urls {
 			// if transport protocol is TCP assume UDP calls will fail/timeout.
 			if urls[i].Proto == ProtoTypeTCP {
-				return
+				continue
 			}
 			wg.Add(1)
 			go func(url URL, network string) {


Sean-Der:
Fantastic work @swagggpickle this looks right to me. I will double check some stuff and then merge this ASAP :) 

Mind adding yourself to the README?

swagggpickle:
@Sean-Der Sure thing I will add myself to the README. I will also add to the unit test to ensure coverage.

gather.go
```diff
@@ -352,6 +352,11 @@ func (a *Agent) gatherCandidatesSrflx(ctx context.Context, urls []*URL, networkT
 			go func(url URL, network string) {
 				defer wg.Done()
 
+				// if transport protocol is TCP assume UDP calls will fail/timeout.
+				if url.Proto == ProtoTypeTCP {
```
zyxar, line 356: if `url.Proto == ProtoTypeTCP`, doesn't even need to start this go routine?

swagggpickle: This is true, I will make an update to the PR.

swagggpickle:
@Sean-Der Hi, Do you have a status update?

PR #380, opened 2021-07-19
Populate RelayProtocol for relay candidates

meelapshah:
#### Description

Candidates have a field `RelayProtocol` on their stat as defined [here](https://www.w3.org/TR/webrtc-stats/#dom-rtcicecandidatestats-relayprotocol):
> It is the protocol used by the endpoint to communicate with the TURN
> server. This is only present for local candidates. Valid values are
> "udp", "tcp", or "tls".

Currently, the field is populated with "udp" for all local and remote candidates of all types. Instead, we should leave it blank for all but local relay candidates in which case we should populate it with the protocol used to talk to the TURN server.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -28,6 +28,7 @@ korymiller1489 <kmiller@unwiredrevolution.com>
 Kyle Carberry <kyle@carberry.com>
 Lander Noterman <lander.noterman@basalte.be>
 Luke Curley <kixelated@gmail.com>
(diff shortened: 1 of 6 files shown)


Sean-Der:
Great PR! Thank you @meelapshah

Sean-Der (approved):

Sean-Der:
@meelapshah should I push a new tag and WebRTC version?

meelapshah:
Yes please, that would be helpful. Thank you!

PR #392, opened 2021-10-09
Implement Active TCP-ICE

dgunay:
#### Description

Implementing Active TCP ICE to enable Pion to connect to passive TCP ICE hosts.

Still relatively new to WebRTC and ICE, so if I stray from [RFC 6544](https://datatracker.ietf.org/doc/html/rfc6544) or otherwise misunderstand something fundamental about the way this should work please don't hesitate to let me know.

#### Reference issue

#245 - Continued ICE improvements

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -130,6 +130,10 @@ type Agent struct {
 	insecureSkipVerify bool
 
 	proxyDialer proxy.Dialer
+
+	tcpConnections map[string]net.Conn
+
+	tcpActive bool
 }
 
 type task struct {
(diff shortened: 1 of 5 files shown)


Sean-Der:
What you will actually want I believe is one `CandidateHost` that contains a TCP Socket.

When `AddICECandidate` comes in with a passive candidate you send off a connection request. 

You will need a `recvLoop` per connection attempt. So I think overriding it completely makes sense!

dgunay:
Voluntarily closing this in favor of #394 - hopefully someone else with the desire or the need will help get it across the finish line.

PR #397, opened 2021-11-16
Update UDPMux to listen on all interfaces 

hex-42:
Before we would only listen on the first. If the first interface isn't the default for the host no traffic would be received.


Sean-Der (approved):

Sean-Der:
Thank you so much for the PR @hex-42 

appreciate you contributing to Pion. If there is anything I can do to make it better I would love to hear :)

PR #403, opened 2021-12-12
WIP: Expand UDP Single Port mode (UDPMux) to ServerReflexive and Relay

### LINKED ISSUE #400: Expand UDP Single Port mode (UDPMux) to all Candidate types
To start off :) 
**A big thank you to the Pion contributors and especially to @Sean-Der for supporting us on this topic.**

## Summary

Expand [Candidate gathering mechanism](https://github.com/pion/ice/blob/72e6a2bab055fd609f18fc430ce959a6137a62e7/gather.go#L82) allowing [UDPMux](https://github.com/pion/ice/blob/master/udp_mux.go) for all the candidate types. 

## Motivation

We use Pion ICE implementation in [Wiretrustee](https://github.com/wiretrustee/wiretrustee) to negotiate direct connections between peers. 
The connections are then used to establish Wireguard tunnels.

- For each of the Wiretrustee peers, we create a new ice.Agent. 
- There can be hundreds or even thousands of peer connections on a single machine.

Using the ICE package with a large number of agents may cause a "Too Many Open files" issue due to the number of open ports for different types of candidates. This may cause issues on systems with default limits configuration. Also, this is kind of wasteful in our case.
We also understand that our use case is not a traditional WebRTC case, probably most of the library users won't have such an issue.

We have discovered that PIon [supports a single UDP port](https://github.com/pion/webrtc/releases/tag/v3.1.0). This is achieved by using [UDPMux](https://github.com/pion/ice/blob/master/udp_mux.go). 
However, it only works for host candidates, srflx and relay candidates are working traditionally.  

**We would like to propose the extension of the candidate gathering mechanism to support all types of candidates when using UDPMux.** 
We also think it won't harm the existing functionality but instead might be very helpful for the use cases like ours. 

We considered terminating the open port from unselected candidates as well, which seems to be a good option to minimize the problem when not using single-port mode.
However, we will still end up having N sockets for N peers connected which is also not the best option for our use case.

Any help on this is highly appreciated!

## Additional context

[UPDMux usage when gathering local candidates ](https://github.com/pion/ice/blob/72e6a2bab055fd609f18fc430ce959a6137a62e7/gather.go#L138)

[Default Srflx Candidates gathering with creating a new listener](https://github.com/pion/ice/blob/72e6a2bab055fd609f18fc430ce959a6137a62e7/gather.go#L357)

mlsmaycon:
#### Description
This PR adds support to ServerReflexive and Relay candidate types.
* check if a.udpMux exists and calls a.udpMux.GetConn(a.localUfrag) to create conn
* check conn is udpMux before closing it

#### Reference issue
Fixes #400


mlsmaycon:
I've tested and when using higher number agents with type Relay and udpMux as I've implemented, the connections fail and I am a bit stuck here. The errors and warnings I see are:

``` 
ice INFO: 2021/12/12 15:07:17 Setting new connection state: Checking
ice WARNING: 2021/12/12 15:07:17 pingAllCandidates called with no candidate pairs. Connection is not possible yet.
ice WARNING: 2021/12/12 15:07:17 Conn is not allocated (could not get server reflexive address udp4 stun:stun.wiretrustee.com:5555: failed to get XOR-MAPPED-ADDRESS response: attribute not found
)
ice WARNING: 2021/12/12 15:07:17 pingAllCandidates called with no candidate pairs. Connection is not possible yet.
ice WARNING: 2021/12/12 15:07:17 discard message from (116.203.205.21:5555), attribute not found
ice WARNING: 2021/12/12 15:07:17 pingAllCandidates called with no candidate pairs. Connection is not possible yet.
ice WARNING: 2021/12/12 15:07:17 discard message from (135.181.92.77:5555), attribute not found
ice WARNING: 2021/12/12 15:07:17 Conn is not allocated (could not get server reflexive address udp4 turn:turn.wiretrustee.com:5555?transport=udp: failed to get XOR-MAPPED-ADDRESS response: attribute not found
)
ice WARNING: 2021/12/12 15:07:21 Discarded message from 192.168.64.1:50505, not a valid remote candidate
ice WARNING: 2021/12/12 15:07:21 Discarded message from 192.168.0.113:50505, not a valid remote candidate
ice WARNING: 2021/12/12 15:07:21 Discarded message from 192.168.0.113:50505, not a valid remote candidate
ice WARNING: 2021/12/12 15:07:21 Discarded message from 192.168.64.1:50505, not a valid remote candidate
ice WARNING: 2021/12/12 15:07:21 Conn is not allocated (Failed to allocate on turn.Client turn.wiretrustee.com:5555 attribute not found
)
ice WARNING: 2021/12/12 15:07:21 Discarded message from 192.168.64.1:50505, not a valid remote candidate
ice WARNING: 2021/12/12 15:07:22 Conn is not allocated (Failed to allocate on turn.Client turn.wiretrustee.com:5555 all retransmissions failed for Sj6djaaKyPbxnBT/
)
ice WARNING: 2021/12/12 15:07:22 Discarded message from 192.168.0.113:50505, not a valid remote candidate
ice WARNING: 2021/12/12 15:07:22 Discarded message from 192.168.64.1:50505, not a valid remote candidate
ice WARNING: 2021/12/12 15:07:22 Conn is not allocated (Failed to allocate on turn.Client turn.wiretrustee.com:5555 all retransmissions failed for 1jqqZmYYs4LWvIWL
)
```

mlsmaycon:
When using multiple agents sharing the same port, packets are being dropped or routed to the wrong agents, it seems like I will have to work on UDPMuxDefault address and connection mappings. This is kind of tricky as STUN Binding Response doesn't carry much information that would allow for good mapping. The same goes for TURN protocol.

gather_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build !js
```
stv0g, line 1: I am a bit confused about this build tag comment here. Whats the purpose of the `// +build` comment?

stv0g:
Hi @mlsmaycon,

thanks for your great work on this :)

> When using multiple agents sharing the same port, packets are being dropped or routed to the wrong agents, it seems like I will have to work on UDPMuxDefault address and connection mappings. This is kind of tricky as STUN Binding Response doesn't carry much information that would allow for good mapping. The same goes for TURN protocol.

I guess we could simply use the [STUN transaction ID](https://datatracker.ietf.org/doc/html/rfc8489#section-5) to facilitate the mapping?

As TURN just uses STUN methods, the packets send for TURN bindings also carry a STUN transaction ID.
So this should also be possible with TURN, I guess?

gather_test.go
mlsmaycon: Just to keep compatibility with older versions as the new directive was introduced in Go v1.7

mlsmaycon:
> Hi @mlsmaycon,
> 
> thanks for your great work on this :)
> 
> > When using multiple agents sharing the same port, packets are being dropped or routed to the wrong agents, it seems like I will have to work on UDPMuxDefault address and connection mappings. This is kind of tricky as STUN Binding Response doesn't carry much information that would allow for good mapping. The same goes for TURN protocol.
> 
> I guess we could simply use the [STUN transaction ID](https://datatracker.ietf.org/doc/html/rfc8489#section-5) to facilitate the mapping?
> 
> As TURN just uses STUN methods, the packets send for TURN bindings also carry a STUN transaction ID. So this should also be possible with TURN, I guess?

Thanks, @stv0g for your comments. 

I've tested a bit with Transactions ID and it could help us address the above issue but not all. 

I think we would need to work on the pion/turn package as well because packets sent from it wouldn't be tracked on ICE and a response from turn ends up going to the last udpMuxedConn that mapped the TURN server's IP. 

Also, mapping Transactions ID will require constant clean-up of the mapped IDs that weren't used (happens on check packets). Another point that I missed when creating the PR, was that using a single ufrag for creating the udpMuxedConn might also lead to responses being handled by the wrong candidates.

@braginini is working on this PR as well.

braginini:
@stv0g @mlsmaycon 
I think that the simplest approach would be to have a single UDPMux per candidate type - UDPMux for Host, UDPMuxSrflx for Server Reflexive, and UDPMuxRelay. 
This could be the first step. I actually made it work more or less for Server Reflexive candidates - check here https://github.com/wiretrustee/ice/tree/singleport
There are a few issues tho when running on many peers simultaneously, plus no tests really. Still WIP.

stv0g:
@braginini Having dedicated muxes per candidate type would force us to use the candidate type for all peers.

As using mixed candidate types would leave us with multiple muxes, each with their own local address.
This is problematic as our Wireguard device can only listen on a single port.

(This could maybe be worked around with some NFtables redirects?)

gather_test.go
stv0g: @mlsmaycon Thanks :)

mlsmaycon:
Closing in favor of #419

PR #419, opened 2022-02-09
Use UDPMux for Server reflexive candidates

braginini:
#### Description
See reference issue for details https://github.com/pion/ice/issues/400

**Core changes:**
* Modified [UDPMux](https://github.com/pion/ice/blob/master/udp_mux.go) to additionally support Server Reflexive candidates (besides Host candidates).
* Added UDPMuxSrflx to Agent
* If UDPMuxSrflx is not provided, the original logic for server reflexive candidates is applied.

The reason why we went with 2 separate UDPMux'ex for Host and Srflx candidates is the conflicts between different candidate types on candidates gathering. Original UDPMux uses uFrag to mux connections which is fine for the case when there is a single candidate type. When there are more different candidate types the packets are getting mixed up. 
There is a possible solution to that which is using the STUN transaction ID map. The logic will be more complicated. 

**Next Steps:**
We will introduce a **UDPMuxRelay** alongside **UDPMux** (for host candidates) and **UDPMuxSrflx** (for server reflexive candidates). At this point, there will be 1 UDP port for each candidate type (host, srflx, relay).

We will then consolidate these 3 UDPMux'es into a single one. I believe that once all the candidate type UDPMux'es are ready we'd be able to come up with a better refactor.

#### Reference issue
https://github.com/pion/ice/issues/400

**EDIT:**

This change breaks the UDPMux interface by adding the GetXORMappedAddr function.
See https://github.com/pion/ice/pull/419#discussion_r802930982

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -122,9 +122,10 @@ type Agent struct {
 	loggerFactory logging.LoggerFactory
 	log           logging.LeveledLogger
 
-	net    *vnet.Net
-	tcpMux TCPMux
-	udpMux UDPMux
+	net         *vnet.Net
+	tcpMux      TCPMux
+	udpMux      UDPMux
+	udpMuxSrflx UniversalUDPMux
 
 	interfaceFilter func(string) bool
 
@@ -319,6 +320,7 @@ func NewAgent(config *AgentConfig) (*Agent, error) { //nolint:gocognit
 		a.tcpMux = newInvalidTCPMux()
 	}
 	a.udpMux = config.UDPMux
+	a.udpMuxSrflx = config.UDPMuxSrflx
 
 	if a.net == nil {
 		a.net = vnet.NewNet(nil)
@@ -891,6 +893,9 @@ func (a *Agent) removeUfragFromMux() {
 	a.tcpMux.RemoveConnByUfrag(a.localUfrag)
 	if a.udpMux != nil {
 		a.udpMux.RemoveConnByUfrag(a.localUfrag)
+	}
+	if a.udpMuxSrflx != nil {
+		a.udpMuxSrflx.RemoveConnByUfrag(a.localUfrag)
 	}
 }
 
--- a/gather.go
+++ b/gather.go
@@ -97,7 +97,11 @@ func (a *Agent) gatherCandidates(ctx context.Context) {
 		case CandidateTypeServerReflexive:
 			wg.Add(1)
 			go func() {
-				a.gatherCandidatesSrflx(ctx, a.urls, a.networkTypes)
+				if a.udpMuxSrflx != nil {
+					a.gatherCandidatesSrflxUDPMux(ctx, a.urls, a.networkTypes)
+				} else {
+					a.gatherCandidatesSrflx(ctx, a.urls, a.networkTypes)
+				}
 				wg.Done()
 			}()
 			if a.extIPMapper != nil && a.extIPMapper.candidateType == CandidateTypeServerReflexive {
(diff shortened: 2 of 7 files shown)


agent.go
```diff
@@ -905,7 +910,13 @@ func (a *Agent) Close() error {
 	})
 	a.err.Store(ErrClosed)
 
-	a.removeUfragFromMux()
+	a.tcpMux.RemoveConnByUfrag(a.localUfrag)
```
Sean-Der, line 913: Why not use `removeUfragFromMux` ?

udp_mux.go
```diff
@@ -35,14 +39,20 @@ type UDPMuxDefault struct {
 	pool *sync.Pool
 
 	mu sync.Mutex
+
+	// since we have a shared socket, for srflx candidates it makes sense to have a shared mapped address across all the agents
+	// todo update it periodically
```
Sean-Der, line 44: Can you explain `todo update it periodically` more?

udp_mux.go
```diff
@@ -204,6 +219,40 @@ func (m *UDPMuxDefault) createMuxedConn(key string) *udpMuxedConn {
 	return c
 }
 
+func (m *UDPMuxDefault) handleRemotePing(msg *stun.Message) (*udpMuxedConn, error) {
+	attr, err := msg.Get(stun.AttrUsername)
+	if err != nil {
+		return nil, err
+	}
+
+	ufrag := strings.Split(string(attr), ":")[0]
```
Sean-Der, line 228: Will this segfault if an attacker sends a message without `:` ?

udp_mux.go
```diff
@@ -204,6 +219,40 @@ func (m *UDPMuxDefault) createMuxedConn(key string) *udpMuxedConn {
...
+	return destinationConn, nil
+}
+
+func (m *UDPMuxDefault) handleXORMappedResponse(stunAddr *net.UDPAddr, msg *stun.Message) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	mappedAddr, ok := m.xorMappedAddr[stunAddr.String()]
+	if !ok {
+		// todo: investigate why remote peers send XORMappedAddr
```
Sean-Der, line 241: Can we remove TODO?

udp_mux.go
```diff
@@ -16,6 +19,7 @@ type UDPMux interface {
 	io.Closer
 	GetConn(ufrag string) (net.PacketConn, error)
 	RemoveConnByUfrag(ufrag string)
+	GetXORMappedAddr(serverAddr net.Addr, deadline time.Duration) (*stun.XORMappedAddress, error)
```
Sean-Der, line 22: Adding `GetXORMappedAddr` breaks the public API.

Does this matter, does anyone implement? If everyone uses our `Default` I think it is ok.

boks1971 (approved):
lgtm! This is very nice!

@davidzhao would be the best person to catch any gotchas here. I am not fully familiar with this module.

udp_mux.go
braginini: Good point and a good question :)  
The thing is that if we want to be universal and support all the types of candidates and not only host candidates then this interface should change.
Also, we will be adding **GetRelayedAddr** function in the next steps to support relay as well. 

One option we could choose not to break changes is to introduce a new interface **UniversalUDPMux** which will eventually replace the original **UDPMux** (which can be deprecated at this point).

I'm hoping that everyone uses the default UDPMux :) 

What are your thoughts on that @Sean-Der @davidzhao @boks1971 ?

braginini: At least I see that it won't break things in pion/webrtc cuz the default one is used
https://github.com/pion/webrtc/blob/master/icemux.go#L22

udp_mux.go
gigovich: No, because `strings.Split` always returns a slice with 1 element, even for an empty string.

Sean-Der: :+1:

gather.go
```diff
@@ -333,6 +337,68 @@ func (a *Agent) gatherCandidatesSrflxMapped(ctx context.Context, networkTypes []
...
+					return
+				}
+
+				xoraddr, err := a.udpMuxSrflx.GetXORMappedAddr(serverAddr, stunGatherTimeout)
+				if err != nil {
+					a.log.Warnf("could not get server reflexive address %s %s: %v\n", network, url, err)
+					return
+				}
+
+				conn, err := a.udpMuxSrflx.GetConn(a.localUfrag)
```
davidzhao, line 367: wouldn't this mean all of the srflx candidates would be using the same conn? If so when any one of the STUN servers fails, it'll end up closing the shared connection in `closeConnAndLog`

this might require a different key that's specific to each ufrag + url.

davidzhao: I remember if a candidate is ultimately not chosen, it'll also get closed by the ICE agent. In that case even the chosen srflx candidate will be closed too.

udp_mux.go
davidzhao: this probably should use a different interface. it doesn't feel that the simple UDPMux should be concerned with STUN or XORMappedAddr.

davidzhao (commented):
I feel a bit unease with `GetXORMappedAddr` in the UDPMux interface. It feels very specific to Srflx and it would complicate UDPMux for those that don't require this functionality. (Since most are running Pion in the server environment, they don't have to worry about srflx).

udp_mux.go
braginini: Got you,

To keep muxing logic separate I think we could then add a layer (struct) that reads from the UDPConn and handles STUN packets before the muxer. But then, this struct has to always wrap the muxer and should be exposed to the agent user. 
I'm happy to jump in a call and discuss this further. We are very much looking forward to improving this part further and making pion/ice more efficient.

braginini: Agreed with @davidzhao that we'll create a separate SRFLXUDPMux

gather.go
braginini: that's right, good point.

braginini:
@davidzhao @Sean-Der 
Updated the PR as per suggestions of @davidzhao 

I kept **UDPMux** and **UDPMuxDefault** as is and added **UniversalUDPMux** that embeds UDPMuxDefault.
UniversalUDPMux is supposed to work with any type of candidate but so far it works only with Server Reflexive. 
I will be adding relay and host candidates support in the following PR.

The core idea of UniversalUDPMuxDefault is to just handle packets coming from the STUN server discovering a [XorMappedAddress](https://github.com/wiretrustee/ice/blob/138db20d36ad19d6561cb6f4739ca4946fea51f6/udp_mux_universal.go#L112). 
All other packets are just forwarded to the UDPMuxDefault that does the actual muxing. 
(UDPMuxDefault is embedded in the [UniversalUDPMuxDefault](https://github.com/wiretrustee/ice/blob/138db20d36ad19d6561cb6f4739ca4946fea51f6/udp_mux_universal.go#L61)).
All this is done by overriding [ReadFrom](https://github.com/wiretrustee/ice/blob/138db20d36ad19d6561cb6f4739ca4946fea51f6/udp_mux_universal.go#L90) of the original UDPConn

I used [STUN URL](https://github.com/wiretrustee/ice/blob/138db20d36ad19d6561cb6f4739ca4946fea51f6/gather.go#L370) to add some uniqueness to the muxedconn and to support multiple STUN servers. 

**Important** I had to modify the original UDPMuxDefault.RemoveConnByUfrag to avoid code duplication.
The change is pretty simple and shouldn't affect anything. Instead of removing conn by exact ufrag match I remove it by [prefix](https://github.com/wiretrustee/ice/blob/138db20d36ad19d6561cb6f4739ca4946fea51f6/udp_mux.go#L109).
This allows me to remove all the connections created by UniversalUDPMuxDefault (with the STUN URL suffix).

I tried to put as many comments as possible but let me know if you need further explanations.
What do you think guys?

braginini:
Hey @Sean-Der @davidzhao 

I reset our fork branch to a clean state and commit the whole work once again.
The reason being Github actions that I completely missed - there were many failing (e.g. lint).
I enabled actions in our fork, fixed all the issues and added more tests. All is green now :)

I'm very sorry for not doing it from the very beginning!
Looking forward for you to checking my work.

Cheers,
Misha

davidzhao (approved):
This is really great work @braginini! It's awesome to have your contributions here.

PR #420, opened 2022-02-11
Implement RTT calculation for candidate pair stats

kcaffrey:
Implement CurrentRoundTripTime, TotalRoundTripTime, RequestsSent, and
ResponsesReceived for candidate pair statistics.

This change on its own is not quite enough to get stats working end to
end. Notably, `(*Agent).GetSelectedCandidatePair()` copies both
candidates, which due to internally marshaling and unmarshaling the
candidate to implement the copy, ends up generating new candidate ids.

Because pion/webrtc stores stats in a map based on candidate ids, this
means that it is currently impractical to actually access these stats.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,6 @@ require (
 	github.com/pion/transport v0.13.1
 	github.com/pion/turn/v2 v2.0.8
 	github.com/stretchr/testify v1.7.1
-	golang.org/x/net v0.0.0-20220728211354-c7608f3a8462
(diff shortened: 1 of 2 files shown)


jech:
Hmm... you're using just the last sample, which might yield an extremely noisy signal.  Shouldn't we be computing an exponential average instead?  (Please see Figure 6 of https://arxiv.org/pdf/1403.3488.pdf for an illustration of what I mean.)

PR #424, opened 2022-02-20
Make TCPMux IPv4/IPv6 aware

Sean-Der:
TCPMux before would create one internal connection per ufrag. This could
cause remote IPv6 traffic to be dispatched to a local IPv4 handler (or
the inverse). The ice.Agent would then discard the traffic since a
candidate pair must be the same IP version.

This commit now creates two connections per ufrag. When requesting a
connection for a ufrag the user must specify if they want IPv4 or IPv6.

Relates to pion/webrtc#2125
Relates to pion/webrtc#1356

### DIFF
--- a/tcp_mux.go
+++ b/tcp_mux.go
@@ -19,7 +19,7 @@ import (
 // 2. allow mocking in tests.
 type TCPMux interface {
 	io.Closer
-	GetConnByUfrag(ufrag string) (net.PacketConn, error)
+	GetConnByUfrag(ufrag string, isIPv6 bool) (net.PacketConn, error)
 	RemoveConnByUfrag(ufrag string)
 }
 
@@ -36,7 +36,7 @@ func (m *invalidTCPMux) Close() error {
 }
 
 // GetConnByUfrag implements TCPMux interface.
-func (m *invalidTCPMux) GetConnByUfrag(ufrag string) (net.PacketConn, error) {
+func (m *invalidTCPMux) GetConnByUfrag(ufrag string, isIPv6 bool) (net.PacketConn, error) {
 	return nil, ErrTCPMuxNotInitialized
 }
 
@@ -49,8 +49,8 @@ type TCPMuxDefault struct {
 	params *TCPMuxParams
 	closed bool
 
-	// conns is a map of all tcpPacketConns indexed by ufrag
-	conns map[string]*tcpPacketConn
+	// connsIPv4 and connsIPv6 are maps of all tcpPacketConns indexed by ufrag
+	connsIPv4, connsIPv6 map[string]*tcpPacketConn
 
 	mu sync.Mutex
 	wg sync.WaitGroup
@@ -72,7 +72,8 @@ func NewTCPMuxDefault(params TCPMuxParams) *TCPMuxDefault {
 	m := &TCPMuxDefault{
 		params: &params,
 
-		conns: map[string]*tcpPacketConn{},
+		connsIPv4: map[string]*tcpPacketConn{},
+		connsIPv6: map[string]*tcpPacketConn{},
 	}
 
 	m.wg.Add(1)
@@ -109,7 +110,7 @@ func (m *TCPMuxDefault) LocalAddr() net.Addr {
 }
 
 // GetConnByUfrag retrieves an existing or creates a new net.PacketConn.
-func (m *TCPMuxDefault) GetConnByUfrag(ufrag string) (net.PacketConn, error) {
+func (m *TCPMuxDefault) GetConnByUfrag(ufrag string, isIPv6 bool) (net.PacketConn, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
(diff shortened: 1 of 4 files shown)


Sean-Der:
@jech Hopefully this fixes the issues you are seeing! Sorry it took so long to understand/reproduce the issue.

tcp_mux.go
```diff
@@ -109,33 +110,34 @@ func (m *TCPMuxDefault) LocalAddr() net.Addr {
 }
 
 // GetConnByUfrag retrieves an existing or creates a new net.PacketConn.
-func (m *TCPMuxDefault) GetConnByUfrag(ufrag string) (net.PacketConn, error) {
+func (m *TCPMuxDefault) GetConnByUfrag(ufrag string, forAddress net.IP) (net.PacketConn, error) {
```
jeremija, line 113: It seems the only thing we need to know here is whether the IP is v4 or v6. Did you have a special reason for passing the whole `net.IP` rather than just the `isIPv6 boolean` value here?

jech:
Thanks a lot, Sean, I'll test it during the night (when nobody is using the sample server).

FWIW, I've seen similar symptoms with UDPMux, see https://github.com/pion/webrtc/issues/1915.  I'm wondering if it's the same issue?

jech:
Perhaps I'm doing something wrong, but it still doesn't work for me.  I've put a TCP listener on port 8444, then did the following on the client to block all UDP and TCP TURN:
```
iptables -I INPUT -s 51.210.14.2 -p udp -j DROP
ip6tables -I INPUT -p udp -j DROP
iptables -I INPUT -p tcp --sport 1194 -j DROP
ip6tables -I INPUT -p tcp --sport 1194 -j DROP
```
The client is the offerer and the media sender.  It connects to the TCP listener, and sends some data:
```
16:17:43.149235 IP 192.168.0.27.43936 > 51.210.14.2.8444: Flags [SEW], seq 1304065160, win 64240, options [mss 1460,sackOK,TS val 3863861007 ecr 0,nop,wscale 7], length 0
16:17:43.156764 IP 51.210.14.2.8444 > 192.168.0.27.43936: Flags [S.E], seq 24426563, ack 1304065161, win 65160, options [mss 1460,sackOK,TS val 4145864461 ecr 3863861007,nop,wscale 7], length 0
16:17:43.156800 IP 192.168.0.27.43936 > 51.210.14.2.8444: Flags [.], ack 1, win 502, options [nop,nop,TS val 3863861015 ecr 4145864461], length 0
```
It's sending some data:
```
\16:17:44.168249 IP 192.168.0.27.43936 > 51.210.14.2.8444: Flags [P.], seq 115:229, ack 1, win 502, options [nop,nop,TS val 3863862026 ecr 4145864799], length 114
16:17:44.175367 IP 51.210.14.2.8444 > 192.168.0.27.43936: Flags [.], ack 229, win 509, options [nop,nop,TS val 4145865479 ecr 3863862026], length 0
```
However, I see no traffic in the opposite direction.  After a while, the connection drops:
```
16:18:28.705050 IP 192.168.0.27.43938 > 51.210.14.2.8444: Flags [P.], seq 2509:2623, ack 1, win 502, options [nop,nop,TS val 3863906563 ecr 4145909334], length 114
16:18:28.712831 IP 51.210.14.2.8444 > 192.168.0.27.43938: Flags [.], ack 2623, win 509, options [nop,nop,TS val 4145910016 ecr 3863906563], length 0
16:18:28.753946 IP 192.168.0.27.43938 > 51.210.14.2.8444: Flags [F.], seq 2623, ack 1, win 502, options [nop,nop,TS val 3863906612 ecr 4145910016], length 0
```
At that point, ICE gets into failed state, which triggers an ICE restart, and the cycle starts again.

tcp_mux.go
Sean-Der: @jeremija fixed!

jeremija: Thanks!

boks1971 (approved):
lgtm!

jech:
Interestingly, I can only reproduce the issue if both TCP/IPv4 and TCP/IPv6 are unblocked.  If I aablesdditionally block TCP/IPv6 (`ip6tables -I INPUT -p tcp -j DROP`), then everything works fine.

davidzhao:
> FWIW, I've seen similar symptoms with UDPMux, see [pion/webrtc#1915](https://github.com/pion/webrtc/issues/1915). I'm wondering if it's the same issue?

Fairly certain UDPMux is also not IPv6 aware. If this works, we'll port the same fix to UDPMux

davidzhao (approved):
lg! I don't have a repro to validate this, and will defer to @jech to confirm that it resolves the issue.

jech:
The behaviour is definitely better than it was: I now get an ICE failure consistently across browsers (tested with Brave and Firefox), which is something I already know how to handle gracefully, while previously I was getting erratic behaviour that I didn't know how to detect.

Hence, unless Sean has any further ideas, I recommend merging this patch while keeping the bug report open.

maeln:
Hi!

Because our use-case rely on two PR that are not merge yet ( https://github.com/pion/webrtc/pull/1812 & https://github.com/pion/mediadevices/pull/338 ), it is a bit difficult to test this PR against our own issue in a timely manner.

So don't wait for my feedback. I will see with my colleague to try to organize our own fork/pr to test this ASAP.

Thanks !

Sean-Der:
@jech I still have ice-tcp issues as well! With FireFox I am not able to connect when Pion is the controlled agent.

* FireFox sends a binding request with UseCandidate. Pion creates a peer reflexive candidate + candidate pair
* Pion sends a binding request (to validate the candidate)
* FireFox sends another binding request with UseCandiate. Pion hasn't gotten a response to it's binding request yet and ignores (candidate pair isn't valid yet)
* FireFox responds with Binding Success. Pion marks candidate pair as valid.

If FireFox sent another Binding Request everything would go to connected, not sure why it doesn't. I have attached the pcap here.  You can reproduce at http://18.216.219.191:8080/ it is just running https://github.com/pion/webrtc/tree/master/examples/ice-tcp

[firefox-tcp-client.pcap.zip](https://github.com/pion/ice/files/8109577/firefox-tcp-client.pcap.zip)

Sean-Der:
I am going to merge this, but just makes this a `Relates to` on the two tickets. It fixes at least one problem

PR #426, opened 2022-02-21
Don't drop Binding Requests in Controlled Agent

Sean-Der:
A controlled Agent would discard incoming Binding Requests if it didn't
cause the pair to be selected. For UDP Candidate this would be
interpreted as packet loss. For TCP Candidates not responding with a
Binding Success could be interpreted as a failure.

Firefox's ICE Agent would disconnect TCP Candidates because of this
behavior.

Resolves to pion/webrtc#2125
Resolves to pion/webrtc#1356
See https://bugzilla.mozilla.org/show_bug.cgi?id=1756460

### DIFF
--- a/candidatepair.go
+++ b/candidatepair.go
@@ -18,12 +18,13 @@ func newCandidatePair(local, remote Candidate, controlling bool) *CandidatePair 
 // CandidatePair is a combination of a
 // local and remote candidate
 type CandidatePair struct {
-	iceRoleControlling  bool
-	Remote              Candidate
-	Local               Candidate
-	bindingRequestCount uint16
-	state               CandidatePairState
-	nominated           bool
+	iceRoleControlling       bool
+	Remote                   Candidate
+	Local                    Candidate
+	bindingRequestCount      uint16
+	state                    CandidatePairState
+	nominated                bool
+	nominateOnBindingSuccess bool
 }
 
 func (p *CandidatePair) String() string {
--- a/selection.go
+++ b/selection.go
@@ -229,13 +229,17 @@ func (s *controlledSelector) HandleSuccessResponse(m *stun.Message, local, remot
 
 	p.state = CandidatePairStateSucceeded
 	s.log.Tracef("Found valid candidate pair: %s", p)
+	if p.nominateOnBindingSuccess {
+		if selectedPair := s.agent.getSelectedPair(); selectedPair == nil {
+			s.agent.setSelectedPair(p)
+		}
+	}
 }
 
 func (s *controlledSelector) HandleBindingRequest(m *stun.Message, local, remote Candidate) {
 	useCandidate := m.Contains(stun.AttrUseCandidate)
 
 	p := s.agent.findPair(local, remote)
-
 	if p == nil {
 		p = s.agent.addPair(local, remote)
 	}
@@ -251,7 +255,6 @@ func (s *controlledSelector) HandleBindingRequest(m *stun.Message, local, remote
 			if selectedPair := s.agent.getSelectedPair(); selectedPair == nil {
 				s.agent.setSelectedPair(p)
 			}
-			s.agent.sendBindingSuccess(m, local, remote)
 		} else {
 			// If the received Binding request triggered a new check to be
 			// enqueued in the triggered-check queue (Section 7.3.1.4), once the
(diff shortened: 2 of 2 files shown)


Sean-Der:
@jech I have it now! Mind trying again? 

The behavior would exhibit itself if FireFox made another request before it responded to Pion. I was able to reproduce by adding some arbitrary delays, then debugged on the FireFox side.

jech:
Very slight improvement:
  - works when Brave (Chromium) is the offerer;
  - fails when Brave is the answerer;
  - fails with Firefox, both as offerer and as answerer.

Sean-Der:
@jech does http://18.216.219.191:8080/ work for you on FireFox (Pion is answerer)

Will try the other configurations now!

jech:
Yes, it works fine.

Sean-Der:
Oh yes that makes sense! Let me try that now also.  In my example I am only allowing TCP. 

* Do you trickle candidates?
* Is the browser sharing candidates, or just the server?

jech:
> Do you trickle candidates?

Yes.

> Is the browser sharing candidates, or just the server?

Both.

Sean-Der:
@jech To repro easier I started https://github.com/Sean-Der/ice-tcp-test it allows the browser to be Offer or Answer. Right now it only does non-Trickle. 

* I blocked all UDP traffic on my server `sudo iptables -I INPUT -p udp -j DROP && sudo ip6tables -I INPUT -p udp -j DROP`
* I run and set my public IP `go run main.go 18.216.219.191`

This example works for me in FireFox/Brave/Chrome. 

I will add Trickle tomorrow and hopefully I am able to reproduce!

boks1971 (approved):
lgtm!

Sean-Der:
@jech oh good point! I moved it to http://143.198.72.93:8080/ all tests done on FireFox+Chrome on MacOS. I have IPv4 and IPv6 connectivity working.

* Tested with IPv4/IPv6 unblocked
* Tested with IPv4 blocked
* Tested with IPv6 blocked

Everything seems to be working with me. Mind testing again?

jech:
That's bizarre... I've successfully blocked UDP, and everything works as expected (host-reflexive TCP candidate).  Let me compare with Galene.

Sean-Der:
Glad to hear it works! Thank you so much for all your patience/debugging help on this.

High point of my week to have these longstanding issues finally figured out. I have been a bit hesitant to take these on....

jech:
Bizarrer and bizarrer.  Your code works fine even when run on my server.  I've tried copying all the details from your code into Galene, I've tried disabling trickle ICE in Galene, I've tried disabling ICE restarts in Galene, and it still fails.

I'm at a loss.

Sean-Der:
@jech I am going to merge + pin, and then will deploy Galene to my server!

I don't have any good debugging ideas. I spent a lot of time printf debugging FireFox, no good single source for info with this stuff :/

jech:
It appears like it's not a matter of who is the offerer: the first few TCP candidates work, then it starts failing.  Only way to fix the issue is to restart the server.

For now, you may test at <https://galene.org:8444/group/public/> (note the non-standard port, port 8443 is running mainline Galene).

To test on your own server, do
```
git clone -b ice-tcp https://github.com/jech/galene/
cd galene
CGO_ENABLED=0 go build
mkdir groups
echo '{"presenter":[{}]}' > groups/public.json
rsync -r galene public groups server.example.com:
ssh server.example.com ./galene -tcp-port 8888
```
Then visit https://server.example.com

Sean-Der:
@jech Does adding a replace statement make a difference for you?

```
diff --git a/go.mod b/go.mod
index 1f4692d..1a5931e 100644
--- a/go.mod
+++ b/go.mod
@@ -17,3 +17,5 @@ require (
        golang.org/x/crypto v0.0.0-20220214200702-86341886e292
        golang.org/x/sys v0.0.0-20220209214540-3681064d5158
 )
+
+replace github.com/pion/ice/v2 => github.com/pion/ice/v2 v2.2.1-0.20220221210118-fca128a058ec
```

Sean-Der:
I tried locally and got the same thing, but I believe the replace statement fixed it! I don't know Go modules well enough to dump what exact versions are being used/dependency tree.

jech:
Not for me, unfortunately.  I'm seeing the same results as above: good behaviour for the first few connections after server restart, then TCP candidates are broken until I restart the server.

Very weird.  Perhaps a data structure somewhere that accumultates obsolete data?

Sean-Der:
@jech I just tested 100 PeerConnections with `ice-tcp-test` and everything worked!

Will test with Galene now

PR #438, opened 2022-04-05
Reduce critical section when writing with TCPMux.

davidzhao:
To avoid one bad connection causing locked processes/cascading failures.

When the connection is stuck, writes will not timeout and will end up holding the lock forever. Attempts to close the connections will be stuck waiting for the mutex.


boks1971 (approved):

jeremija (approved):
LGTM, thanks!

Issue #305, opened 2020-11-28
Pion chooses wrong remote candidate


jech:
Unnamed SFU using Pion running on remote host.  Client is Chromium on Linux and is multihomed: has both an Ethernet and a Wifi connection.  The client offers four host candidates, IPv4+IPv6 for each of the two interfaces.

When Chomium is the controlling agent, then the Ethernet interface is chosen.  When Pion is the controlling agent, on the other hand, the WiFI interface is chosen deterministically.

A few other elements:

  - Chromium sets the candidate priorities correctly (Ethernet higher than WiFi);
  - Chromium sets the `network-cost` to 10 for the WiFi candidates; I believe this is described in https://tools.ietf.org/html/draft-thatcher-ice-network-cost (expired).

It is my understanding that Pion does not use the network cost, but that the candidate priority should be enough to do the right thing.  This may be related to https://github.com/pion/ice/issues/82.

jech:
I can confirm that the priorities of Chromium's host candidates are reasonable:
  * 2122262783 for the IPv6 address of the Ethernet interface;
  * 2122194687 for the IPv4 address of the Ethernet interface;
  * 2122131711 for the IPv6 address of the WiFi interface; and
  * 2122063615 for the IPv4 address of the WiFi interface.

So it definitely looks like there's something wrong with Pion's candidate selection.

jech:
I am no longer able to reproduce this — it looks like Pion is choosing the candidate with the highest priority, as it should.  Closing for now.

jech:
The issue cannot be reproduced locally, but it is definitely there when the server is remote.

Interestingly enough, if I trigger an ICE restart after the connection is established, then Pion switches to the right candidate.  So this might very well be a timing issue: the WiFi candidate is chosen before the Ethernet candidate is received.  Perhaps delaying ICE by a fraction of a second would help?

jech:
@Sean-Der has suggested that this might be due to `defaultHostAcceptanceMinWait` in `agent_config.go` being set to 0.  More experimentation is needed.

jech:
Adding
```
s.SetHostAcceptanceMinWait(200 * time.Millisecond)
```
at the right place does not solve the issue — the wrong candidate is still chosen.

aziztitu:
@jech Were you able to find a solution for this? I'm facing a similar problem at the moment. Pion seems to choose a Relay candidate with higher priority while a srflx candidate with lower priority is available

jech:
[...]

PR #446, opened 2022-04-22
Enable acceptAggressiveNomination by default

Sean-Der:
Always accept Aggressive Nominations by remote agent

Relates to 18f7a214aae82a41a6d91a90a7652cf66e3e90e2

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -133,8 +133,6 @@ type Agent struct {
 	insecureSkipVerify bool
 
 	proxyDialer proxy.Dialer
-
-	acceptAggressiveNomination bool
 }
 
 type task struct {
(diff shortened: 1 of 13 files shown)


Sean-Der:
@jech Anyone have concerns/doubts about this?

jech:
I haven't tested yet, but it looks correct to me.  I see nothing in RFC 5245 that makes aggressive nomination optional, so my understanding is that we must expect the peer to do aggressive nomination if it so desires.  If my understanding is correct, we must accept aggressive nomination, no ifs, no buts.

jech:
I've done some testing with a multihomed client (Ethernet+WiFi), and it doesn't seem to break anything (tested on Brave and Firefox).  OTOH, it doesn't seem to fix #305.

PR #450, opened 2022-05-04
No log newlines

stv0g:
This PR harmonizes the use of newlines in log statements.

My custom logger was a little confused by a mix of logging statements with and without newlines.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -610,7 +610,7 @@ func (a *Agent) pingAllCandidates() {
 		}
 
 		if p.bindingRequestCount > a.maxBindingRequests {
-			a.log.Tracef("max requests reached for pair %s, marking it as failed\n", p)
+			a.log.Tracef("max requests reached for pair %s, marking it as failed", p)
 			p.state = CandidatePairStateFailed
 		} else {
 			a.selector.PingCandidate(p.Local, p.Remote)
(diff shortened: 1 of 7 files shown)


Sean-Der:
Yea that's no good!

Does this change the output that we have today? If you do `export PION_LOG_TRACE=ice` does it look different?

Would it be possible to write a linter/check so this doesn't break again in the future.

stv0g:
> Does this change the output that we have today? If you do export PION_LOG_TRACE=ice does it look different?

No it does not. Mainly because `log.Logger.Output()` will check if the log message ends with a newline or not.
If not, it will add one automatically:

https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/log/log.go;l=192

So the change does not change the output for the default logger which uses the stdlib log.Logger

Sean-Der:
@stv0g That's great news! I am all for merging this. 

If you have the bandwidth would you mind writing a `sh` script that uses `grep` to catch this? If you can just copy `lint-disallowed-functions-in-library.sh`

If you don't the bandwidth totally understand. I would just hate to see your work regressed by someone in the future :(

stv0g:
@Sean-Der Sure :) Here it is https://github.com/pion/.goassets/pull/88

I will rebase this PR and enable the linter once the goasset PR has been merged into pion/ice.

Sean-Der (approved):

PR #451, opened 2022-05-08
Minor tweaks to improve log traces and error output

stv0g:
Really nothing too special.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -568,16 +568,16 @@ func (a *Agent) updateConnectionState(newState ConnectionState) {
 }
 
 func (a *Agent) setSelectedPair(p *CandidatePair) {
-	a.log.Tracef("Set selected candidate pair: %s", p)
-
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -568,12 +568,14 @@ func (a *Agent) updateConnectionState(newState ConnectionState) {
 }
 
 func (a *Agent) setSelectedPair(p *CandidatePair) {
-	a.log.Tracef("Set selected candidate pair: %s", p)
-
 	if p == nil {
+		a.log.Tracef("Reset selected candidate pair")
```
Sean-Der, line 572: `Unset selected candidate pair` or `Set candidate pair to nil` maybe? I would assume that reset was performed upon the candidate pair. Like reseting their state/clearing the count of packets set etc...

Sean-Der (approved):

PR #466, opened 2022-07-20
add nonblock buffer option to TCPMux

cnderrauber:
When sfu to broadcast track to multiple subscriber, a common case is broadcast rtp packets in a loop, it is ok if all subscriber use Udp candidate since it will not block. Like this:

```
pkt := upstreamTrack.ReadRTP()
for _, dt := range downstreamTracks {
    dt.Write(pkt)  // the write will go to net.Conn Write()
}
```

But if a user use tcp candidate, the tcp write would block and make subsequent write op to wait, that means the slow tcp connection cause all subscribers slow. This PR add a nonblock write option to TcpMux for the issue


boks1971 (approved):
lgtm!

PR #469, opened 2022-08-25
fix: UDPMux memory leaks

Antonito:
This PR fixes memory leaks I noticed with my SFU deployment.

- Clients performing an early termination/cancelation would cause `udpMuxedConn` to
leak addresses into `UDPMuxDefault`.
  - A call to `(udpMuxedConn).Close()` would set `(udpMuxedConn).addresses` to `nil`, which prevented their deletion from `(UDPMuxDefault).addressMap`
- Small refacto of `UDPMuxDefault` that replaces `removeConn` by `RemoveConnByUfrag`
- Before that the gatherer would ignore some host candidates
  - I can submit the `gather.go` updates in a separate PR if it helps land this one faster


davidzhao (approved):
great work! thanks for fixing this.

PR #471, opened 2022-09-01
Use new pion/transport Net interface

stv0g:
See https://github.com/pion/transport/pull/204 for a detailed description

Also closes #509


udp_mux.go
```diff
@@ -49,6 +50,7 @@ const maxAddrSize = 512
 type UDPMuxParams struct {
 	Logger  logging.LeveledLogger
 	UDPConn net.PacketConn
+	Net     transport.Net
```
at-wat, line 53: https://github.com/pion/ice/pull/471/files?diff=unified&w=1#diff-a8e331e8f7ba49ca6692af8fc922967307339ad6fe58eb62549602b8988d9bb3L64-L66
> 		// For unspecified addresses, the correct behavior is to return errListenUnspecified, but
> 		// it will break the applications that are already using unspecified UDP connection
> 		// with UDPMuxDefault, so print a warn log and create a local address list for mux.

Seems be used only for the fallback to the wrong usage of `NewUDPMuxDefault`.
If user want to specify `transport.New`, they should use `NewMultiUDPMuxFromPort` instead.

It would be better to add some comment to this field.

enobufs: @stv0g 
This is an API change beyond the what this PR is aiming for. I think the reason why uses vnet (or transport.Net) is just to reuse localInterfaces() that requires vnet/transport.Net. In fact, it only requires net.Interfaces, and
```go
vnet.NewNet(nil) // <-- *NetConfig being nil
````
would return what you can get from the original net.Interfaces, so the code we have here intended to use "virtual interface", and using vnet just for convenience. Therefore, I think we don't need to add transport.Net field to this public struct. (Let's focus on introducing transport.Net for now with minimal impact to users)

**Review by enobufs — COMMENTED** [review]
I've left one comment. (I will review again when it comes out of the draft state.)

File: udp_mux.go (hunk 1 above)
Comment (stv0g) [reply udp_mux.go:53]: I actually tried it without adding the field. However that breaks the tests as the integration will create a `UDPMux` which attempts to listen on addresses of the real host rather than the virtual interfaces of the `vnet.VNet`.

This will cause `vnet/Net.ListenUDP` to fail as it can not bind to a non-existing local address.

Comment (stv0g) [reply udp_mux.go:53]: For normal (non-test usage) the field will be initialized with `nil` and hence use the `stdnet.Net` implementation

Comment (stv0g) [reply udp_mux.go]: @at-wat I added a comment to the new field.

File: candidate_server_reflexive.go [hunk 2]
```diff
@@ -1,4 +1,5 @@
 // Package ice ...
+//
 //nolint:dupl
```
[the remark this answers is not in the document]
stv0g: I removed the duplicate package comment

stv0g:
Thanks @enobufs and @at-wat for your reviews.

> I will review again when it comes out of the draft state

The PR is now ready for merging when you are happy with it :)

gather.go
```diff
@@ -596,18 +596,24 @@ func (a *Agent) gatherCandidatesRelay(ctx context.Context, urls []*URL) { //noli
...
+				udpAddr, connectErr := a.net.ResolveUDPAddr(network, TURNServerAddr)
 				if connectErr != nil {
 					a.log.Warnf("Failed to resolve UDP Addr %s: %v", TURNServerAddr, connectErr)
 					return
 				}
 
-				conn, connectErr := dtls.Dial(network, udpAddr, &dtls.Config{ //nolint:contextcheck
+				udpConn, dialErr := a.net.DialUDP("udp", nil, udpAddr)
+				if dialErr != nil {
+					a.log.Warnf("Failed to dial DTLS Address %s: %v", TURNServerAddr, connectErr)
```
edaniels, line 607:
```suggestion
					a.log.Warnf("Failed to dial DTLS Address %s: %v", TURNServerAddr, dialErr)
```

udp_mux_multi.go
```diff
@@ -203,3 +217,12 @@ func UDPMuxFromPortWithLoopback() UDPMuxFromPortOption {
 		},
 	}
 }
+
+// UDPMuxFromPortWithNet set loopback interface should be included
```
edaniels, line 221:
```suggestion
// UDPMuxFromPortWithNet sets the network transport to use.
```

edaniels (changes requested):
Nice! One actual change to make and one suggested.

stv0g:
Thanks @edaniels,

I committed your proposed changed 👍

edaniels (approved):
LGTM.

PR #472, opened 2022-09-04
Add multi-port wrappers for UDPMux and TCPMux

kcaffrey:
These wrappers allow a caller to provide UDPMux and TCPMux instances
to the ICE agent that represents multiple open ports. This can be
desirable in what would otherwise be single-port deployments, as it
increases the chance that one of the fixed ports will not be blocked
by a user's firewall.


davidzhao (commented):
Great work with this @kcaffrey! It's a neat way of keeping the interface the same while introducing multiple port capabilities.

Curious, what's the practical use for TCP? I think with UDP it was to give it more port variety like Google Meet? Are you looking to send multiple TCP-passive candidates to the peer as well?

kcaffrey:
> Curious, what's the practical use for TCP? I think with UDP it was to give it more port variety like Google Meet? Are you looking to send multiple TCP-passive candidates to the peer as well?

I personally have no use case for TCP. I just figured if I was doing UDP, I should do TCP as well for completeness. However, I'm also fine dropping that part if you prefer.

kcaffrey:
Hm, I could use some help on understanding why the WASM test is failing @davidzhao 

I used [testMuxConnectionPair](https://github.com/pion/ice/pull/472/files#diff-37139adb527a25384695a89da0471660c781583f4b77033eda9a38dc13f3ada7R128) from `udp_mux_test.go` inside `udp_mux_multi_test.go` [here](https://github.com/pion/ice/pull/472/files#diff-c4249a77110d4aa163252bde9f750a4fbda7a1fc2248856e60855f6ced632fbdR80) to avoid code duplication. Should I simply copy that method into `udp_mux_multi_test.go`? Does the WASM test build each test file individually (thus not resolving that reference)?

davidzhao (approved):
lgtm. There's just one failing test on wasm, which I think just needs a exclude flag.

davidzhao:
> I used [testMuxConnectionPair](https://github.com/pion/ice/pull/472/files#diff-37139adb527a25384695a89da0471660c781583f4b77033eda9a38dc13f3ada7R128) from `udp_mux_test.go` inside `udp_mux_multi_test.go` [here](https://github.com/pion/ice/pull/472/files#diff-c4249a77110d4aa163252bde9f750a4fbda7a1fc2248856e60855f6ced632fbdR80) to avoid code duplication. Should I simply copy that method into `udp_mux_multi_test.go`? Does the WASM test build each test file individually (thus not resolving that reference)?

I think it just needs this at the top:

```
//go:build !js
// +build !js
```

kcaffrey:
Thanks for the review @davidzhao !

PR #474, opened 2022-09-17
Attempt to fix deadlock in TCPMux

kcaffrey:
Previously `(*TCPMux).RemoveConnByUfrag` would close the connections while holding its lock, but it is possible for closing to block and/or take a while.

If closing one of the connections blocks (which was occurring rarely due to blocking on a channel send), then the entire `TCPMux` would deadlock and all further operations with the ICE agent would end up getting blocked due to tying up the task loop.

We have not seen any more deadlocks on our servers since applying this patch.


davidzhao (approved):
Great work!

PR #475, opened 2022-09-23
Fix muxed conn shared by candidate

cnderrauber:
When we have multiple host candidates and been mux to a single port, if these candidates share a same conn (either tcp or udp), they might read other's msg
cause ice connection cost long time or failed.

This should also fix #473 

#### Reference issue
Fixes #...

### DIFF
--- a/udp_mux.go
+++ b/udp_mux.go
@@ -15,7 +15,7 @@ import (
 // UDPMux allows multiple connections to go over a single UDP port
 type UDPMux interface {
 	io.Closer
-	GetConn(ufrag string, isIPv6 bool) (net.PacketConn, error)
+	GetConn(ufrag string, isIPv6 bool, local net.IP) (net.PacketConn, error)
 	RemoveConnByUfrag(ufrag string)
 }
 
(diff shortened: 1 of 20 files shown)


udp_mux.go
```diff
@@ -40,10 +45,22 @@ type UDPMuxDefault struct {
 
 const maxAddrSize = 512
 
+type UdpMuxConn interface {
+	net.PacketConn
+
+	// ReadMsgUdp used to get detination address when received a udp packet
```
boks1971, line 51: :nit: destination

boks1971 (commented):
Looks good to me but I am not well versed with the depths of this.

davidzhao:
Just tying to understand the change better. Is the situation related to IPv6? If we are looking up by ufrag, it seems that it shouldn't mix up between agents?

cnderrauber:
> Just tying to understand the change better. Is the situation related to IPv6? If we are looking up by ufrag, it seems that it shouldn't mix up between agents?

No, it happens when sfu have multiple host address can be accessed from client. we are not mix up between agents, but candidates, since candidate of single agents share same ufrag.

davidzhao (approved):
I understand now, thank you for fixing this!

davidzhao:
> No, it happens when sfu have multiple host address can be accessed from client. we are not mix up between agents, but candidates, since candidates of single agent share same ufrag. This issue not caused by IPV6, but easy reproduced when ipv6 enabled.

Another question. If we are using TCP/UDP muxes, why would we add several candidates for different IPs? Since listening is performed by the user, it seems that returning a single muxed candidate per network type would be sufficient.

cnderrauber:
> > No, it happens when sfu have multiple host address can be accessed from client. we are not mix up between agents, but candidates, since candidates of single agent share same ufrag. This issue not caused by IPV6, but easy reproduced when ipv6 enabled.
> 
> Another question. If we are using TCP/UDP muxes, why would we add several candidates for different IPs? Since listening is performed by the user, it seems that returning a single muxed candidate per network type would be sufficient.

we can't return only one candidate since user may listen at all network interfaces (0.0.0.0 and ::).

PR #476, opened 2022-09-26
IpMapper return locIP if corresponding map not set

cnderrauber:
#### Description
If we only set v4 or v6 NAT1to1Mapping, then findExternalIP will failed for the unset ip mapper, the gatherer will complain 1:1 NAT mapping is enabled but no external IP is found, cause that candidate for udp can't be gathered.

#### Reference issue
Fixes #...


davidzhao (approved):
lgtm

PR #478, opened 2022-09-29
Add IPFilter to AgentConfig

cnderrauber:
When machine's network interface have more than one ip address and user don't want expose one of these ips to remote peer, interface filter can't work in this case, so add a ip filter for that.


davidzhao (approved):
lgtm!

PR #479, opened 2022-09-30
Fix udpmux not work on AWS/ECS

cnderrauber:
ECS network might not set destination address of udp packet to local address, so we can't only rely on it to dispatch packet.

related to https://github.com/livekit/livekit/pull/1036 user report

### DIFF
--- a/udp_mux.go
+++ b/udp_mux.go
@@ -365,13 +365,12 @@ func (m *UDPMuxDefault) connWorker() { //nolint:gocognit
 		var destinationConn *udpMuxedConn
 		m.addressMapMu.Lock()
 		if conns, ok := m.addressMap[addr.String()]; ok {
-			if localHost.IsUnspecified() {
+			destinationConn, ok = conns[ipAddr(localHost.String())]
+			if !ok {
 				for _, c := range conns {
 					destinationConn = c
 					break
 				}
-			} else {
-				destinationConn = conns[ipAddr(localHost.String())]
 			}
 		}
 		m.addressMapMu.Unlock()


Sean-Der:
Hey @cnderrauber, would you mind explaining a bit more?

Can we ever depend on the destination address? If anyone can modify it do we have security/DoS concerns?

Sean-Der:
I am also curious/don't understand the use case https://github.com/pion/ice/commit/93980395c8b7f6e4ad5ac6627734cdd03e2c087b

cnderrauber:
> Hey @cnderrauber, would you mind explaining a bit more?
> 
> Can we ever depend on the destination address? If anyone can modify it do we have security/DoS concerns?

I think we don't have those concerns. If it is modified, the router will not deliver the packet to our network interface.

udp_mux.go
```diff
@@ -365,13 +365,12 @@ func (m *UDPMuxDefault) connWorker() { //nolint:gocognit
 		var destinationConn *udpMuxedConn
 		m.addressMapMu.Lock()
 		if conns, ok := m.addressMap[addr.String()]; ok {
-			if localHost.IsUnspecified() {
+			destinationConn, ok = conns[ipAddr(localHost.String())]
```
boks1971, line 368: Does this break Windows build again @cnderrauber ? You had made this change to fix Windows build in https://github.com/pion/ice/commit/12148c5d62514bcfc88b2c47a6a7a56a7774e3a3. Would this be an issue again now?

udp_mux.go
```diff
@@ -365,13 +365,12 @@ func (m *UDPMuxDefault) connWorker() { //nolint:gocognit
...
+			if !ok {
 				for _, c := range conns {
```
boks1971, line 370: Dumb question? When there is no match, does it make sense to all the conns rather than just the first one?

udp_mux.go
cnderrauber: No, windows should work well for this

udp_mux.go
cnderrauber: Dispatch to all the conns will cause repeated data packet received. Actually, I'm thinking of (and should do) re-design of udp mux, this fix PR will have same behavior as the old version(before my fix), both them can work but not ideal.

boks1971: Sounds good. Yeah, getting back to previous behaviour first is good.

boks1971 (approved):

cnderrauber:
I think this pr can be merged first, then draft a new pr to implement the new way for udp mux (gather fetch candidates from mux, and mux enumerate interfaces to listen all address instead of `0.0.0.0`).

Issue #435, opened 2022-03-29
gatherCandidatesLocalUDPMux gathers candidates not on socket


50infivedays:
### Your environment.
  - Version: *ice v2.2.2*
  - Browser: *any version*
![image](https://user-images.githubusercontent.com/2149747/160547959-7cd67bb2-503f-465b-a82f-e85c49f70a47.png)

### What did you expect?
i'm expecting it should only gather candidates on socket

### What happened?
it will gather every ip as candiddate , if you have multi ip on your interface , it will make a wrong candidate

50infivedays:
@Sean-Der is this a bug? i think it should skip the ip when it is not matched with the socket

50infivedays:
> @50infivedays that looks like a bug to me!
> 
> `pion/ice` was written with the assumption that every interface will only have one IP, so I bet we have lots of bugs/behaviors that aren't expected.

will multi-IP on one interface be supported in the future?

stv0g:
I think it is already supported.

There is just another caveat here: pion/ice uses pion/transports `vnet.Nat` to get a list of interfaces and addresses during `ice.NewAgent()` which remains static for the lifetime of the agent.

No changes of interfaces or addresses will be taken into account.

cnderrauber:
Fixed by #483 , gather will collect candidates from Mux's socket now.

PR #483, opened 2022-10-08
Use MultiUDPMux to implement listen any address

### LINKED ISSUE #482: failed to build for freebsd_amd64_v1
```
Error: mod/github.com/pion/ice/v2@v2.2.10/udpmsghelper.go:18:73: undefined: syscall.IPV6_2292PKTINFO
Error: mod/github.com/pion/ice/v2@v2.2.10/udpmsghelper.go:21:71: undefined: syscall.IP_PKTINFO
Error: mod/github.com/pion/ice/v2@v2.2.10/udpmsghelper.go:34:63: undefined: syscall.IP_PKTINFO
Error: mod/github.com/pion/ice/v2@v2.2.10/udpmsghelper.go:35:26: undefined: syscall.Inet4Pktinfo
Error: mod/github.com/pion/ice/v2@v2.2.10/udpmsghelper.go:40:65: undefined: syscall.IPV6_2292PKTINFO
```

cnderrauber:
#### Description
In #475, import low-level API (ReadMsgUDP) to determine
destination interface for packets received by UDPConn listen
at any(unspecified) address, to fix msg received by incorrect
candidate that shared same ufrags. But the api has compatibility
issues, also not reliable in some special network cases like
AWS/ECS.

So this pr revert that change, and make UDPMuxDefault not
accept Conn listen at unspecified address. Also provide a
NewMultiUDPMuxFromPort helper function to create a MultiUDPMux
to listen at all addresses.

For ice gather, it will use UDPMux's listen address to generate
canidates instead of create it from interfaces.

#### Reference issue
Fixes #482 #435


Sean-Der:
LGTM!

Is `make UDPMuxDefault not accept Conn listen at unspecified address.` a breaking change? Should we make it a `log.Warn` then?

When this merges we should also update `pion/webrtc` examples to use `NewMultiUDPMuxFromPort` instead? I can also make an announcement on Twitter/LinkedIn that if users are using the UDP/TCP Mux to use new API.

`pion/ice` makes a lot of incorrect assumptions around how networking works and it is starting to hurt. Probably good to do a total review of how the package works. 

thank you so much for fixing this @cnderrauber! It is a lot of hard work, really appreciate it :)

cnderrauber:
@Sean-Der Good point, this break the current user's application if unspecified UDPConns has been used (it is high probably  if user need udp mux), if we can output a warn log and keep old behavior(it is not ideal but can work), that will be best. Will find a way for that, maybe some redundant code is needed.
For pion/examples, it should update after this PR merged.

udp_mux.go
```diff
@@ -26,200 +28,150 @@ type UDPMuxDefault struct {
...
 
+	var localAddrsForUnspecified []net.Addr
+	if addr, ok := params.UDPConn.LocalAddr().(*net.UDPAddr); !ok {
+		return nil, errInvalidAddress
+	} else if ok && addr.IP.IsUnspecified() {
+		// return nil, errListenUnspecified
+		// For unspecified addresses, the correct behavior is to return errListenUnspecified, but
+		// it will break the applications that are already using unspecified UDP connection
+		// with UDPMuxDefault, so print a warn log and create a local address list for mux.
+		params.Logger.Warn("UDPMuxDefault should not listening on unspecified address, use NewMultiUDPMuxFromPort instead")
```
cnderrauber, line 68: For unspecified connection, print a warn log and generate addresses list, application using unspecified connection will work but should update to NewMultiUDPMuxFromPort

stv0g:
This PR is breaking the API as  `ice.NewUDPMuxDefault` and `ice.NewUniversalUDPMuxDefault` return an additional argument

stv0g:
Not sure if this is caused by this PR. But I get a lot of the following warnings on `GOOS=darwin`

`Discard success message from (83.243.41.189:63168), no such remote`

cnderrauber:
> Not sure if this is caused by this PR. But I get a lot of the following warnings on `GOOS=darwin`
> 
> `Discard success message from (83.243.41.189:63168), no such remote`

@stv0g  From the code, it received a stun message but can't find the remote candidate. This PR should have same behavior as ealy version, can you revert to early version and try again? (v2.2.7)

udp_mux.go
```diff
@@ -26,200 +28,150 @@ type UDPMuxDefault struct {
...
+func NewUDPMuxDefault(params UDPMuxParams) (*UDPMuxDefault, error) {
 	if params.Logger == nil {
 		params.Logger = logging.NewDefaultLoggerFactory().NewLogger("ice")
 	}
 
+	var localAddrsForUnspecified []net.Addr
+	if addr, ok := params.UDPConn.LocalAddr().(*net.UDPAddr); !ok {
+		return nil, errInvalidAddress
+	} else if ok && addr.IP.IsUnspecified() {
+		// return nil, errListenUnspecified
```
boks1971, line 64: can remove this return for now.

boks1971 (approved):
LDTM @cnderrauber . Great work!!! Some small comments.

udp_mux_universal.go
```diff
@@ -84,48 +88,44 @@ func (m *UniversalUDPMuxDefault) GetRelayedAddr(turnAddr net.Addr, deadline time
...
 		if err = msg.Decode(); err != nil {
 			c.logger.Warnf("Failed to handle decode ICE from %s: %v", addr.String(), err)
-			err = nil
+			return n, addr, nil
+		}
+
+		udpAddr, ok := addr.(*net.UDPAddr)
+		if !ok {
+			// message about this err will be logged in the UDPMux
 			return
```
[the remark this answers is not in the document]
cnderrauber: it's ok here, will pass the packet to UDPMux

cnderrauber:
@stv0g I can reproduce the `discard ...` log now, it is caused by the mixup IPv6 and v4 connections in this change, I have fix it with newest commit, can you pull and try it again?

stv0g:
@cnderrauber Just a question: How does this relate the universal UDPMux for Srflx candidates? Can we use the new MultiUDPMux also in combination with the UniversalUDPMux?

cnderrauber:
> @cnderrauber Just a question: How does this relate the universal UDPMux for Srflx candidates? Can we use the new MultiUDPMux also in combination with the UniversalUDPMux?

There is no change on UniversalUDPMux. UniversalUDPMux is  individual field `udpMuxSrflx` in ice.Agent, I think it can't combine with other mux into MultiUDPMux. This PR don't relate to that part

cnderrauber:
> @cnderrauber Thanks for the clarification :) Do you think the same originial issue which motivated this PR also is applicable to Srflx candidates? If so, I assume that we also need a UniversalMultiUDPMux? (Those names become ridiculous by now :D)

For UniversalMultiUDPMux, I think it can't work well with unspecified address on multiple interfaces too, so things like `UniversalMultiUDPMux` is needed in that case, though the API design looks not so good for these UDPMux/MultiMux/UniversalMux/MultiUniversalMux. Now we have a workaround is combine SetNATMapping with UDPMuxltiMux, you can solve each reflex IP address through stun server for each local address, and SetNATMapping for them, then use UDPMux directly.

cnderrauber:
For the srflx candidate, there is RelAddr and RelPort, can you use them to
identify the quic.Transport?

dinvlad ***@***.***> 于2023年9月30日周六 05:08写道：

cnderrauber:
The `RelateRelatedAddress()` method of Candidate will return it.

dinvlad ***@***.***> 于2023年10月6日周五 09:45写道：

PR #490, opened 2022-11-12
Cleanups

stv0g:
I've spent some time auditing the code base for gaining a better understanding.

Here are a few cleanups and refactorings.

The PR includes no functional changes.

PR #492, opened 2022-11-15
Do not show a warning when reading from a closed connection

stv0g:
Follow-up PR to #490


stv0g:
I am not exactly sure if `io.EOF` and `net.ErrClosed` are the correct error types to catch here..

cnderrauber (approved):

stv0g:
Thanks @cnderrauber

PR #497, opened 2022-11-22
Add option to include loopback candidate

### LINKED ISSUE #494: Allow ICE to open sockets listening on the loopback interface
## Summary

Allow (and force?) ICE to open sockets listening on the loopback interface.
This can be useful when lo is carrying the IP address one would want to expose.

## Motivation

My VPS host configures its VMs to have their public IPs on the loopback interface.
Therefore, when using Livekit Server, I have to set up DNAT in order to NAT incoming packets on the public IP to the private IP, on which ICE opens a socket.
It would be nice to be able to force ICE to listen on the loopback interface, in my case exposing the public IP.

DNAT as stated above :)

## Additional context

Initial discussion here, on the Livekit Server project: https://github.com/livekit/livekit/issues/1160

Thanks in advance!

cnderrauber:
#### Description
User report that VM's public ip has been mapped to loopback interface, but pion doesn't listen/collect the loopback interface cause it can't establish the connection. This PR adds an option to enable loopback interface candidates.

#### Reference issue
Fixes #494

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -130,6 +130,7 @@ type Agent struct {
 
 	interfaceFilter func(string) bool
 	ipFilter        func(net.IP) bool
+	includeLoopback bool
 
 	insecureSkipVerify bool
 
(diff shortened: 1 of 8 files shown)


davidzhao (approved):
lgtm

PR #499, opened 2022-11-22
Fix IPv6 formatting in Candidate.String()

Genteure:
A `Candidate` with IPv6 address was formatted similar to `fe80::1:51820`.
`net.JoinHostPort` will add `[` `]` around the address if it's IPv6, so it would look like `[fe80::1]:51820` instead.


stv0g (approved):

stv0g:
@Genteure Thanks :) The PR looks good to me. Can you fix the CI check which fails?

PR #441, opened 2022-04-11
Add option for accept aggressive nominate

cnderrauber:
#### Description
 Add config for accept aggressive nomination, it is useful for compatible with chrome or other clients use aggressive nominate to update nominate candidate pairs.

#### Reference issue
Fixes #440

PR #500, opened 2022-11-23
Add aggressive nomination when nominateOnBindingSuccess

San9H0:
#### Description

we are not working in case aggressive nomination in firefox in vpn 

#### Reference issue
related PR
https://github.com/pion/ice/pull/441
https://github.com/pion/ice/pull/446

I thinks Fix is missing.

Add aggressive nomination when nominateOnBindingSuccess

Now in case aggressive nomination in firefox we are working well


cnderrauber (approved):
looks good. firefox uses aggressive nomination too now? I remember it not use that when I'm testing that

cnderrauber:
@San9H0 The lint failed for the commit message format, I have updated it and merge this PR.

PR #504, opened 2022-12-07
Check if conn is closed before writing to it.

### LINKED ISSUE #503: tcp-mux: failed to send packet: io: read/write on closed pipe
When connecting to TCP MUX, having data flown though it and then disconnecting on the client side, I am flooded with `failed to send packet: io: read/write on closed pipe` until the server side is closed. It looks like, the close event is not propagated correctly.

When reverting this commit https://github.com/pion/ice/commit/7f13fd1947277496282a32bd7bffcbd21bb02247 it works as expected. That means, connection was never closed before.

Checking if connection is actually open before writing to it seems to fix this issue. Adding following code to `func (c *candidateBase) writeTo(raw []byte, dst Candidate) (int, error)`:

```go
select {
case <-c.closedCh:
	return 0, io.ErrClosedPipe
default:
}
```

I suspect that this unconditional writing might lead to race condition that causes deadlock in very rare cases (https://github.com/pion/ice/commit/5475fdb7960284145cfbe7b7bd3c8a628576fbb1), and proactively closing the channel on server made this issue more visible. I am still investigating and trying to reproduce it.

### Environment.
  - Version: v2.2.12
  - Browser: Chrome 107

m1k1o:
#### Description
Check if connection is closed before writing anything to it.

#### Reference issue
Fixes #503

### DIFF
--- a/candidate_base.go
+++ b/candidate_base.go
@@ -308,6 +308,10 @@ func (c *candidateBase) close() error {
 func (c *candidateBase) writeTo(raw []byte, dst Candidate) (int, error) {
 	n, err := c.conn.WriteTo(raw, dst.addr())
 	if err != nil {
+		// If the connection is closed, we should return the error
+		if errors.Is(err, io.ErrClosedPipe) {
+			return n, err
+		}
 		c.agent().log.Infof("%s: %v", errSendPacket, err)
 		return n, nil
 	}
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -38,6 +38,7 @@ Meelap Shah <meelapshah@gmail.com>
 Michael MacDonald <github@macdonald.cx>
 Michael MacDonald <mike.macdonald@savantsystems.com>
 Mikhail Bragin <misha@wiretrustee.com>
+Miroslav Šedivý <sedivy.miro@gmail.com>
 Nevio Vesic <nevio@subspace.com>
 Ori Bernstein <ori@eigenstate.org>
 Robert Eperjesi <eperjesi@uber.com>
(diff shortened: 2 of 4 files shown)


ashellunts:
Hi @m1k1o, is it possible to write a test for the fix?

m1k1o:
Simplified by checking the error we receive. Reason for errors not being logged I found in [this commit](https://github.com/pion/ice/commit/1a4dd6394c1a43e41caa209cd08c2741854241d0) to be `Downstream isn't able to distinguish a temporary from a Permanent error yet.` But `io.ErrClosedPipe` is permanent error, therefore it should be returned.

ashellunts (approved):
👍

candidate_base.go
```diff
@@ -308,6 +308,10 @@ func (c *candidateBase) close() error {
 func (c *candidateBase) writeTo(raw []byte, dst Candidate) (int, error) {
 	n, err := c.conn.WriteTo(raw, dst.addr())
 	if err != nil {
+		// If the connection is closed, we should return the error
+		if errors.Is(err, io.ErrClosedPipe) {
+			return n, err
+		}
 		c.agent().log.Infof("%s: %v", errSendPacket, err)
```
stv0g, line 315: Just wondering why this is using the info level? Shouldnt we print this at least as a warning?

stv0g (approved):
Looks good to me :) Just the commit message CI check is failing :(

stv0g:
@m1k1o Do you mind fixing the CI commit message check which fails?

Or, are you okay if I do so for you?

m1k1o:
I can do it later today. But feel free to do it now if you want.

stv0g:
@m1k1o I adjusted the commit messages. However, it seems like I am lacking the permissions to push the changes to your branch. :(

m1k1o:
@stv0g i added permissions, can you try again?

candidate_base.go
m1k1o: For ref: https://github.com/m1k1o/pion-ice/commit/8c1adf18f14735ea6c35e8a714392f509f84957d

PR #508, opened 2022-12-15
Add RemoteCandidateCache for better performance

San9H0:
#### Description
Check sourceAddress from first packet validation of STUNTraffic. If validation is true, store it in the cache.

Use cache for performance

#### Reference issue
In my pprof, `validateNonSTUNTraffic` takes cpu than expected. 
I propose you `validateSTUNTrafficCache `

ASIS : 5.7%
스크린샷 2022-12-15 오후 8 44 25

TOBE : 1.4%
스크린샷 2022-12-15 오후 8 46 31


stv0g:
Nice idea :) But, I think we can do better by using a map for the cache. Ideally we would use `netip.AddrPort` as the map key. But for that we need to check if we want to bump pion/ice's minimum required Go version

jech:
You don't need 1.19 in order to use socket addresses as keys:
```Go
type AddrPort [18]byte

func toAddrPort(addr UDPAddr) {
    var ap AddrPort
    copy(AddrPort, addr.IP.To16()
    ap[16] = uint8(addr.Port>>8)
    ap[17] = uint8(addr.Port)
    return ap
}
```

addr.go
```diff
@@ -46,3 +48,21 @@ func addrEqual(a, b net.Addr) bool {
...
 	return aType == bType && aIP.Equal(bIP) && aPort == bPort
 }
+
+// AddrPort is  an IP and a port number.
+type AddrPort [18]byte
+
+func toAddrPort(addr net.Addr) AddrPort {
+	var ap AddrPort
+	switch addr := addr.(type) {
+	case *net.UDPAddr:
```
stv0g, line 58: Is it possible to combine the two case branches?
E.g.;

```go
switch addr := addr.(type) {
	case *net.UDPAddr, *net.TCPAddr:
	    [...]
}
```

agent.go
```diff
@@ -1135,19 +1135,18 @@ func (a *Agent) handleInbound(m *stun.Message, local Candidate, remote net.Addr)
 
 // validateNonSTUNTraffic processes non STUN traffic from a remote candidate,
 // and returns true if it is an actual remote candidate
-func (a *Agent) validateNonSTUNTraffic(local Candidate, remote net.Addr) bool {
-	var isValidCandidate uint64
+func (a *Agent) validateNonSTUNTraffic(local Candidate, remote net.Addr) (Candidate, bool) {
+	var remoteCandidate Candidate
 	if err := a.run(local.context(), func(ctx context.Context, agent *Agent) {
```
stv0g, line 1140: I am wondering if we really need to run this in the agents task loop?
It seem to be a hot path and we might see even better performance improvements if we protect the cache with an `sync.RWMutex`

stv0g: A nevermind, I see that you already do the cache lookup outside the task loop.
But i am a bit concerned if this is really thread safe?

candidate_base.go
```diff
@@ -231,6 +233,21 @@ func (c *candidateBase) recvLoop(initializedCh <-chan struct{}) {
 	}
 }
 
+func (c *candidateBase) validateSTUNTrafficCache(addr net.Addr) bool {
+	if candidate, ok := c.remoteCandidateCaches[toAddrPort(addr)]; ok {
```
stv0g, line 237: A we sure `validateSTUNTrafficCache()` is not invoked concurrently?

addr.go
San9H0: It seems impossible to typecast two different structures with switch case

stv0g: Ah okay thanks for checking :)

candidate_base.go
San9H0: If `start function of candidateBase` was not atomic, It should be not thread safe.

I thinks it is fine because `addCandidate function of Agent` is checking that candidateBase start is atomic.
and Candidate always created before called `addCandidate`.

stv0g: Ah I see. Thanks for the clarification.

I was somehow under the impression that we have a cache shared by all candidates. But that does not make sense at all..

stv0g:
I have a last question regarding this cache:

I dont see any place where the cache is cleared again. Should we do this maybe in `start()` or `Close()`?

San9H0:
Not now.

`localCandidate and remoteCandidate of Agent` are always deleted together in `deleteAllCandidates of Agent`
I thinks that means `candidateBase` and `remoteCandidate` are deleted at same time. It is why don't need the clear yet

Later, If only `remoteCandidate of Agent` will be deleted, remoteCandidateCache also should be deleted.
It seems to be possible by adding the deletion time to the cache.

stv0g (approved):

PR #525, opened 2023-02-04
Remove redundant check for null pointer

ashellunts:
Pointer is already checked earlier in the function


stv0g (approved):

stv0g:
Thanks @ashellunts :)

PR #531, opened 2023-02-09
Refactor test pair search

ashellunts:
- Rename.
- Use a fixture to simplify the test.
- Remove call to findPair as it has no effect on the test.
- Replace assert with require, remove redundant error messages.
- Better naming.


stv0g (approved):
Great 👍

PR #533, opened 2023-02-09
Misc cleanups and refactoring

stv0g:
I started to refactor some part of the module into an internal package.


internal/atomic.go
```diff
@@ -0,0 +1,19 @@
+package internal
```
at-wat, line 1: I think it's better to add `atomic` and `fakepacketconn` (or `fakenet`?) directories

errors.go
```diff
@@ -128,7 +128,6 @@ var (
 	errTooManyColonsAddr             = errors.New("too many colons in address")
 	errRead                          = errors.New("unexpected error trying to read")
 	errUnknownRole                   = errors.New("unknown role")
-	errMismatchUsername              = errors.New("username mismatch")
```
edaniels, line 131: [side] we ought to expose some of these for proper use of `errors.Is`

edaniels (approved):
LGTM mod at-wat's suggestion.

PR #539, opened 2023-02-23
Add ServerName to tlsConfig for TLS-over-TCP

rasmusHanningOpen:
#### Description
Since the recent changes in https://github.com/pion/ice/commit/0194bd6ce840c49dd743171e6e4cab3ddf3c5beb it appears that TLS over TCP is no longer working properly unless you explicitly provide the InsecureSkipVerify flag to the tls.Config (by providing it when the agent is instantiated). The call to tls.Client() requires either ServerName or InsecureSkipVerify to be set, and since you most likely don't want to enable InsecureSkipVerify this change does what is necessary. The call to conn.HandshakeContext following this will always fail with the error `Failed to connect to relay: tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config` unless InsecureSkipVerify is enabled.

The behaviour this pull request sets for the tls.Config is also already used by the dtls.Config for DTLS over UDP in the code. The most recent version of the implementation in [pion/webrtc](https://github.com/pion/webrtc) uses this very code, meaning that, at least in my experiments, TLS over TCP does not work at all if you are using `turns`. If I am missing something, I'll gladly learn about it!

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -615,6 +615,7 @@ func (a *Agent) gatherCandidatesRelay(ctx context.Context, urls []*URL) { //noli
 				}
 
 				conn := tls.Client(tcpConn, &tls.Config{
+					ServerName:         url.Host,
 					InsecureSkipVerify: a.insecureSkipVerify, //nolint:gosec
 				})
 


Sean-Der:
We shouldn’t break behavior in minor release ever. Was `InsecureSkipVerify` enabled before?

IMO we should re-enable it so we don’t break users unexpectedly 

@stv0g

rasmusHanningOpen:
> We shouldn’t break behavior in minor release ever. Was `InsecureSkipVerify` enabled before?
> 
> IMO we should re-enable it so we don’t break users unexpectedly
> 
> @stv0g

It appears 2.3.0 must have broken behavior. 
Previously the TLS-over-TCP connection was verified through:

`conn, connectErr := tls.Dial(NetworkTypeTCP4.String(), TURNServerAddr, &tls.Config{
	InsecureSkipVerify: a.insecureSkipVerify, 
})`

which means the insecureSkipVerify was used the same way as right now, but the difference is that tls.Dial implicity infers the ServerName from the provided URL if is not set in the tls.Config. When calling tls.Client and then client.HandshakeContext nobody will infer the ServerName from the URL so it needs to be explicitly set.

rasmusHanningOpen:
Commit message lint error should be fixed now.

Sean-Der:
@rasmusHanningOpen thanks for fixing this!

merging now

Sean-Der:
@rasmusHanningOpen also! I really appreciate you upstreaming this. If you are interested would love to have your continued involvement :)

We are all on https://pion.ly/slack and would love your feedback/input on how to make the project better (as a user)

Sean-Der (approved):

Sean-Der:
@rasmusHanningOpen tagged `v3.1.56` in pion/webrtc

PR #542, opened 2023-03-02
Fix packet out-of-order with udpMuxed

ziminghua:
#### Description
I get out-of-order packet when use udpMuxed, and linstener addr is 0.0.0.0. I find I have three candidate, this create three `recvLoop` goroutine,  and `c.conn`  create by `ufrag`. The three candidate use same `ufrage`, then share same buffer.
so,  may cause the problem of first read but last write when parallel, cause out-of-order.
I try to fix this bug, but there is no good way. I can only lock the connection. Do you have any better way? Or use this? Thank you.

PS:我的英文不太好，拿中文在描述一遍.
我在测试过程中使用UDP的端口复用，并且监听的地址为0.0.0.0，发现我有3个candidate。根据代码会创建3个recvLoop的协程。并且他们引用的是同一个udpMuxedConn对象。因为这个conn是通过ufrag创建的，他们会共享同一个buffer。所以在并发时会造成先读到的buffer内容但是后写入Agent的buffer，造成Packet的乱序。我不知道怎么修改这个问题比较好。只是采用了对conn加锁，控制读取和写入顺序来解决，如果有什么更好办法，请帮忙修复它。谢谢

#### Reference issue
Fixes #...

### DIFF
--- a/candidate_base.go
+++ b/candidate_base.go
@@ -207,7 +207,6 @@ func (c *candidateBase) start(a *Agent, conn net.PacketConn, initializedCh <-cha
 }
 
 func (c *candidateBase) recvLoop(initializedCh <-chan struct{}) {
-	a := c.agent()
 
 	defer close(c.closedCh)
 
@@ -219,16 +218,28 @@ func (c *candidateBase) recvLoop(initializedCh <-chan struct{}) {
 
 	buf := make([]byte, receiveMTU)
 	for {
-		n, srcAddr, err := c.conn.ReadFrom(buf)
-		if err != nil {
-			if !(errors.Is(err, io.EOF) || errors.Is(err, net.ErrClosed)) {
-				a.log.Warnf("Failed to read from candidate %s: %v", c, err)
-			}
+		if err := c.readAndHandlePacket(buf); err != nil {
 			return
 		}
-
-		c.handleInboundPacket(buf[:n], srcAddr)
 	}
+}
+
+func (c *candidateBase) readAndHandlePacket(buf []byte) error {
+	a := c.agent()
+	muxedConn, isUdpMuxed := c.conn.(*udpMuxedConn)
+	if isUdpMuxed {
+		muxedConn.readMu.Lock()
+		muxedConn.readMu.Unlock()
+	}
+	n, srcAddr, err := c.conn.ReadFrom(buf)
+	if err != nil {
+		if !(errors.Is(err, io.EOF) || errors.Is(err, net.ErrClosed)) {
+			a.log.Warnf("Failed to read from candidate %s: %v", c, err)
+		}
+	}
+
+	c.handleInboundPacket(buf[:n], srcAddr)
+	return err
 }
 
 func (c *candidateBase) handleInboundPacket(buf []byte, srcAddr net.Addr) {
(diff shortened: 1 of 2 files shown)


Sean-Der:
Hi @ziminghua 

What problems are out of order packets causing? SCTP and RTP/RTCP should both be able to handle that.

ziminghua:
I find RTP

ziminghua:
> Hi @ziminghua
> 
> What problems are out of order packets causing? SCTP and RTP/RTCP should both be able to handle that.

RTP packet's sequence out of order cause after misbelieve that the packet is lost and send NACK.

ziminghua:
@Sean-Der We develop a interceptor need check sequence continuation, so I think this may be a bug.

ziminghua:
@Sean-Der  Sorry,to bother you, but this problem has troubled me for a long time. I need your help. Thank you. :)

ziminghua:
I finally found the reason. My test environment has three network ports, and the three network ports are all connected. At this time, three polls will match the same cache.  It's up to you to assess whether need repaired.

stv0g:
I would say that ICE/UDP has no ordering guarantees. This is handled by upper layers (e.g. SCTP). Hence I would say this is a wont-fix?

ziminghua:
> I would say that ICE/UDP has no ordering guarantees. This is handled by upper layers (e.g. SCTP). Hence I would say this is a wont-fix?

I very much agree this is handled by upper layers , but I don't know how to fix it in upper layer.  The upper layer is `udpMuxedConn` with use `ICEUDPMux`, but many candidate readloop share same udpMuxedConn, so  when writing packet in parallel to the lower buffer, it will cause out-of-order.

candidate_base.go
```diff
@@ -219,16 +218,28 @@ func (c *candidateBase) recvLoop(initializedCh <-chan struct{}) {
...
+	}
+}
 
-		c.handleInboundPacket(buf[:n], srcAddr)
+func (c *candidateBase) readAndHandlePacket(buf []byte) error {
+	a := c.agent()
+	muxedConn, isUdpMuxed := c.conn.(*udpMuxedConn)
+	if isUdpMuxed {
+		muxedConn.readMu.Lock()
+		muxedConn.readMu.Unlock()
```
stv0g, line 232: Just for understanding this PR: are you missing a `defer` here?

stv0g:
> so when writing packet in parallel to the lower buffer, it will cause out-of-order.

I am not sure if I understand the issue completely.. I do not think that the candidates should share the same buffer? Woulnd't this cause a data race?

stv0g:
I also think that there is another issue: `UDPMux` is currently an interface. Hence your type assertion on `udpMuxConn` currently only works for the default `UDPMux` implementation which is included in pion/ice. But how about other implementations of `UDPMux`? (I have one of those in my project)

candidate_base.go
ziminghua: I fixed it, but I don't think it is a good idea. Do you have anything other batter way?

ziminghua:
> I also think that there is another issue: `UDPMux` is currently an interface. Hence your type assertion on `udpMuxConn` currently only works for the default `UDPMux` implementation which is included in pion/ice. But how about other implementations of `UDPMux`? (I have one of those in my project)

There are a lot of triggering conditions, which require Linstener addr 0.0.0.0 and multiple network interface. Multiple network of client and Pion can be connected. The problem of sharing a 'udpMuxConn' will occur. So I think this may not be a problem, but the actual situation may not happen.

ziminghua:
I hope there is a better way to solve this problem. At the same time, it is better to support PION to PION with udpMuxConn

ziminghua:
You can close this issue.   :)

stv0g:
Have you seen the `MultiUDPMux`? It creates a seperate `UDPMux` for each local interface. I think it could solve your issue as well.

PR #545, opened 2023-03-27
Fix gathering candidates race condition

### LINKED ISSUE #445: Close 'closed channel' cause panic
### Your environment.
  - Version: *pion/ice/v2@v2.2.3*
  - Browser: *Chrome newest version*
  - Other Information - *using simulcast feature*

### What did you do?
I'm trying to re-handshake peerconnection between `pion` and `chrome` continuously

### What happened?
I think the root cause is `gatherCandidateDone` channel was closed before but still to do it again.

Screen Shot 2022-04-22 at 16 27 26

Screen Shot 2022-04-22 at 16 21 44

m1k1o:
#### Description
Canceling previous gathering does not happen synchronously. If it takes a little bit longer, it can close newly created `gatherCandidateDone` channel.  If we pass channel to the `gatherCandidates` function, it will always close its done channel and only the latest done channel will be available in `gatherCandidateDone`.

#### Reference issue
Fixes #445

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -51,17 +51,18 @@ func (a *Agent) GatherCandidates() error {
 		a.gatherCandidateCancel() // Cancel previous gathering routine
 		ctx, cancel := context.WithCancel(ctx)
 		a.gatherCandidateCancel = cancel
-		a.gatherCandidateDone = make(chan struct{})
+		done := make(chan struct{})
+		a.gatherCandidateDone = done
 
-		go a.gatherCandidates(ctx)
+		go a.gatherCandidates(ctx, done)
 	}); runErr != nil {
 		return runErr
 	}
 	return gatherErr
 }
 
-func (a *Agent) gatherCandidates(ctx context.Context) {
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):

Sean-Der:
Nice job @m1k1o 

Do you think it would be possible to get a crash for the old implementation? If you try to gather STUN on `127.0.0.1` and have it timeout for 30 seconds I bet you could get it to close twice?

m1k1o:
@Sean-Der i can reproduce it by adding another `assert.NoError(t, a.GatherCandidates())` after:
https://github.com/pion/ice/blob/be72bc709bb290aba9bdc842988d2f893300d967/gather_test.go#L159

image

I'll extract it to custom minimal unit test.

Sean-Der:
@m1k1o tagged!

Renovate Bot should open a PR agaist https://github.com/pion/webrtc soon, feel free to merge that when it passes! Then can tag a new pion/webrtc

PR #550, opened 2023-04-12
Fix regression with UDPMuxDefault and NAT1To1IPs

### LINKED ISSUE #505: udp-mux: RemoveConnByUfrag removes existing connection when having 1:1 NAT mapping
In recent commit https://github.com/pion/ice/commit/7f13fd1947277496282a32bd7bffcbd21bb02247 closing connection on duplicate candidates has been added.

E.g. we have 2 `localAddresses`, and `extIPMapper` as `CandidateTypeHost` with `ipSole`. This means, both of our local addresses will be overwritten with a single IP, here:
https://github.com/pion/ice/blob/ea00dc7424d06d6ed4c642639d59bcdf41156d0a/gather.go#L284-L291

When we continue [adding candidate](https://github.com/pion/ice/blob/ea00dc7424d06d6ed4c642639d59bcdf41156d0a/gather.go#L310), we check inside for duplicates. It adds first candidate and when processing second candidate, [it notices](https://github.com/pion/ice/blob/ea00dc7424d06d6ed4c642639d59bcdf41156d0a/agent.go#L825), there is already existing candidate with given addr so attempts to close it.

After closing candidate we do not need, [this cleanup task](https://github.com/pion/ice/blob/ea00dc7424d06d6ed4c642639d59bcdf41156d0a/udp_mux.go#L150-L153) will be executed, that attempts to remove conn by ufrag. But since both connections shared the same ufrag, it removes the first connection instead.

For tcp_mux it is not an issue, because it is [identified by local host](https://github.com/pion/ice/blob/ea00dc7424d06d6ed4c642639d59bcdf41156d0a/tcp_mux.go#L172) as well.

I tried to implement [the same per-addr separation](https://github.com/pion/ice/files/10190227/udp_mux.go.txt) as there is for tcp_mux, but since udp works differently, i was not able to get reliably correct addr when reading packets. Easier solution would be just to add [duplicate check for udp_mux ips](https://github.com/pion/ice/commit/23d31f85061f826331cc017fc88ca767d7071ceb).

### Your environment.
  - Version: v2.2.12

aler9:
#### Description

PR #496 added a mechanism to close connections of duplicate candidates and avoid memory leaks. When UDPMuxDefault and NAT1To1IPs are used together, candidates may share the same connection, and candidates may be equal, since their IPs are overridden by the one specified in NAT1To1IPs. When a duplicate candidate is found, its connection is closed, making impossible to use any other candidate that shares the same connection.

This PR fixes the issue as it prevents duplicate candidates from creating connections and getting their connection closed.

#### Reference issue

Fixes #505

Related to https://github.com/aler9/mediamtx/issues/1680

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -255,6 +255,7 @@ func (a *Agent) gatherCandidatesLocalUDPMux(ctx context.Context) error { //nolin
 	}
 
 	localAddresses := a.udpMux.GetListenAddresses()
+	existingConfigs := make(map[CandidateHostConfig]struct{})
 
 	for _, addr := range localAddresses {
 		udpAddr, ok := addr.(*net.UDPAddr)
@@ -272,15 +273,24 @@ func (a *Agent) gatherCandidatesLocalUDPMux(ctx context.Context) error { //nolin
 			candidateIP = mappedIP
 		}
 
-		conn, err := a.udpMux.GetConn(a.localUfrag, udpAddr)
-		if err != nil {
-			return err
-		}
 		hostConfig := CandidateHostConfig{
 			Network:   udp,
 			Address:   candidateIP.String(),
 			Port:      udpAddr.Port,
 			Component: ComponentRTP,
+		}
+
+		// Detect a duplicate candidate before calling addCandidate().
+		// otherwise, addCandidate() detects the duplicate candidate
+		// and close its connection, invalidating all candidates
+		// that share the same connection.
+		if _, ok := existingConfigs[hostConfig]; ok {
+			continue
+		}
+
+		conn, err := a.udpMux.GetConn(a.localUfrag, udpAddr)
+		if err != nil {
+			return err
 		}
 
 		c, err := NewCandidateHost(&hostConfig)
@@ -297,6 +307,8 @@ func (a *Agent) gatherCandidatesLocalUDPMux(ctx context.Context) error { //nolin
 			closeConnAndLog(conn, a.log, fmt.Sprintf("Failed to add candidate: %s %d: %v", candidateIP, udpAddr.Port, err))
 			continue
 		}
+
+		existingConfigs[hostConfig] = struct{}{}
 	}
 
 	return nil
--- a/gather_test.go
+++ b/gather_test.go
@@ -599,6 +599,51 @@ func TestTURNProxyDialer(t *testing.T) {
 	assert.NoError(t, a.Close())
 }
 
+// TestUDPMuxDefaultWithNAT1To1IPsUsage asserts that candidates
+// are given and connections are valid when using UDPMuxDefault and NAT1To1IPs.
+func TestUDPMuxDefaultWithNAT1To1IPsUsage(t *testing.T) {
+	report := test.CheckRoutines(t)
+	defer report()
+
+	lim := test.TimeOut(time.Second * 30)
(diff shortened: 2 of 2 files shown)


gather.go
```diff
@@ -272,17 +273,27 @@ func (a *Agent) gatherCandidatesLocalUDPMux(ctx context.Context) error { //nolin
...
-			return err
-		}
 		hostConfig := CandidateHostConfig{
 			Network:   udp,
 			Address:   candidateIP.String(),
 			Port:      udpAddr.Port,
 			Component: ComponentRTP,
 		}
 
+		// detect a duplicate candidate before calling addCandidate().
```
stv0g, line 283: Please start comments with an uppercase letter.

gather_test.go
```diff
@@ -599,6 +599,47 @@ func TestTURNProxyDialer(t *testing.T) {
 	assert.NoError(t, a.Close())
 }
 
+// Assert that candidates are given and connections are valid
```
stv0g, line 602: Please start function comments in the form:

```go
// TestUDPMuxDefaultWithNAT1To1IPsUsage asserts that ...
```

stv0g (changes requested):

stv0g:
Thanks @aler9! Thats great :) I have two smaller comments on your PR.

And, I've noticed that the PR is based on an outdated `master` branch.
Could you rebase your changes on the latest `master`?

Thanks a lot :)

stv0g:
I also assume that this will fix https://github.com/pion/ice/issues/505?

@aler9 could you confirm this?

aler9:
Hello @stv0g, i have applied all the requested changes and i can confirm that this fixes #505. Furthermore, the PR adds a unit test with the exact use case (UDPDefaultMux + NAT1To1).

The question is that UDPMux is used in two ways:

* with `NewMultiUDPMuxFromPort` like [livekit](https://github.com/livekit/livekit) and @cnderrauber do. This caused the memory leak that was solved with #496
* with `NewUDPMuxDefault` like [mediamtx](https://github.com/aler9/mediamtx) and i do. This has the advantage of not having to scan interfaces manually, but stopped working after #496

Personally i don't have a preference and it's not my role to decide, but i'd like for both methods to continue working in order not to break existing code.

stv0g (approved):

stv0g:
Thanks @aler9,

It looks good now. We can ignore the failing CI tests for Go 1.18.

I will @cnderrauber some time to also review the change.
But from my side, we could merge it 👍

cnderrauber (commented):
Thanks for improving this!

gather.go
```diff
@@ -272,17 +273,27 @@ func (a *Agent) gatherCandidatesLocalUDPMux(ctx context.Context) error { //nolin
...
 		}
 
+		// Detect a duplicate candidate before calling addCandidate().
+		// otherwise, addCandidate() detects the duplicate candidate
+		// and close its connection, invalidating all candidates
+		// that share the same connection.
+		if _, ok := configs[hostConfig]; ok {
+			continue
+		}
+		configs[hostConfig] = struct{}{}
```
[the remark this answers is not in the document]
aler9: thanks, i moved the line after addCandidate()

cnderrauber (approved):

PR #225, opened 2020-07-03
Refactor agent task runner

at-wat:
- Refactor agent task runner
  - Control blocking tasks by context
  - Add tests to check deadlock by calling Close/Restart in state change callback
  - Remove unnecessary chan
- Fix deadlock in SelectedCandidatePairChange
  - Call handler in different routine

PR #563, opened 2023-05-05
Simplify Agent.startOnConnectionStateChangeRoutine()

### LINKED ISSUE #556: Rename Agent.startOnConnectionStateChangeRoutine() to Agent.startRoutines()
Currently the method appears to be doing more than starting the `OnConnectionStateChange` routine.
I propose to rename the method to `startRoutines()`:

https://github.com/pion/ice/blob/5ef81b65f45a6c05f3553385d4b40e32b0601243/agent.go#L422

stv0g:
Fixes #556

### DIFF
--- a/agent_handlers.go
+++ b/agent_handlers.go
@@ -41,39 +41,20 @@ func (a *Agent) onConnectionStateChange(s ConnectionState) {
 	}
 }
 
-func (a *Agent) startOnConnectionStateChangeRoutine() {
-	go func() {
-		for {
-			// CandidatePair and ConnectionState are usually changed at once.
-			// Blocking one by the other one causes deadlock.
-			p, isOpen := <-a.chanCandidatePair
-			if !isOpen {
-				return
-			}
-			a.onSelectedCandidatePairChange(p)
-		}
-	}()
-	go func() {
-		for {
-			select {
-			case s, isOpen := <-a.chanState:
-				if !isOpen {
-					for c := range a.chanCandidate {
-						a.onCandidate(c)
-					}
-					return
-				}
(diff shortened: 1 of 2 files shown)


agent_handlers.go
```diff
@@ -41,39 +41,20 @@ func (a *Agent) onConnectionStateChange(s ConnectionState) {
 	}
 }
 
-func (a *Agent) startOnConnectionStateChangeRoutine() {
-	go func() {
-		for {
-			// CandidatePair and ConnectionState are usually changed at once.
-			// Blocking one by the other one causes deadlock.
```
at-wat, line 48: Are these comments ok to be removed?

stv0g: Thanks, no probably not.. The agents behaviour in respect to calling handlers is anyway pretty strange to me.. :(
E.g. do you have an explanation why the connection state handler is called asynchronously in its own Goroutine, while other handlers arnt?

https://github.com/pion/ice/pull/563/files#diff-033dc53d3d924a20f1f9a13378ce32d1461740b98aab338e3cc891bcdca387abR52


I will add the comments again..

at-wat: It seems be added by myself in this PR: https://github.com/pion/ice/pull/225
iirc, there was a deadlock bug before that and I fixed it

stv0g: I've added the comments back here https://github.com/pion/ice/pull/563/files#diff-07afbb40d3dc2699797d28c0689ba544f67ca642d0d3efcff3c02a0bbb3604fdR362

at-wat:
reopening PR to trigger CI

at-wat (approved):

PR #564, opened 2023-05-07
Extract TestOnSelectedCandidatePairChange to its own file

ashellunts:
Also made refactoring of the test.


internal/fakenet/mock_conn.go
stv0g, line 1: I am wondering whether it would be useful to keep this in `pion/transport`?

stv0g (approved):
LGTM :) But I am wondering. Is this PR just refactoring, or do you also try to, or fix some of the issues related to the wrong candidate selection?

internal/fakenet/mock_conn.go
ashellunts: I would keep it in ice package, because it is only used here at the moment.

PR #565, opened 2023-05-09
Support active TCP candidates (RFC6544)

ashellunts:
This PR implements support for active TCP candidates as specified in [RFC6544](https://datatracker.ietf.org/doc/html/rfc6544)

Open issue: local active TCP candidate can be connected only with 1 remote.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -71,6 +72,8 @@ type Agent struct {
 	srflxAcceptanceMinWait time.Duration
 	prflxAcceptanceMinWait time.Duration
 	relayAcceptanceMinWait time.Duration
+
+	tcpPriorityOffset uint16
 
 	portMin uint16
 	portMax uint16
@@ -585,6 +588,44 @@ func (a *Agent) getBestValidCandidatePair() *CandidatePair {
 }
 
 func (a *Agent) addPair(local, remote Candidate) *CandidatePair {
+	if local.TCPType() == TCPTypeActive && remote.TCPType() == TCPTypeActive {
+		return nil
+	}
+
+	if local.TCPType() == TCPTypeActive && remote.TCPType() == TCPTypePassive {
+		addressToConnect := net.JoinHostPort(remote.Address(), strconv.Itoa(remote.Port()))
+
+		conn, err := a.net.Dial("tcp", addressToConnect)
+		if err != nil {
+			a.log.Errorf("Failed to dial TCP address %s: %v", addressToConnect, err)
+			return nil
+		}
+
+		packetConn := newTCPPacketConn(tcpPacketParams{
+			ReadBuffer: tcpReadBufferSize,
+			LocalAddr:  conn.LocalAddr(),
+			Logger:     a.log,
+		})
+
+		if err = packetConn.AddConn(conn, nil); err != nil {
+			a.log.Errorf("Failed to add TCP connection: %v", err)
+			return nil
+		}
+
+		localAddress, ok := conn.LocalAddr().(*net.TCPAddr)
+		if !ok {
+			a.log.Errorf("Failed to cast local address to TCP address")
+			return nil
+		}
+
+		localCandidateHost, ok := local.(*CandidateHost)
+		if !ok {
+			a.log.Errorf("Failed to cast local candidate to CandidateHost")
+			return nil
+		}
+		localCandidateHost.port = localAddress.Port // this causes a data race with candidateBase.Port()
+		local.start(a, packetConn, a.startedCh)
+	}
 	p := newCandidatePair(local, remote, a.isControlling)
 	a.checklist = append(a.checklist, p)
 	return p
@@ -761,7 +802,9 @@ func (a *Agent) addCandidate(ctx context.Context, c Candidate, candidateConn net
 			}
 		}
 
-		c.start(a, candidateConn, a.startedCh)
+		if c.TCPType() != TCPTypeActive {
+			c.start(a, candidateConn, a.startedCh)
+		}
 
 		set = append(set, c)
 		a.localCandidates[c.NetworkType()] = set
@@ -1029,6 +1072,11 @@ func (a *Agent) handleInbound(m *stun.Message, local Candidate, remote net.Addr)
 				return
 			}
 
+			remoteTCPType := TCPTypeUnspecified
+			if local.TCPType() == TCPTypePassive {
+				remoteTCPType = TCPTypeActive
+			}
+
 			prflxCandidateConfig := CandidatePeerReflexiveConfig{
 				Network:   networkType.String(),
 				Address:   ip.String(),
@@ -1036,6 +1084,7 @@ func (a *Agent) handleInbound(m *stun.Message, local Candidate, remote net.Addr)
 				Component: local.Component(),
 				RelAddr:   "",
 				RelPort:   0,
+				TCPType:   remoteTCPType,
 			}
 
 			prflxCandidate, err := NewCandidatePeerReflexive(&prflxCandidateConfig)
--- a/agent_active_tcp_test.go
+++ b/agent_active_tcp_test.go
@@ -1,0 +1,162 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+//go:build !js
+// +build !js
+
+package ice
+
+import (
+	"net"
+	"testing"
+	"time"
+
+	"github.com/pion/logging"
+	"github.com/pion/transport/v2/stdnet"
+	"github.com/pion/transport/v2/test"
+	"github.com/stretchr/testify/require"
+)
+
+func getLocalIPAddress(t *testing.T, networkType NetworkType) net.IP {
+	net, err := stdnet.NewNet()
+	require.NoError(t, err)
+	localIPs, err := localInterfaces(net, nil, nil, []NetworkType{networkType}, false)
+	require.NoError(t, err)
+	require.NotEmpty(t, localIPs)
+	return localIPs[0]
+}
+
+func ipv6Available(t *testing.T) bool {
+	net, err := stdnet.NewNet()
+	require.NoError(t, err)
+	localIPs, err := localInterfaces(net, nil, nil, []NetworkType{NetworkTypeTCP6}, false)
+	require.NoError(t, err)
+	return len(localIPs) > 0
+}
+
+func TestAgentActiveTCP(t *testing.T) {
+	report := test.CheckRoutines(t)
+	defer report()
+
+	lim := test.TimeOut(time.Second * 5)
+	defer lim.Stop()
+
+	const listenPort = 7686
+	type testCase struct {
+		name                    string
+		networkTypes            []NetworkType
+		listenIPAddress         net.IP
+		selectedPairNetworkType string
+	}
+	testCases := []testCase{
+		{
+			name:                    "TCP4 connection",
+			networkTypes:            []NetworkType{NetworkTypeTCP4},
+			listenIPAddress:         getLocalIPAddress(t, NetworkTypeTCP4),
+			selectedPairNetworkType: tcp,
+		},
+		{
+			name:                    "UDP is preferred over TCP4", // fails some time
+			networkTypes:            supportedNetworkTypes(),
+			listenIPAddress:         getLocalIPAddress(t, NetworkTypeTCP4),
+			selectedPairNetworkType: udp,
+		},
+	}
+
+	if ipv6Available(t) {
+		tcpv6Cases := []testCase{
+			{
+				name:                    "TCP6 connection",
+				networkTypes:            []NetworkType{NetworkTypeTCP6},
+				listenIPAddress:         getLocalIPAddress(t, NetworkTypeTCP6),
+				selectedPairNetworkType: tcp,
+			},
+			{
+				name:                    "UDP is preferred over TCP6", // fails some time
+				networkTypes:            supportedNetworkTypes(),
+				listenIPAddress:         getLocalIPAddress(t, NetworkTypeTCP6),
+				selectedPairNetworkType: udp,
+			},
+		}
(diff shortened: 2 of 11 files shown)


agent.go
```diff
@@ -138,6 +139,7 @@ type Agent struct {
 
 	interfaceFilter func(string) bool
 	ipFilter        func(net.IP) bool
+	ActiveTCP       bool
```
stv0g, line 142: As you mentioned in your PR description, we should use something like a bitmask here.
Ideally using the bit positions of the [`TCPType` type](https://github.com/pion/ice/blob/9afe3ad2d02084a6d1153d98c19308357e44f5e5/tcptype.go#L10).

E.g. 

```go
type AgentConfig struct {
    ...
    TCPTypes int
    ....
}


cfg := AgentConfig{
    TCPTypes: 1<<TCPTypeActive | 1<<TCPTypePassive
}
```

agent_config.go
```diff
@@ -144,6 +144,8 @@ type AgentConfig struct {
 	// the ips which are used to gather ICE candidates.
 	IPFilter func(net.IP) bool
 
+	ActiveTCP bool
```
stv0g, line 147: See comment from above in `agent.go`

agent_active_tcp_test.go
```diff
@@ -0,0 +1,94 @@
...
+	})
+
+	defer func() {
+		_ = tcpMux.Close()
+	}()
+
+	r.NotNil(tcpMux.LocalAddr(), "tcpMux.LocalAddr() is nil")
+
+	passiveAgent, err := NewAgent(&AgentConfig{
+		TCPMux:          tcpMux,
```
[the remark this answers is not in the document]
ashellunts: That is also a question I was asking myself. I have not found a way to use TCP without a TCPMux.

gather.go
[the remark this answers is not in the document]
ashellunts: That is a good hint, thanks. I can try to implement that in next steps.

gather.go
```diff
@@ -165,44 +165,49 @@ func (a *Agent) gatherCandidatesLocal(ctx context.Context, networkTypes []Networ
...
-				var muxConns []net.PacketConn
-				if multi, ok := a.tcpMux.(AllConnsGetter); ok {
-					a.log.Debugf("GetAllConns by ufrag: %s", a.localUfrag)
-					muxConns, err = multi.GetAllConns(a.localUfrag, mappedIP.To4() == nil, ip)
-					if err != nil {
-						a.log.Warnf("Failed to get all TCP connections by ufrag: %s %s %s", network, ip, a.localUfrag)
-						continue
-					}
+				if a.ActiveTCP {
+					conns = append(conns, connAndPort{nil, 0})
```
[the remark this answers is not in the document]
ashellunts: Agree, that is also needed in order to support both active, passive candidates at the same time.

ashellunts:
Another thing I am thinking. Would it be a good idea to test in different network conditions using vnet? @stv0g

stv0g:
@ashellunts Unfortuantely, `transport/vnet` has no TCP support yet: https://github.com/pion/transport/issues/27

I also dont expect this to land soon. Hence, we should proceed without vnet tests for now..

Sean-Der:
Amazing work @ashellunts so happy to see this land. Huge thank you @stv0g for being a great reviewer :)

Passive requires using a Mux. I think that makes sense. TCP is only preferable in restrictive environments. Those environments usually only allow a few TCP ports. 

I think we can put this behind Type+NetworkTypes in Config! If Host+TCP is enabled we should have Active TCP.

My only call out is we should prefer Host+Srflx UDP over TCP. TCP I think is preferable to Relay IMO

I would love to see this land ASAP! If it has bugs/TODOs we can always fix those :)

stv0g:
Hi @Sean-Der,

thanks for your feedback!

> TCP is only preferable in restrictive environments. Those environments usually only allow a few TCP ports.

I assumed that we will see the passive side usually deployed at a Pion agent running at some public server.
In that case the active agents would be running in a restricted network.

> My only call out is we should prefer Host+Srflx UDP over TCP. TCP I think is preferable to Relay IMO

Good point. Actually [RFC6544 Section 4.2](https://datatracker.ietf.org/doc/html/rfc6544#section-4.2) specifies this behaviour:

> The transport protocol itself is a criteria for choosing one candidate over another.
> If a particular media stream can run over UDP or TCP, the UDP candidates might be preferred over the TCP candidates.
> This allows ICE to use the lower latency UDP connectivity if it exists but fallback to TCP if UDP doesn't work.

(The RFC continues to elaborate here...)

> I would love to see this land ASAP! If it has bugs/TODOs we can always fix those :)

I agree :) We just need to get the API right :-S

ashellunts:
> I think we can put this behind Type+NetworkTypes in Config! If Host+TCP is enabled we should have Active TCP.

Currently passive candidates are collected when Host+TCP is used. Do you suggest to enable active ones always together with passive ones, when Host+TCP is enabled?

Does it make sense to have option to choose form following cases:
- only passive ones (as it is right now); this one allows to disable active candidates so no changes in behavior for library clients;
- only active ones; for client-side applications;
- both active and passive; not sure use case here. @stv0g you had some ideas regarding this. Is it server to server communication?

Sean-Der:
> Currently passive candidates are collected when Host+TCP is used. Do you suggest to enable active ones always together with passive ones, when Host+TCP is enabled?

I think we should do this.

Enable Active by default. If AgentConfig has `ICECandidateTypeHost` and `NetworkTypeTCP` (the default values) we should connect to remote Passives.

This will have zero effect on most users today. If the remote only publishes UDP candidates no TCP connections will be made. If the remote publishes a TCP candidate we will attempt a connection.

---

Leave Passive disabled (same behavior we have now). Don't modify any passive UDP code and leave UDP Mux etc.. as it is.

Passive candidates require special configuration. You usually want them on :80 or :443 and you want all your ICE Agents sharing that. If we just listen on random TCP ports I don't think we are giving users the best experience possible.

I also believe that listening on a new random TCP Port would be confusing to users.

gather.go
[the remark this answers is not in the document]
Sean-Der: @ashellunts 

Is it possible to do

```
if a.ActiveTCP {
  ...
  ...
  break
}
```

Makes the code a little easier to follow/smaller diff I think

stv0g:
> Enable Active by default.

👍 

> Leave Passive disabled (same behavior we have now).

👍

stv0g:
> Both active and passive; not sure use case here. @stv0g you had some ideas regarding this. Is it server to server communication?

I think also local LAN connectivity can profit from TCP candidates if requested.
Using TCP candidates can have also its advantages based on the use case (e.g. file transfers instead of video streaming).

And while pion/ice is mainly used for pion/webrtc, WebRTC is not the only use case (I have others :D)

ashellunts:
OK, let me try to summarize:
- ActiveTCP flag is removed, so no API changes needed.
- If ICECandidateTypeHost and NetworkTypeTCP is set, active TCP candidate is generated.
- If (ICECandidateTypeHost and NetworkTypeTCP is set) AND TCPMux is set, both active and passive candidates are generated.
- We shall check that UDP candidates pair (any type) have more priority than TCP one.

agent.go
ashellunts: As discussed, active candidates will be enabled by default.

gather.go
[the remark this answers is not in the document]
ashellunts: Extracted passive ICE handling to make code easier to follow.

candidatetype.go
```diff
@@ -38,7 +38,21 @@ func (c CandidateType) String() string {
 // The RECOMMENDED values are 126 for host candidates, 100
 // for server reflexive candidates, 110 for peer reflexive candidates,
 // and 0 for relayed candidates.
-func (c CandidateType) Preference() uint16 {
+func (c CandidateType) Preference(networkType NetworkType) uint16 {
```
stv0g, line 41: I dont think this exactly reflects the recommendation of [RFC6544 Section 4.2](https://www.rfc-editor.org/rfc/rfc6544.html#section-4.2):

> When both UDP and TCP candidates are offered for the same media stream, and one transport protocol should be preferred over the other, the type preferences for the preferred transport protocol candidates SHOULD be increased and/or the type preferences for the other transport protocol candidates SHOULD be decreased.

So I think we should have a user-configurable offset which is added/substracted from the type-dependent preference.

> How much the values should be increased or decreased depends on whether it is more important to choose a certain transport protocol or a certain candidate type. If the candidate type is more important (e.g., even if UDP is preferred, TCP host candidates are preferred over UDP server reflexive candidates) changing type preference values by one for the other transport protocol candidates is enough.  On the other hand, if the transport protocol is more important (e.g., any UDP candidate is preferred over any TCP candidate), all the preferred transport protocol candidates SHOULD have type preference higher than the other transport protocol candidates.  However, it is RECOMMENDED that the relayed candidates are still preferred lower than the other candidate types. For RTP-based media streams, it is RECOMMENDED that UDP candidates are preferred over TCP candidates.

gather.go
```diff
@@ -225,7 +199,7 @@ func (a *Agent) gatherCandidatesLocal(ctx context.Context, networkTypes []Networ
 					Address:   address,
 					Port:      connAndPort.port,
 					Component: ComponentRTP,
-					TCPType:   tcpType,
+					TCPType:   connAndPort.tcpType,
```
stv0g, line 202: Can we rename this simply to `cfg`? And the corresponding `conns` to `cfgs`?

agent.go
```diff
@@ -1023,13 +1064,18 @@ func (a *Agent) handleInbound(m *stun.Message, local Candidate, remote net.Addr)
 				return
 			}
 
+			tcpType := TCPTypeUnspecified
+			if networkType == NetworkTypeTCP4 && local.NetworkType() == NetworkTypeTCP4 && local.TCPType() == TCPTypePassive {
```
stv0g, line 1068: How about IPv6?

```go
if networkType.IsTCP() && networkType == local.NetworkType() && local.TCPType() == TCPTypePassive {
```

stv0g (changes requested):
Hey, good work :)

I think we are not yet following the RFC in terms of type and local preferences.

agent.go
```diff
@@ -579,6 +580,44 @@ func (a *Agent) getBestValidCandidatePair() *CandidatePair {
...
 
 func (a *Agent) addPair(local, remote Candidate) *CandidatePair {
+	if local.TCPType() == TCPTypeActive && remote.TCPType() == TCPTypeActive {
+		return nil
+	}
+
+	if local.TCPType() == TCPTypeActive && remote.TCPType() == TCPTypePassive {
+		addressToConnect := net.JoinHostPort(remote.Address(), strconv.Itoa(remote.Port()))
+
+		conn, err := net.Dial("tcp", addressToConnect)
```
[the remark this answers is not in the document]
ashellunts: done

agent.go
ashellunts: done, simplified to 
`if local.TCPType() == TCPTypePassive {`

becase local and remote network types should be equal anyway, and if tcpType is set it means it is TCP

ashellunts:
I have added tests that check if both TCP/UDP are enabled, UDP is selected pair but the test fails sometime. Not sure why... any ideas? @stv0g 

Looks like if TCP is selected at first, later it is not changed to UDP (I have added sleep and selected candidate does not change), though UDP pair priority is higher.

gather.go
ashellunts: I prefer more verbose naming. What about connectionConfiguration?

candidate_base.go
```diff
@@ -355,7 +355,7 @@ func (c *candidateBase) Priority() uint32 {
 	// candidates for a particular component for a particular data stream
 	// that have the same type, the local preference MUST be unique for each
 	// one.
-	return (1<<24)*uint32(c.Type().Preference()) +
+	return (1<<24)*uint32(c.Type().Preference(c.networkType)) +
 		(1<<8)*uint32(c.LocalPreference()) +
```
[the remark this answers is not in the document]
ashellunts: See implementation of LocalPreference(), it already conforms to RFC6544.

agent_config.go
```diff
@@ -46,6 +46,9 @@ const (
 
 	// maxBindingRequestTimeout is the wait time before binding requests can be deleted
 	maxBindingRequestTimeout = 4000 * time.Millisecond
+
+	// tcpReadBufferSize is the size of the read buffer of tcpPacketConn used by active tcp candidate
+	tcpReadBufferSize = 8
```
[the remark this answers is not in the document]
ashellunts: Note, this is not in bytes, but size of data channel of tcpPacketConn where "packets" are placed (buffer that is read from network in one call).
I am not sure what is good default here. Probably we should have it in user configuration.

ashellunts:
Looks like it is expected behavior.

If I add timeout for host candidate, UDP is always selected.

candidatetype.go
ashellunts: That would mean we need to change existing default type priority for UDP.
Because currently host UDP type priority is 126, max value. So we can't increase it anymore.
If we want to use offsets, we need to make lower.

That may affect existing applications. Personally I would not want to affect existing library clients if they don't use TCP at all.

ashellunts: See also https://github.com/pion/ice/commit/92be181f41dde956502806bc283aba289dccca7a

ashellunts: Though we may use offset to reduce TCP priorities and not change UDP ones. That would conform to the standard.

Sean-Der (approved):
Fantastic work! I am all for merging this @ashellunts 

We should then cut a release and reach out to the user that was looking to use Active TCP. Tell me if there is anything I can do to help :)

Sean-Der:
@ashellunts mind just squashing into one commit? Just to make debugging easier!

really excited to see this long standing issue resolved:)

agent_config.go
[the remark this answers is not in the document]
stv0g: Thanks, I see. Lets try how it works out. If it becomes an issue, we can still increase it or make it user-configurable.

The test uses the value 2048: https://github.com/pion/ice/blob/72773dfdd1507d0062fede528fbedc8b9f055320/candidate_test.go#L359

stv0g (approved):

Issue #532, opened 2023-02-09
 TestAcceptAggressiveNomination failed in CI


ashellunts:
https://github.com/pion/ice/actions/runs/4134589539/jobs/7145911149

```
❌ TestAcceptAggressiveNomination (1.41s)
      agent_test.go:[17](https://github.com/pion/ice/actions/runs/4134589539/jobs/7145911149#step:7:18)80: 
          	Error Trace:	/home/runner/work/ice/ice/agent_test.go:1780
          	Error:      	Should be true
          	Test:       	TestAcceptAggressiveNomination
```

stv0g:
This seems to be a flaky test. It only fails sporadically.

ashellunts:
Another failure:
https://github.com/pion/ice/actions/runs/4908267974/jobs/8763923028?pr=564

PR #570, opened 2023-05-17
Test refactoring

stv0g:
This is also a shot at fixing #532


agent_handlers_test.go
```diff
@@ -39,14 +39,10 @@ func TestOnConnectionStateChangeCallback(t *testing.T) {
 	}
 
 	aAgent, err := NewAgent(cfg)
-	if err != nil {
-		t.Error(err)
-	}
+	assert.NoError(err)
```
ashellunts: I think we shall not use assert except we really need it.
Require is better because test stops when error happens.
With assert test continues, and most likely we don't want that.

ashellunts: I see, you already replaced to "require".

ashellunts (approved):
Wonderful!

agent_handlers_test.go
stv0g: Yes, I started by refactoring it while maintaining its original behaviour (`t.Error()` => `assert.XXX`).
But you are correct.. I think we often should really use `require` instead.

Sean-Der:
cherry picked assert -> require 05ab6847411265463a1852a0de4d597f1d9d99bc

Sean-Der:
Cherry picked CheckRoutines simplification in 01c35354b07b10903f3dfac64c883356958241a7

Sean-Der:
Cherry picked test.TimeOut change with 85a3a7f52407586fea6957187c461aaa08241611

Sean-Der:
Cherry-picked `t.Fatal` -> `require.NoError` with 52f2075c2e91a4912adcf4a7a13cdf019fbd5bd6

Issue #581, opened 2023-05-23
Active TCP performs network operations in main thread


davidzhao:
### Your environment.
  - Version: 2.3.4
  - Client: any client that generates passive TCP candidates (iOS, Go, other libwebrtc clients)

### What did you do?
Reproducible with all iOS devices using WebRTC M103+. This is because iOS is generating passive TCP candidates that triggers active TCP.

The issue appears to be due to how connection to active TCP candidate pair are implemented. When a new TCP candidate is discovered, it's attempting to dial to it in a blocking manner [here](https://github.com/pion/ice/blob/master/agent.go#L598).

Because this is blocking the Agent's runloop, it can cause other candidates to fail. This is particularly noticeable when a remote candidate is added with an address behind NAT. That connection would never resolve.

### What did you expect?

the presence of Active TCP candidates do not cause other ICE candidates to fail.

davidzhao:
<img width="1261" alt="Screenshot 2023-05-23 at 2 42 29 PM" src="https://github.com/pion/ice/assets/320582/409b7ae1-8d98-488a-bb8a-63fa5291adf4">

Sean-Der:
@ashellunts I was able to reproduce this with the following test

```
// Assert that Active TCP connectivity isn't established inside
// the main thread of the Agent
func TestAgentActiveTCP_NonBlocking(t *testing.T) {
        report := test.CheckRoutines(t)
        defer report()

        lim := test.TimeOut(time.Second * 5)
        defer lim.Stop()

        cfg := &AgentConfig{
                NetworkTypes: supportedNetworkTypes(),
        }

        aAgent, err := NewAgent(cfg)
        if err != nil {
                t.Error(err)
        }

        bAgent, err := NewAgent(cfg)
        if err != nil {
                t.Error(err)
        }

        isConnected := make(chan interface{})
        err = aAgent.OnConnectionStateChange(func(c ConnectionState) {
                if c == ConnectionStateConnected {
                        close(isConnected)
                }
        })
        if err != nil {
                t.Error(err)
        }

        // Add a invalid ice-tcp candidate to each
        invalidCandidate, err := UnmarshalCandidate("1052353102 1 tcp 1675624447 192.0.2.1 8080 typ host tcptype passive")
        if err != nil {
                t.Fatal(err)
        }
        aAgent.AddRemoteCandidate(invalidCandidate)
        bAgent.AddRemoteCandidate(invalidCandidate)

        connect(aAgent, bAgent)

        <-isConnected
[...]

PR #582, opened 2023-05-23
Introduce flag to opt-into active TCP

davidzhao:
Temporary workaround for #581.


Sean-Der:
Really sorry about the breakage @davidzhao 

I reverted. When this lands again I will make sure this can be disabled AND has zero impact if enabled.

davidzhao:
Thanks Sean! Really appreciate all your efforts here.

PR #586, opened 2023-06-15
Implement Active ICE TCP Candidates

Sean-Der:
This is undoing my revert in abc1ca33b5557f51f94152d0b2442f039a9256b6

I have also
* Added a test to make sure we don't block the main thread
* Added a config option so it can be disabled

I would like to have it on by default to benefit users as much as possible. I would prefer if pion/ice connected in as many networks as possible. It can be turned off if users wish to conserve resources though.


davidzhao (approved):
lgtm!

PR #596, opened 2023-07-03
Add `GetRemoteCandidates()`

stv0g:
This is analogue to `GetLocalCandidates()` and provides the user with the ability to get a slice of previously added remote candidates without keeping track of them in the application code.

### DIFF
--- a/agent_test.go
+++ b/agent_test.go
@@ -1322,6 +1322,74 @@ func TestGetRemoteCredentials(t *testing.T) {
 	assert.NoError(t, a.Close())
 }
 
+func TestGetRemoteCandidates(t *testing.T) {
+	var config AgentConfig
+
+	a, err := NewAgent(&config)
+	if err != nil {
+		t.Fatalf("Error constructing ice.Agent: %v", err)
+	}
+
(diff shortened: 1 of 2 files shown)


at-wat (commented):
The code looks good!
It would be better to add a test if it's not hard.

stv0g:
Hi @at-wat,

sure. I also added one for `GetLocalCandidates()`.

agent_test.go
```diff
@@ -1322,6 +1322,74 @@ func TestGetRemoteCredentials(t *testing.T) {
...
+			Port:      1000 + i,
+			Component: 1,
+		}
+
+		cand, errCand := NewCandidateHost(&cfg)
+		assert.NoError(t, errCand)
+
+		expectedCandidates = append(expectedCandidates, cand)
+
+		err = a.addCandidate(context.Background(), cand, dummyConn)
```
at-wat, line 1382: Lint error

stv0g: Thanks :) Its fixed in the latest version

at-wat (approved):
LGTM

PR #598, opened 2023-07-14
Harmonize errors and error logging

stv0g:


at-wat (commented):
In my understanding, `newActiveTCPConn` is called for each candidate and these warn logs will be shown for every interface addresses?
If it's correct, I think these logs should be info level.

stv0g:
Hi @at-wat,

thanks, you are right. I've changed it

at-wat (approved):
LGTM

PR #607, opened 2023-08-24
No ice

lisay-yan:
#### Reference issue
Fixes #...

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -392,8 +392,11 @@ func (a *Agent) startConnectivityChecks(isControlling bool, remoteUfrag, remoteP
 		return ErrMultipleStart
 	default:
 	}
-	if err := a.SetRemoteCredentials(remoteUfrag, remotePwd); err != nil { //nolint:contextcheck
-		return err
+
+	if !a.iceNone {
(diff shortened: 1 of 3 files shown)


stv0g:
Can you please add a description and motivation of this PR?

Sean-Der:
Sorry this didn't get a response sooner. I spent a lot of time prototyping/trying different things here.

I think if someone doesn't want ICE using ORTC is a better bet. Like [this](https://github.com/pion/webrtc/tree/master/examples/ortc-media)

I have added a few hooks like [BindingRequestHandler](https://github.com/pion/ice/commit/07b74c8a970fae136e3abd4074a74657edfba82e) to pion/ice and that seems to work well. I tried to make the `SelectionStrategy` an interface by the diff was really large. 

I then anticipated what if someone likes WebRTC, but wants to use QUIC for DataChannels? What if someone likes WebRTC but wants to use something besides DTLS etc... the answer to that was DTLS.

I am sorry I didn't share this conclusion sooner. I feel bad closing PRs, so I kept trying different ways to make it work. I should have just come to that conclusion sooner :/

If you want to get on a call https://siobud.com/meeting I would love to help port things to ORTC and do anything I can to help.

Issue #571, opened 2023-05-17
Fix type of CandidatePairState enum


stv0g:
The constants are currently typed as `int`. We should use the specific `CandidatePairState` type instead.

PR #608, opened 2023-08-29
Improve performance of UDPMux: Add BatchIO and multiple ports options

### LINKED ISSUE #128: reduce `candidateBase.writeTo` cpu cost
## Summary

image

## Motivation
reduce `candidateBase.writeTo` cpu cost

Maybe we can use bufio  in `candidateBase.writeTo`, like:
https://github.com/grpc/grpc-go/pull/1544/commits/30d8f0749874c54017578cf0760e18c01d27a3a2

cnderrauber:
Performance improvements:
Add BatchIO option to NewMultiUDPMuxPort(s), so the UDPMux can use batch write to improve write throughput.
Add multiple ports support to MultiUDPMuxDefault, this option will listen to multiple ports on a single interface for traffic load balance.


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/pion/mdns v0.0.7
 	github.com/pion/randutil v0.1.0
 	github.com/pion/stun v0.6.1
-	github.com/pion/transport/v2 v2.2.1
+	github.com/pion/transport/v2 v2.2.2-0.20230829043045-6a34769ff4b0
```
cnderrauber, line 13: Will change it once the related transport PR is merged.

boks1971 (approved):
LGTM! Great work @cnderrauber !!!

udp_mux_multi.go
```diff
@@ -18,20 +20,93 @@ import (
...
+	}
+	return leastAddr
+}
+
+// MultiUDPMuxOption provide options for NewMultiUDPMuxDefault
+type MultiUDPMuxOption func(*multipleUDPMuxDefaultParams)
+
+// MultiUDPMuxOptionWithPortBalance enable traffic balance on ports that belongs to same IP address,
+// that means the MultiUDPMuxDefault will return a single port that has least connection for each IP address
+// in GetListenAddresses return.
```
davidzhao, line 58:
```suggestion
// MultiUDPMuxOptionWithPortBalance enables load balancing traffic on multiple ports belonging to the same IP
// When enabled, GetListenAddresses will return the port with the lease number of connections for each corresponding IP
```

udp_mux_multi.go
```diff
@@ -76,6 +170,12 @@ func (m *MultiUDPMuxDefault) GetListenAddresses() []net.Addr {
 // NewMultiUDPMuxFromPort creates an instance of MultiUDPMuxDefault that
 // listen all interfaces on the provided port.
 func NewMultiUDPMuxFromPort(port int, opts ...UDPMuxFromPortOption) (*MultiUDPMuxDefault, error) {
+	return NewMultiUDPMuxFromPorts([]int{port}, opts...)
+}
+
+// NewMultiUDPMuxFromPorts creates an instance of MultiUDPMuxDefault that
+// listen all interfaces and balance traffic on the provided ports.
```
davidzhao, line 177:
```suggestion
// listens to all interfaces and balances traffic on the provided ports. 
```

udp_mux_multi.go
```diff
@@ -64,8 +148,18 @@ func (m *MultiUDPMuxDefault) Close() error {
 	return err
 }
 
-// GetListenAddresses returns the list of addresses that this mux is listening on
+// GetListenAddresses returns the list of addresses that this mux is listening on,
+// if port balance enabled and there are multiple mux listen on different ports of a same IP addr,
+// will return the mux who has least connections of that IP addr.
```
davidzhao, line 153:
```suggestion
// GetListenAddresses returns the list of addresses that this mux is listening on,
// if port balance enabled and there are multiple muxes listening to different ports of the same IP addr,
// it will return the mux that has the least number of connections.
```

davidzhao (approved):
Really cool to see this all come together. Nicely done!

just a few suggestions on comments

udp_mux_multi.go
```diff
@@ -53,9 +53,8 @@ func (mpa *multiPortsAddress) next() net.Addr {
 // MultiUDPMuxOption provide options for NewMultiUDPMuxDefault
 type MultiUDPMuxOption func(*multipleUDPMuxDefaultParams)
 
-// MultiUDPMuxOptionWithPortBalance enable traffic balance on ports that belongs to same IP address,
-// that means the MultiUDPMuxDefault will return a single port that has least connection for each IP address
-// in GetListenAddresses return.
+// MultiUDPMuxOptionWithPortBalance enables load balancing traffic on multiple ports belonging to the same IP
+// When enabled, GetListenAddresses will return the port with the lease number of connections for each corresponding IP
```
davidzhao, line 57: Sorry, I had a typo
```suggestion
// When enabled, GetListenAddresses will return the port with the least number of connections for each corresponding IP
```

udp_mux.go
```diff
@@ -23,6 +23,7 @@ type UDPMux interface {
 	GetConn(ufrag string, addr net.Addr) (net.PacketConn, error)
 	RemoveConnByUfrag(ufrag string)
 	GetListenAddresses() []net.Addr
+	ConnCount() int
```
stv0g, line 26: Unfortunately, this breaks the API :(

We need to bump the major version then.

stv0g (changes requested):
Hi @cnderrauber,

great to see progress in this direction :)

A few points:

1. This PR breaks the API of `UDPMux`. So we need to bump the major version of the module. Thats something I am open to, but I think we should batch this with other API-breaking changes (like #571).
2. Cant you simply use `AllConnsGetter` to get the count of connections?
3. I do not the the batched connection interface being used anywhere yet. Do you have or are planning to open a dedicated PR for this?
4. It would be good if we can also add those new features to the standard `UDPMux`. Or is there a reasoning behind adding this only to the `MultiUDPMux`?

Best regards,
Steffen

cnderrauber:
> 2. Cant you simply use `AllConnsGetter` to get the count of connections?

I think it can't be used. The `AllConnsGetter` is implented by tcpmux and it also doesn't have a method to get the count of connections.

> 3. I do not the the batched connection interface being used anywhere yet. Do you have or are planning to open a dedicated PR for this?

I can't get the point of this question, the batched connection interface is in the pion/transport repo, what do you mean a dedicated PR?

> 4. It would be good if we can also add those new features to the standard `UDPMux`. Or is there a reasoning behind adding this only to the `MultiUDPMux`?

The UDPMux can have the batch feature simply by pass a PacketConn to its constructor `NewUDPMuxDefault`. And it should only listen on a single port so no need the ports balance option.

udp_mux.go
cnderrauber: That's a question... though I think there is no developer has implemented his own UDPMux but keep API compatibility will be good. I will try to import a new interface to the port balance feature and only enable it if the child Muxes support the new API to keep the `UDPMux` unchanged.

stv0g:
> I can't get the point of this question, the batched connection interface is in the pion/transport repo, what do you mean a dedicated PR?

Sorry. I dont see where the batched connection interface is used in pion/ice. I assume we need another PR to actually use it? Or did I missed something?

> The UDPMux can have the batch feature simply by pass a PacketConn to its constructor NewUDPMuxDefault. And it should only listen on a single port so no need the ports balance option.

Alright. Thanks for the explanation.

udp_mux.go
davidzhao: good catch @stv0g! It'd be good to keep UDPMux interface simpler without this additional requirement.

udp_mux_multi.go
```diff
@@ -95,20 +194,29 @@ func NewMultiUDPMuxFromPort(port int, opts ...UDPMuxFromPortOption) (*MultiUDPMu
...
+				break
+			}
+			if params.readBufferSize > 0 {
+				_ = conn.SetReadBuffer(params.readBufferSize)
+			}
+			if params.writeBufferSize > 0 {
+				_ = conn.SetWriteBuffer(params.writeBufferSize)
+			}
+			if params.batchWriteSize > 0 {
+				conns = append(conns, tudp.NewBatchConn(conn, params.batchWriteSize, params.batchWriteInterval))
```
cnderrauber, line 212: @stv0g The BatchConn interface has been used here which is implemented in the pion/transport.

udp_mux_multi_test.go
```diff
@@ -110,46 +111,107 @@ func testMultiUDPMuxConnections(t *testing.T, udpMuxMulti *MultiUDPMuxDefault, u
...
+				testMultiUDPMuxConnections(t, udpMuxMulti, "ufrag1", udp)
+			}()
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				testMultiUDPMuxConnections(t, udpMuxMulti, "ufrag2", udp4)
+			}()
+
+			// Skip IPv6 test on i386
+			const ptrSize = 32 << (^uintptr(0) >> 63)
```
stv0g, line 167: Maybe we should check against `runtime.GOARCH` here?

udp_muxed_conn.go
```diff
@@ -80,10 +89,18 @@ func (c *udpMuxedConn) WriteTo(buf []byte, rAddr net.Addr) (n int, err error) {
 	if c.isClosed() {
 		return 0, io.ErrClosedPipe
 	}
-	// Each time we write to a new address, we'll register it with the mux
-	addr := rAddr.String()
-	if !c.containsAddress(addr) {
-		c.addAddress(addr)
+
+	// Only checking the address at the ice connecting stage to reduce the check cost
```
stv0g, line 93:
```suggestion
	// Only check the address at the ICE connecting stage to reduce the check cost
```

udp_mux.go
```diff
@@ -176,6 +181,13 @@ func (m *UDPMuxDefault) GetConn(ufrag string, addr net.Addr) (net.PacketConn, er
 	return c, nil
 }
 
+// ConnCount return count of working connections created by UDPMuxDefault
```
stv0g, line 184:
```suggestion
// ConnCount returns count of working connections created by UDPMuxDefault
```

udp_mux_multi.go
```diff
@@ -4,35 +4,130 @@
...
 	"net"
+	"time"
 
 	"github.com/pion/logging"
 	"github.com/pion/transport/v2"
 	"github.com/pion/transport/v2/stdnet"
+	tudp "github.com/pion/transport/v2/udp"
 )
 
+var errPortBalanceRequireConnCount = errors.New("Port balance requires UDPMux implements MuxConnCount interface")
```
stv0g, line 18:
```suggestion
var errPortBalanceRequireConnCount = errors.New("port balancing requires a UDPMux which implements MuxConnCount interface")
```

udp_mux_multi.go
```diff
@@ -4,35 +4,130 @@
...
+	// avoid duplicate ip candidates be gathered for a single ice agent.
+	multiPortsAddresses []*multiPortsAddress
+}
+
+type addrMux struct {
+	addr net.Addr
+	mux  MuxConnCount
+}
+
+// each multiPortsAddress represents muxes listen on different ports of a same IP
```
stv0g, line 39:
```suggestion
// Each multiPortsAddress represents muxes listen on different ports of a same IP
```

udp_mux_multi.go
stv0g: I mean, I dont find any code in pion/ice which calls `WriteBatch` or `ReadBatch`.
Maybe I am overseeing something, but having a batched UDPMux conn only helps if its also used by `ice.Conn`.

See: https://github.com/pion/ice/blob/d41eb5aae74782d5abb117efb52187f8caa06710/transport.go#L70

stv0g (changes requested):

udp_mux_multi.go
cnderrauber: Why ice need to call WriteBatch or ReadBatch? It has been wrapped in transport/udp.BatchConn

udp_mux_multi_test.go
cnderrauber: I think they are identical, and if one day there is a new GOARCH, the test will be failed.

udp_mux_multi.go
davidzhao: @stv0g pion/transport uses batching internally. `WriteTo` would enqueue packets for batching [here](https://github.com/pion/transport/blob/master/udp/batchconn.go#L109)

stv0g:
Hi @cnderrauber,

(This is a reply to https://github.com/pion/ice/pull/608#discussion_r1312620889, but I think its a more general concern which deserves its own comment.)

I am sorry to raise objections about this PR. But I am afraid that this PR introduces even more complexity into the UDPMux logic. And given the current state, we already have three different types for UDP muxing and we have not yet achieved some of the initial goals to allow single-port muxing for relayed candidates.

I was under the impression that the PR attempt to introduce a batched connection interface to `ice.Conn` and allow passing batches of packets all the way to the mux.

I am not really that happy with hiding the packet batching/queuing inside the mux.
This mixing concerns here. The mux should mux packets over a single port.
It should not doing port balancing and queuing in my opinion.

I cleaner implementation would expose a batched connection interface to the user of pion/ice (in the form of `ice.Conn`) so that you can implement the queuing in your application logic.

Because of the implementation inside the UDP muxing logic, the port balancing and queuing is also limited to configurations which use the mux. How about configuration without a UDPMux? Or relayed candidates?

cnderrauber:
>And given the current state, we already have three different types for UDP muxing and we have not yet achieved some of the initial goals to allow single-port muxing for relayed candidates.

That's true we already have three different types of UDPMux but actually they are all based on the UDPMuxDefault so they will all get benefits from the change. And I can't see why this will blocked other goals like `relayed candidates`.

> I am not really that happy with hiding the packet batching/queuing inside the mux. This mixing concerns here. The mux should mux packets over a single port. It should not doing port balancing and queuing in my opinion.
> I cleaner implementation would expose a batched connection interface to the user of pion/ice (in the form of `ice.Conn`) so that you can implement the queuing in your application logic.

Port balancing and Batch are two standalone options to improve the mux throughput. A Mux only listens on a single port without batch will have a poor performance that can't be used in a large-scale production environment.  
I don't think ice.Conn need the detail of batch as the batch only makes sense to the udp mux. Because in normal non-mux sessions, every single port only has limited traffic and packet rates that can't queue enough packet to write them in batch, otherwise, remote peer will get a very high jitter by the queueing.

> 
> Because of the implementation inside the UDP muxing logic, the port balancing and queuing is also limited to configurations which use the mux. How about configuration without a UDPMux? Or relayed candidates?

So port balance and batch should limit to the mux.

davidzhao:
> I am sorry to raise objections about this PR. But I am afraid that this PR introduces even more complexity into the UDPMux logic. And given the current state, we already have three different types for UDP muxing and we have not yet achieved some of the initial goals to allow single-port muxing for relayed candidates.
>
> I was under the impression that the PR attempt to introduce a batched connection interface to `ice.Conn` and allow passing batches of packets all the way to the mux.

Don't be sorry! Appreciate you bringing up these points. It's worth discussing the tradeoffs that are being made. Let me share a bit more color here on the intent.

The goal of the PR is to improve performance of Pion when writing out to clients. In benchmarking, we discovered the bottleneck to be syscalls made during the `writeTo` path. In a typical case, the Pion-based server is sending packets simultaneously to multiple clients, so @cnderrauber came up with the proposal of batching packets intended for different clients together in the same syscall.

Benchmarks have shown that it's improved write performance by over 100%, compared to using standard pion/ice connections without batching. Also, by batching packets to multiple clients together, it helps to keep jitter under control, versus queuing sufficient packets before calling `WriteBatch` explicitly to each destination.

On the point of why it would use multiple ports. We've discovered additional bottlenecks writing out using a single UDP port. Locks within the write path would limit the number of cores we can effectively utilize. By allowing multiple ports to be used, we were able to get around some of these bottlenecks. Now, this is an option that isn't enabled by default. And for those that don't care about squeezing these bits of performance out of the system, the standard UDPMux is still available.

I hear you on the complexity it introduces in the code, though I hope what's in this PR is a reasonable set of tradeoffs that doesn't make the interface anymore difficult to use.

We are opening the PR here because we thought the work could benefit everyone using Pion. But we'd also respect your feedback if you feel that these improvements should be kept in our own repo instead.

stv0g:
Hi @davidzhao,

Thanks for the detailed description of the motivation behind this PR. I fully agree with you on these points. Batching can really help here to increase the throughput. I would also love to profit from this feature in my own app.

There is also a very good article by Tailscale where they describe how they have been able to accelerate the WireGuard Go implementation to achieve over 10Gbps throughput in Userspace by using batching: https://tailscale.com/blog/more-throughput/

I am only arguing about the API which we use to realize it. I would love to have a batching API also for `ice.Conn`. But it seems like you mainly profit by the fact the you accumulate all your connections behind the mux. The just passing through batches from `ice.Conn` to the batched mux connection would not really help?

Have you thought about abandoning the use of muxes? Using dedicated connections per ICE agent could help you with the load balancing as you are dealing with a separate per connection?

Regarding the port balancing: Have you considered using `SO_REUSE_PORT`?
See:
- https://lwn.net/Articles/542629/
- https://www.kernel.org/doc/Documentation/networking/scaling.txt

udp_mux_multi.go
stv0g: Thanks. Understood :)

streamer45:
Apologies for the intrusion but port balancing on the same address/port through `unix.SO_REUSEADDR` is pretty much how we improved write throughput on our side as well. We implemented a `multiConn` wrapper implementing `net.PacketConn` that does some round robin on all available connections (e.g. one per CPU core). Batching sounds like a great option to add on top of that, looking forward to it.

kcaffrey:
We are also load balancing on a single port using `SO_REUSEADDR`, probably in a very similar way that @streamer45 is. The wrapper still implements `net.PacketConn`, so we can plug it into the existing `UDPMux` without modifying `UDPMux` itself.

Having batching as well would be nice, but I echo the earlier comments regarding the API. It would be nice if the batching wasn't coupled so tightly to the "multi udp mux" variant, but rather something that played nicely with the default UDP mux too.

davidzhao:
@kcaffrey @streamer45 any interest in adding port balancing into the standard UDPMux? It should be straight-forward to integrate with batching at that point.

It seems that the primary objection to the PR is the implication that batching will only be available to UDPMuxMulti and nothing else. That's not the case here. Most of the work had been put into `transport/udp` and can be leveraged by any of the muxes. To me it seems completely reasonable to have UDPMuxMulti to be the first consumer of this capability.

Am I misunderstanding the concerns?

AshishKumar4:
Hello! Just checking on the progress of this PR. We our building a stack (consisting of an SFU) based on Pion, and have encountered similar performance bottlenecks on the writeTo pathway and were actually considering forking and implementing batching ourselves before stumbling on this PR. We would really appreciate to know about the progress here as indeed we expect huge performance gains from this. In the meanwhile, I would try to understand the codebase better as we have only majorly played with the exposed APIs. Btw thanks for implementing this PR in the first place, and the awesome pion project!

davidzhao:
@AshishKumar4 the PR was ready to merge months ago, but several folks took issues with certain design decisions. We could merge as is, but in trying to be good citizens, we are holding off until @stv0g or others give the 👍. Until then, I'm afraid this PR is stuck here.

kcaffrey:
I have no major objections to merging as is. My earlier note was primarily intended to provide information on how we are using the API currently (that is, passing in a socket we create ourselves so that we can perform load balancing over a single port).

It looks like it would be possible to get batching (even without this PR) for any user who is passing a packet conn to UDPMux (rather than having one of the factory methods create and listen on the socket). I think it will be good to maintain this for any future improvements. For example, the batching strategy used by the implementation in pion/transport is not quite what we would want as it adds latency and jitter, so it is important to us that we can continue plugging in our own implementation of `net.PacketConn` (we would be fine implementing additional interfaces on the connection if needed).

kcaffrey:
It makes sense to me to make it easy to enable batching from `pion/webrtc` via a setting engine option, but I don't think it should be the default.

The [current implementation](https://github.com/pion/transport/blob/master/udp/batchconn.go#L76-L87) in pion/transport batches on a fixed interval, which adds half the batch interval on average to the latency. Even worse (for many use cases), it will increase the apparent jitter by an amount proportional to the batch interval (assuming that packet write attempts arrive randomly within the interval), which will in turn cause client-side jitter buffers to grow to compensate.

Back-pressure-triggered batching, where writes are queued but attempted immediately, and only batched when there are multiple queued writes available when starting a new syscall, would be a much safer default in my opinion. CPU savings will be lower or even negligible (especially if throughput is not bottlenecked), but overall throughput would be increased without adding any additional latency or jitter.

If we were to have any on-by-default batching, I think it should be the opportunistic/"only batch when blocked" style described above. There are certainly applications where the fixed-interval-batching is a better tradeoff (where end to end latency isn't that important but overall CPU usage IS important), but I imagine that would not be true for a majority of use cases.

cnderrauber:
> @cnderrauber @boks1971 do you guys still wanna merge this, it's been years, and I think the design issues can be addressed in later refactors or versions. I can fix the conflicts.

Will be good to have this in pion, it is useful in high throughput case that other pion users could get benefit from it.

Sean-Der:
Sorry we didn't make more progress on this!

I am going to read all the Mux implementations right now and figure out the right way to do this I want to support.

* ECN
* Listening on multiple ports
* Batched Writes
* Caching STUN Gathering
* ?

Lots of features we use and I want to understand it all before proposing a design.

PR #612, opened 2023-09-02
Clear selected pair on ICE failed

boks1971:
When ICE times out and fails, all candidates are deleted.

That means all the candidates are closed and their underlying conns are clsoed.

But, the selected pair could still be valid. On a subsequenct `Write`, ICE transport conn will get the selected pair and write to the pair. As the pair is still valid, write will flow through to the local candidate `writeTo`.

But, as all candidates and their underlying conns are closed, `Write` will return a `io.ErrClosedPipe` error.

There are cases where it is not ignored and causes a broken pipe after an ICERestart.

When the `Write` error propagates back to sctp/association, the writeLoop is exited.

So, sending data channel traffic after a successful ICERestart still fails as the SCTP association errored out and write loop exited.

I have copied the changes that are done when ICERestart happens to when ICE state is set to failed (except for gathering state and resetting ufrag/pwd). In my testing, it is working well, i. e. can continue data channel after ICE Restart whereas previously it was failing every time. But, I am not sure of all the implications of this change.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -15,6 +15,7 @@ Artur Shellunts <shellunts.artur@gmail.com>
 Assad Obaid <assad@lap5cg901003r.se.axis.com>
 Atsushi Watanabe <atsushi.w@ieee.org>
 backkem <mail@backkem.me>
+boks1971 <raja.gobi@tutanota.com>
 buptczq <czq@qq.com>
 cgojin <gongjin21@hotmail.com>
 Chao Yuan <yuanchao0310@163.com>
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
This looks good to me!

I don't see any issues with locking interactions. For the second concern, I think the expectation is that a restart is necessary after reaching failed state. so when ICE Restart is performed, the same logic will be executed again.

Agreed on the last one.. I guess that can be addressed separately since that race is not introduced by this PR.

boks1971:
Merging this based on @davidzhao 's review + approval. If there is further feedback, I will address it in a subsequent PR.

PR #625, opened 2023-10-10
Respect IncludeLoopback in UDPMuxDefault

dinvlad:
#### Description

Currently, when using `UDPMuxDefault` with unspecified address, the loopback address is [included](https://github.com/pion/ice/blob/5e6a7d1b3b9be2908def8e21b6858c5bbf81a179/udp_mux.go#L96) by default, but `agentConfig.IncludeLoopback` is not respected when gathering local candidates.

The same holds true when `UDPMuxDefault` is configured with a loopback address, but `agentConfig.IncludeLoopback` is not explicitly set to `true`.

This commit adds an extra check to `gatherCandidatesLocalUDPMux()` for respecting that setting in both cases.


Sean-Der:
Cherry-picked with 77cc354d7ff64da809e1b2b36b27ff443f7415ef

PR #629, opened 2023-11-24
Prioritize TURN servers according to they relay protocol

karaziox:
#### Description

Today, all relay candidates from Pion have the same priority. This PR attempts to reproduce libwebrtc's behavior, where the TURN servers candidates priority is based on the underlying relay protocol. UDP are preferred over TCP, which are preferred over the TLS options. 

We reuse the same factors as libwebrtc:
https://github.com/mozilla/libwebrtc/blob/1389c76d9c79839a2ca069df1db48aa3f2e6a1ac/p2p/base/turn_port.cc#L61

### DIFF
--- a/candidate_relay.go
+++ b/candidate_relay.go
@@ -70,6 +70,23 @@ func NewCandidateRelay(config *CandidateRelayConfig) (*CandidateRelay, error) {
 	}, nil
 }
 
+// LocalPreference returns the local preference for this candidate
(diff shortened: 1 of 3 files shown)


Sean-Der:
Thank you so much @karaziox I am merging this now!

This is a pretty deep change! Anything else with Pion you would like to see improved? What else do you think could be better? Would love to hear what else you are looking at :)

Sean-Der (approved):

PR #631, opened 2023-12-01
Allow generating candidates for both local and external addresses

streamer45:
#### Description

PR makes some changes to `externalIPMapper`  in order to allow us to generate an additional candidate for a local IP in case the same address is already mapped to an external one.

Before this PR, a mapping such as `[]string{"EA/IA", "IA/IA"}` would throw an error since the internal address appears twice. With this PR, that same mapping should be allowed and generate two candidates, one with the external address and one with the local address.

To do this we are converting the `findExternalIP` methods to `findExternalIPs` and return a slice which we iterate on. I tried to avoid changing the logic too much so the changes should be backwards compatible. That means that we are still allowing only one external IP for a given local IP.

That said, I am not very experienced with this logic so I'd like to discuss with people more familiar with the code base whether the proposed changes make sense and if there could be edge cases in which this addition may not work as desired. 

The original discussion along with a practical example can be found at https://gophers.slack.com/archives/CAK2124AG/p1701379410647329

### DIFF
--- a/external_ip_mapper.go
+++ b/external_ip_mapper.go
@@ -18,9 +18,13 @@ func validateIPString(ipStr string) (net.IP, bool, error) {
 
 // ipMapping holds the mapping of local and external IP address for a particular IP family
 type ipMapping struct {
-	ipSole net.IP            // When non-nil, this is the sole external IP for one local IP assumed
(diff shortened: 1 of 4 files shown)


streamer45:
@cnderrauber @stv0g Not urgent but would love your feedback if you have a chance to look at this. Thank you!

/cc @Sean-Der

PR #634, opened 2023-12-22
Allow using mDNS candidates with UDP multiplexing

Aleksbgbg:
#### Description

The UDP mux code path for gathering candidates is different from the main path in that it doesn't take the mDNS setting into account. This pull request fixes the issue.

In the future, we should look into sharing the code as much as possible between the code paths to avoid having to duplicate the logic (and therefore inadvertently making mistakes).

### DIFF
--- a/net.go
+++ b/net.go
@@ -106,7 +106,7 @@ func listenUDPInPortRange(n transport.Net, log logging.LeveledLogger, portMax, p
 	var i, j int
 	i = portMin
 	if i == 0 {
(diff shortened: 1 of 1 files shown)


Sean-Der:
Thank you so much @Aleksbgbg I cherry-picked with a72844663f279caa422e97ae760509d8916bc708

Do you have any interest in fixing the duplicate logic? I would love to see things improved :) I haven't had a chance to look at that recently though

PR #641, opened 2024-01-11
Trace log inbound messages

boks1971:
On an ICE restart in controlled mode, seeing incoming messages getting discarded to due to username mismatch. That is because the broswer is still using its old candidate and user name. As the controlled agent waits for `useCandidate` from the controlling agent, the controlled agent does not get to connected/nominated state inspite of getting several success responses. Suspect the controlling side does not have `useCandidate` for the new pair and it is still sending it for the old pair. Logging more details in trace to understand it better.


davidzhao (approved):

boks1971:
Unsure what is causing all the CI failures. Happens on a local run too. Did not expect my changes to produce that. Will dig.

boks1971:
oh, the tests are comparing string versions of the pairs.

Need to enhance tests or figure out a way to log differently.

PR #652, opened 2024-02-18
Avoid priv. ports

### LINKED ISSUE #651: Start the listening port by default above 1024
## Summary

Given of often users of ICE may run at non priveleged users - it may make sense to make the 'port min' default in https://github.com/pion/ice/blob/ab18a94710c2fd279759ade01f19fc1f0e775af1/net.go#L109 

from 1 to the unprivileged 1024

## Motivation

Easy of use

One can explicitly configure this - so pure nice to have.

dirkx:
#### Description

Avoid the priveleged port range when selecting a port to listen/bind to.

#### Reference issue
Fixes #651

### DIFF
--- a/net.go
+++ b/net.go
@@ -106,7 +106,7 @@ func listenUDPInPortRange(n transport.Net, log logging.LeveledLogger, portMax, p
 	var i, j int
 	i = portMin
 	if i == 0 {
-		i = 1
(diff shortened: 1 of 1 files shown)


Sean-Der:
Cherry-picked as 970978e8c79a2d68234ff1e85b655dbde0b090c9 thank you @dirkx 

Just had to fix a typo in the comment to pass CI

PR #653, opened 2024-02-22
Refine TCPMux memory usage

cnderrauber:
Reduce read buf of first stun message.
Add timeout to read first message to clean interrupted connection earlier.
Add alive duration for gather to access connection created from stun bind, avoid connection leak from the malicious client.


boks1971 (approved):
LGTM!!!!

Issue #624, opened 2023-09-28
ConnectionState changes may be reported incorrectly


sirzooro:
I already reported this in pion/webrtc but it is bug in pion/ice, so I am adding it here too for visibility. Original issue: https://github.com/pion/webrtc/issues/2578

Sometimes ConnectionState changes may be reported incorrectly - instead of getting Checking first and then Connected, state changes first to Connected, then Checking. This problem is caused by fact that Agent.connectionStateRoutine creates new goroutine to send new state (https://github.com/pion/ice/blob/master/agent_handlers.go#L52). Sometimes goroutine created to send Checking state is not executed immediately, but after another one used to send Connected state. I tried to remove go keyword there and looks that it fixed this problem. To avoid potential problems elsewhere, it would be good to add some buffer to chanState channel too.

Sean-Der:
Great find @sirzooro!

I think we need to keep the goroutine, but deliver the events in order. If we drop the goroutine the user can block our main thread.

Maybe we can internally have a buffered channel and have a goroutine deliver them to the user?

Sean-Der:
If you are interested in implementing this I would love to work with you on it!

sirzooro:
Yes, I can create PR for this.
`connectionStateRoutine` is started in separate goroutine, so adding buffer to `chanState` should be enough. Probably 10 entries would be enough.
BTW, `chanCandidate` and `chanCandidatePair` channels also do not have buffers - adding buffers of size 10 and 2 respectively probably would help them too.

Sean-Der:
great! Totally agree with that.

Could you write a test that blocks the callback with something long running? Just to confirm we aren't introducing a behavior change?

PR #656, opened 2024-03-13
Ensure sequential execution of connection state callbacks

sukunrt:
Ice update connection state callbacks need to happen sequentially otherwise a later callback could overwrite a newer callback. 

Fixes: #624

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -130,7 +130,7 @@ type Agent struct {
 
 	chanCandidate     chan Candidate
 	chanCandidatePair chan *CandidatePair
-	chanState         chan ConnectionState
+	stateNotifier     *connectionStateNotifier
 
 	loggerFactory logging.LoggerFactory
 	log           logging.LeveledLogger
@@ -227,7 +227,6 @@ func (a *Agent) taskLoop() {
 
 		after()
 
-		close(a.chanState)
 		close(a.chanCandidate)
 		close(a.chanCandidatePair)
 		close(a.taskLoopDone)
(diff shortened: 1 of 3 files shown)


agent_handlers.go
```diff
@@ -48,8 +50,14 @@ func (a *Agent) candidatePairRoutine() {
...
 func (a *Agent) connectionStateRoutine() {
+	var wg sync.WaitGroup
 	for s := range a.chanState {
-		go a.onConnectionStateChange(s)
+		wg.Add(1)
+		go func() {
+			a.onConnectionStateChange(s)
+			wg.Done()
+		}()
+		wg.Wait()
```
Jorropo, line 60: The whole sync goroutine does nothing, if this is correct just call the function.

agent.go
```diff
@@ -278,7 +278,7 @@ func NewAgent(config *AgentConfig) (*Agent, error) { //nolint:gocognit
 
 	a := &Agent{
 		chanTask:          make(chan task),
-		chanState:         make(chan ConnectionState),
+		chanState:         make(chan ConnectionState, 4),
```
Jorropo, line 281: Why 4 ?
Won't this deadlock if we get unlucky and there is 5 events queued up.

Jorropo (changes requested):

agent_handlers.go
sukunrt: Thats a good point. Let me check if there arent any locks held here and we can remove the go routine.

agent.go
sukunrt: We only need this to close the agent correctly. The current behaviour allows agent closing in the callback which with a channel capacity of 0 would lead to deadlock. 

I will see what the right way to handle this is.

agent_handlers.go
sukunrt: that didn't make a lot of sense :sweat_smile: 
Removed this. Thanks!

agent.go
sukunrt: Considering that we can go Connection to Disconnected to Connected many many times, I don't think a limit to the channel makes sense. Especially considering that we don't want to block the state updating go routine. So I've introduced a queue here which runs the callbacks in a separate goroutine

Sean-Der (approved):

PR #660, opened 2024-03-25
Support IPv6 from mDNS

edaniels:


transport_test.go
```diff
@@ -175,13 +176,20 @@ func gatherAndExchangeCandidates(aAgent, bAgent *Agent) {
 
 	candidates, err := aAgent.GetLocalCandidates()
 	check(err)
+
 	for _, c := range candidates {
+		if addr, err := netip.ParseAddr(c.Address()); err == nil {
+			if shouldFilterLocationTrackedIP(addr) {
```
edaniels, line 182: This is a nice check to make sure we never ever publish what are currently defined as location tracked IPs

Sean-Der:
@edaniels Would it possible to limit this just to mDNS for now?

I would love to see these other things fixed, but as unique efforts/PRs.

The `Mux` setup in ICE in particular needs evaluated. We have 3 different UDP Mux's I believe, and they all have different issues. Before we break any APIs I would like to understand them all and do one clean break!

edaniels:
> @edaniels Would it possible to limit this just to mDNS for now?
> 
> I would love to see these other things fixed, but as unique efforts/PRs.
> 
> The `Mux` setup in ICE in particular needs evaluated. We have 3 different UDP Mux's I believe, and they all have different issues. Before we break any APIs I would like to understand them all and do one clean break!

I want to say yes but let's see what breaks when I undo those changes in this PR. They must have happened in combination with a TCP mux plus mDNS tests. If they don't break, I'll keep it undone. If they do, I'll see if I can create a temporary hidden interface that is an experimental API that works for the mixed implemented in the pkg!

Sean-Der (approved):

PR #675, opened 2024-04-10
refactor: rename test_utils to utils_test to remove stretchr/testify dep

paralin:
I noticed that the pion-ice package had ~1MB of extra binary size due to the dependency on stretchr/testify.

Renaming test_utils.go to utils_test.go removes stretchr/testify from any non-test build and fixes this dependency. The tests still pass & the package still builds the same.

One can check the dependency graph with goda:

https://github.com/loov/goda

goda graph "reach(.:all, github.com/stretchr/testify/require)" | dot -Tsvg -o graph.svg

image


Sean-Der:
Nice catch! Do we have this problem anywhere else? 

I will add this check to CI

paralin:
After applying this PR I don't see any other reference to testify in my project's build which also imports the main webrtc packages. So, I think this is probably the only case where that happened.

Sean-Der (approved):

PR #676, opened 2024-04-13
Reuse connectivity check ticker

paulwe:
we're seeing high `runtime.siftdownTimer` cpu on some busy nodes and profiling indicates there are ~20k timers allocated here. recreating these on every loop is inefficient.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -8,6 +8,7 @@ package ice
 import (
 	"context"
 	"fmt"
+	"math"
 	"net"
 	"net/netip"
 	"strings"
@@ -372,6 +373,9 @@ func (a *Agent) connectivityChecks() {
 		}
 	}
 
+	t := time.NewTimer(math.MaxInt64)
+	t.Stop()
+
 	for {
 		interval := defaultKeepaliveInterval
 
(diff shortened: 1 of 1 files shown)


boks1971 (approved):
lgtm!

agent.go
```diff
@@ -392,7 +396,8 @@ func (a *Agent) connectivityChecks() {
 		updateInterval(a.disconnectedTimeout)
 		updateInterval(a.failedTimeout)
 
-		t := time.NewTimer(interval)
+		t.Reset(interval)
```
davidzhao, line 399: There is a path where the previous timer has not been reset properly. If `a.forceCandidateContact` was set.

we'd need to do this here before resetting

```
if !t.Stop() {
	<-t.C
}
```

boks1971: There is a `Stop` call as soon as the timer is created ☝️ . So, a `Reset` can be applied any time after that. There are a couple of `Stop` calls 👇 , but I guess they become no-ops. Did I misread it?

davidzhao: actually I've misread 🤦. the there is a Stop there as well. all good here :)

PR #677, opened 2024-04-16
Handle timer stop race in agent connectivity check

paulwe:


boks1971 (approved):
@davidzhao caught this in the last PR itself. I argued it was not an issue without reading up on the behaviour. My bad. Sorry.

PR #684, opened 2024-04-20
Prevent allocation from encoding UDP address

paulwe:


davidzhao (approved):
lg!

how do the benchmarks compare before & after?

paulwe:
45ns vs 16ns but it saves 2 allocations per packet

boks1971 (approved):

PR #688, opened 2024-04-30
Skip UDP address serialization in muxed conn

paulwe:


udp_muxed_conn.go
```diff
@@ -28,52 +34,62 @@ type udpMuxedConn struct {
...
+
+			if pkt == c.bufHead {
+				c.bufHead = nil
+			}
+			c.mu.Unlock()
+
+			if len(b) < len(pkt.buf) {
+				err = io.ErrShortBuffer
+			} else {
+				n = copy(b, pkt.buf)
```
boks1971, line 68: Should this return `io.ErrShortBuffer` if the given buffer is short?

udp_muxed_conn.go
```diff
@@ -28,52 +34,62 @@ type udpMuxedConn struct {
...
-		return 0, nil, err
-	}
-
-	dataLen := int(binary.LittleEndian.Uint16(buf.buf[:2]))
-	if dataLen > total || dataLen > len(b) {
-		return 0, nil, io.ErrShortBuffer
-	}
+	for {
+		c.mu.Lock()
+		for c.bufTail != nil {
```
boks1971, line 56: just a style comment, looks like this can be `if c.bufTail != nil`. I thought this loop around based on `for`, but looks like it does not.

udp_muxed_conn.go
```diff
@@ -178,79 +200,40 @@ func (c *udpMuxedConn) containsAddress(addr ipPort) bool {
...
+	pkt.buf = append(pkt.buf[:0], data...)
+	pkt.addr = addr
 
-	// Write address first, leaving room for its length
-	n, err := encodeUDPAddr(addr, buf.buf[offset+2:])
-	if err != nil {
-		return err
+	c.mu.Lock()
+	if c.state == udpMuxedConnClosed {
+		c.mu.Unlock()
```
boks1971, line 214: should this put back the new packet back to sync.Pool if already closed?

udp_muxed_conn.go
paulwe: it does later after the packet is returned to the pool

boks1971 (approved):
lgtm! linked-list ftw!!!

PR #703, opened 2024-06-30
Escape spaces in IPv6 ZoneID with %20

sirzooro:
Link-local IPv6 addresses may have ZoneID attached at the end. ZoneId is name of corresponding network interface. On Windows usually it is "Local connection". Space in ZoneID causes problem when ICE candidate is parsed so it has to be escaped. It looks that there is no escaping defined specifically for SDP; the closest one is RFC 6874 for URI. This fix partially uses this approach and replaces spaces with %20.

#### Reference issue
Fixes #...

### DIFF
--- a/candidate_host.go
+++ b/candidate_host.go
@@ -52,8 +52,12 @@ func NewCandidateHost(config *CandidateHostConfig) (*CandidateHost, error) {
 		network: config.Network,
 	}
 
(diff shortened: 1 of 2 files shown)


sirzooro:
I have read more about this ZoneID and found that it has only local meaning and should not be sent out over the wire. So this PR is invalid, and different fix is needed here. I will create it.

PR #704, opened 2024-06-30
Remove IPv6 ZoneID from ICE candidates

sirzooro:
Link-local IPv6 addresses may have ZoneID attached at the end. It has local meaning only and should not be send to other parties. This change removes ZoneID from generated candidate string, and ignores ZoneID when received candidate is parsed.

### DIFF
--- a/candidate_test.go
+++ b/candidate_test.go
@@ -392,7 +392,7 @@ func TestCandidateMarshal(t *testing.T) {
 
 			require.NoError(t, err)
 
-			require.True(t, test.candidate.Equal(actualCandidate))
+			require.Truef(t, test.candidate.Equal(actualCandidate), "%s != %s", test.candidate.String(), actualCandidate.String())
 			require.Equal(t, test.marshaled, actualCandidate.Marshal())
 		})
 	}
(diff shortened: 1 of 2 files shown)


candidate_test.go
```diff
@@ -437,3 +437,33 @@ func TestCandidateWriteTo(t *testing.T) {
...
+	})
+	candidateWithZoneIDStr := "750 0 udp 500 fcd9:e3b8:12ce:9fc5:74a5:c6bb:d8b:e08a%eth0 53987 typ host"
+	candidate2, err := UnmarshalCandidate(candidateWithZoneIDStr)
+	require.NoError(t, err)
+	require.Truef(t, candidate.Equal(candidate2), "%s != %s", candidate.String(), candidate2.String())
+
+	candidateWithZoneIDStr2 := "750 0 udp 500 fcd9:e3b8:12ce:9fc5:74a5:c6bb:d8b:e08a%eth0%eth1 53987 typ host"
+	candidate2, err = UnmarshalCandidate(candidateWithZoneIDStr2)
+	require.NoError(t, err)
+	require.NoError(t, err)
```
boks1971, line 467: Duplicate check

boks1971 (approved):
Lg, just a small comment

candidate_test.go
sirzooro: fixed

PR #706, opened 2024-07-01
Cleanly close agent goroutines

edaniels:
This ensures no goroutines are running when an ICE agent closes. I'd also like to backport this into v2 so that webrtc/v3 can get it.


davidzhao:
@paulwe tagging you on this as it closely matches with your initial comments about not leaving behind rogue goroutines

stv0g (approved):

PR #721, opened 2024-08-12
DRAFT - Adding more metrics into the Candidate Pair Stats / Transport Stats

danjenkins:
I'd like to do more than just add BytesSent/Received and PacketsSent/Received but we all need to start somewhere.

Before I go and make tests etc... I want someone else to look at these and say "yes, this makes sense" because I'm not certain I'm right in what I'm doing here.

Connected to https://github.com/pion/webrtc/issues/2779


philipch07:
Closed in favor of #809 (it's basically an updated version of this PR).

PR #731, opened 2024-09-16
Add round trip time measurement to candidate pair

boks1971:
Use the round trip time measurement to populate RTT fields in CandidatePairStats.

### DIFF
--- a/candidatepair.go
+++ b/candidatepair.go
@@ -32,8 +32,8 @@ type CandidatePair struct {
 	nominateOnBindingSuccess bool
 
 	// stats
-	currentRoundTripTime atomic.Pointer[time.Duration]
-	totalRoundTripTime   atomic.Pointer[time.Duration]
+	currentRoundTripTime int64 // in ns
+	totalRoundTripTime   int64 // in ns
 	responsesReceived    uint64
 }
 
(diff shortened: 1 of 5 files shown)


cnderrauber (approved):
looks good!

Sean-Der:
Thank you @boks1971 looks great! Mind adding some simple coverage to this.

Just assert values are non-zero, so someone doesn't regress it.

candidatepair.go
```diff
@@ -28,6 +30,12 @@ type CandidatePair struct {
 	state                    CandidatePairState
 	nominated                bool
 	nominateOnBindingSuccess bool
+
+	// stats
+	statsMu              sync.RWMutex
```
Sean-Der, line 35: What do you think about making these atomic (if possible?)

Having another lock on such a critical path (inbound network traffic) might eventually be targeted for optimization.

Sean-Der (changes requested):
possible to make atomics, add tests

candidatepair.go
boks1971: The STUN pings are usually once every few seconds. So, I think it should be fine.

But, I will look at atomics.

boks1971:
> possible to make atomics, add tests

@Sean-Der Can you please have a look now? Made changes in this [commit](https://github.com/pion/ice/pull/731/commits/28798100702545bbf2d5a004899e00eda55f7777).

I am still not able to squash commit the changes. Lint fails. Will fix that.

boks1971:
> > possible to make atomics, add tests
> 
> @Sean-Der Can you please have a look now? Made changes in this [commit](https://github.com/pion/ice/pull/731/commits/28798100702545bbf2d5a004899e00eda55f7777).
> 
> I am still not able to squash commit the changes. Lint fails. Will fix that.

There is a race with updating `totalRoundTripTime` because it is a read-modify-write. What is the best way to handle that with atomics?

candidatepair.go
```diff
@@ -100,3 +107,46 @@ func (a *Agent) sendSTUN(msg *stun.Message, local, remote Candidate) {
...
+// accumulates total round trip time and responses received
+func (p *CandidatePair) UpdateRoundTripTime(rtt time.Duration) {
+	p.currentRoundTripTime.Store(&rtt)
+
+	prevTotalRoundTripTime := p.totalRoundTripTime.Load()
+	totalRoundTripTime := rtt
+	if prevTotalRoundTripTime != nil {
+		totalRoundTripTime += *prevTotalRoundTripTime
+	}
+	p.totalRoundTripTime.CompareAndSwap(prevTotalRoundTripTime, &totalRoundTripTime)
```
Sean-Der, line 121: Can this just be Store?

Sean-Der (approved):

Sean-Der:
Is the race possible (since we only process one packet at a time?)

The issue I was avoiding was users being able to block recv by calling APIs.

candidatepair.go
boks1971: The following sequence will get the wrong value
1. goroutine A reads value A
2. Another update happens and goroutine B reads value A
3. goroutine A adds to the total (value A + rtt A)
4. goroutine B adds to the total (value A + rtt B) <-- this is the problematic 
5. Depending on who write first, the final value could be `value A + rtt A` or `value A + rtt B`. Both are wrong. The correct value is `value A + rtt A + rtt B`.

There is no strong reason to use `CompareAndSwap`. `Store` would store both values whatever order it happens in. `CompareAndSwap` will store only the first value.

I think I am doing to change to this to just `int64` and do `Duration.Nanoseconds()` so that I can use `atomic.AddInt64()`.

Will make that change now.

boks1971:
> Is the race possible (since we only process one packet at a time?)
> 
> The issue I was avoiding was users being able to block recv by calling APIs.

That's true. But, to make things simpler, I just changed it to `int64` nanoseconds in this [commit](https://github.com/pion/ice/pull/731/commits/47deda5d529e609fe4f3f479a750627d72d383b1). Will go ahead and use that. Can change it later if needed.

PR #739, opened 2024-10-30
Accept use-candidate for ice-lite

boks1971:
There could be a mismatch between the two ends in candidate priority
when using peer reflexive. It happens in the following scenario

1. Client has two srflx candidates.
   a. The first one gets discovered by LiveKit server as prflx.
   b. The second one gets added via ice-trickle first and then
      gets a STUN ping. So, it is srflx remote candidate from
      server's point-of-view.
2. This leads to a priority issue.
   a. Both candidates have same priority from client's point-of-view
      (both are srflx).
   b. But, from server's point-of-view, the first candidate has
      higher priority (prflx).
3. The first candidate establishes connectivity and becomes
   the selected pair (client is ICE controlling and server is
   ICE controlled, server is in ICE lite).
4. libwebrtc does a sort and switch some time later based on RTT.
   As client side has both at same priority, RTT based sorting
   could make the second candidate the preferred one.
   So, the client sends useCandidate=1 for the second candidate.
   pion/ice does not switch because the selected pair is at
   higher priority due to prflx candidate.
5. STUN pings do not happen and the ICE connection eventually fails.

ICE controlled agent should accept use-candidate unconditionally if
it is an ICE lite agent.
Just in case existing behaviour is needed, it can be configured
using `EnableUseCandidateCheckPriority`.

NOTE: With aggressive nomination, the selected pair could change
a few times, but should eventually settle on what the controlling
side wants.

### DIFF
--- a/selection.go
+++ b/selection.go
@@ -241,7 +241,7 @@ func (s *controlledSelector) HandleSuccessResponse(m *stun.Message, local, remot
 	s.log.Tracef("Found valid candidate pair: %s", p)
 	if p.nominateOnBindingSuccess {
 		if selectedPair := s.agent.getSelectedPair(); selectedPair == nil ||
-			(selectedPair != p && selectedPair.priority() <= p.priority()) {
+			(selectedPair != p && (!s.agent.needsToCheckPriorityOnNominated() || selectedPair.priority() <= p.priority())) {
 			s.agent.setSelectedPair(p)
 		} else if selectedPair != p {
 			s.log.Tracef("Ignore nominate new pair %s, already nominated pair %s", p, selectedPair)
@@ -266,7 +266,7 @@ func (s *controlledSelector) HandleBindingRequest(m *stun.Message, local, remote
 			// generated a valid pair (Section 7.2.5.3.2).  The agent sets the
 			// nominated flag value of the valid pair to true.
 			selectedPair := s.agent.getSelectedPair()
-			if selectedPair == nil || (selectedPair != p && selectedPair.priority() <= p.priority()) {
+			if selectedPair == nil || (selectedPair != p && (!s.agent.needsToCheckPriorityOnNominated() || selectedPair.priority() <= p.priority())) {
 				s.agent.setSelectedPair(p)
 			} else if selectedPair != p {
 				s.log.Tracef("Ignore nominate new pair %s, already nominated pair %s", p, selectedPair)
--- a/agent.go
+++ b/agent.go
@@ -145,6 +145,8 @@ type Agent struct {
 	insecureSkipVerify bool
 
 	proxyDialer proxy.Dialer
+
+	enableUseCandidateCheckPriority bool
 }
 
 // NewAgent creates a new Agent
@@ -219,6 +221,8 @@ func NewAgent(config *AgentConfig) (*Agent, error) { //nolint:gocognit
 		disableActiveTCP: config.DisableActiveTCP,
 
 		userBindingRequestHandler: config.BindingRequestHandler,
+
+		enableUseCandidateCheckPriority: config.EnableUseCandidateCheckPriority,
 	}
 	a.connectionStateNotifier = &handlerNotifier{connectionStateFunc: a.onConnectionStateChange, done: make(chan struct{})}
 	a.candidateNotifier = &handlerNotifier{candidateFunc: a.onCandidate, done: make(chan struct{})}
@@ -1218,4 +1222,8 @@ func (a *Agent) setGatheringState(newState GatheringState) error {
 
 	<-done
 	return nil
+}
+
+func (a *Agent) needsToCheckPriorityOnNominated() bool {
+	return !a.lite || a.enableUseCandidateCheckPriority
 }
--- a/agent_config.go
+++ b/agent_config.go
@@ -200,6 +200,13 @@ type AgentConfig struct {
 	// * Implement draft-thatcher-ice-renomination
 	// * Implement custom CandidatePair switching logic
 	BindingRequestHandler func(m *stun.Message, local, remote Candidate, pair *CandidatePair) bool
+
+	// EnableUseCandidateCheckPriority can be used to enable checking for equal or higher priority to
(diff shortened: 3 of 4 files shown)


nils-ohlmeier:
Interesting scenario. 
I assume 1.a gets added because the client uses it's host candidate to send a binding request. And 1.b takes longer because of the additional round trip time for the STUN checks to discovers it's srflx and transmitting it to the server.

The part which concerns me here is the proposed solution: I don't think that the right solution here should be to disable peer reflexive altogether. This will slow down ICE connectivity in certain cases.
In 3. you say that Pion in ICE controlled (because it is using ICE lite). So when the ICE controller sends a USE_CANDIDATE Pion needs to follow that. Priorities should only matter for the order in which candidates get tried. The controlled can not use priorities to ignore a USE_CANDIDATE from the controller.

boks1971:
> Interesting scenario. I assume 1.a gets added because the client uses it's host candidate to send a binding request. And 1.b takes longer because of the additional round trip time for the STUN checks to discovers it's srflx and transmitting it to the server.
> 
> The part which concerns me here is the proposed solution: I don't think that the right solution here should be to disable peer reflexive altogether. This will slow down ICE connectivity in certain cases. In 3. you say that Pion in ICE controlled (because it is using ICE lite). So when the ICE controller sends a USE_CANDIDATE Pion needs to follow that. Priorities should only matter for the order in which candidates get tried. The controlled can not use priorities to ignore a USE_CANDIDATE from the controller.

Thank you @nils-ohlmeier for the feedback.

That was my original inclination too (accepting `USE-CANDIDATE` - see a bit of [discussion](https://gophers.slack.com/archives/CAK2124AG/p1730310887845679) in Pion Slack). But, wanted to do something that did not change existing behaviour. But, after discussion with @Sean-Der in Slack and your feedback here, convinced that it is the right approach. So, I am going to change this PR to accept `USE-CANDIDATE` (will still have a flag to fall back to older behaviour just in case somebody wants it). Will tag you for review once I finish the changes.

There is this [long standing FIrefox issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1034964) where it does aggressive nomination even if the peer is ice-lite. I guess in that case the selected pair will change a bunch of times and settle on the final one. Is that correct?

boks1971:
@nils-ohlmeier Please have a look at this PR when you get a chance and let me know if it looks okay.

cnderrauber:
I think we'd better to make it as a default behavior for controlled ice-lite agent insead of all controlled agent. The failure case is chrome/firefox does renomiantion/aggressive-nomination when remote peer is ice lite, and there is no problem for full agent. So the current behavior of pion comply with the rfc, we don't need to break it for the bad behavior cause by remote peer.

cnderrauber:
Compare priority is ok for both nomiation type in rfc5245:
* regular: there are only 1 nomiate so it is ok
* aggressive: `ICE always selects the highest-priority nominated candidate pair from
   the valid list as the one used for media`
Checked the libwebrtc code, it also compares the nominated connections from remote side when it picks controlled role, the conditions are complex but contains the priority.

boks1971:
@cnderrauber Thank you for the review. I have updated the code to apply it only for ice-lite agent. Can you please re-review when you get a chance?

selection.go
```diff
@@ -241,7 +241,7 @@ func (s *controlledSelector) HandleSuccessResponse(m *stun.Message, local, remot
 	s.log.Tracef("Found valid candidate pair: %s", p)
 	if p.nominateOnBindingSuccess {
 		if selectedPair := s.agent.getSelectedPair(); selectedPair == nil ||
-			(selectedPair != p && selectedPair.priority() <= p.priority()) {
+			(selectedPair != p && ((!s.agent.enableUseCandidateCheckPriority && s.agent.lite) || selectedPair.priority() <= p.priority())) {
```
cnderrauber, line 244: how about make it a agent method `needsToCheckPriorityOnNominated`

cnderrauber (approved):
looks good!

selection.go
boks1971: Good call @cnderrauber . Updated with a helper function.

boks1971:
@Sean-Der @davidzhao @nils-ohlmeier going to merge this based on discussions in Slack/this PR and @cnderrauber 's review. Want to port this to v2 and start using it in LiveKit project. But, please do review when you get a chance and let me know if you have feedback and I will address it in a subsequent PR.

Sean-Der:
@boks1971 sounds good to me! 

What do you need to move to /v3? If I port livekit/livekit would that be a good start?

boks1971:
Nothing specifically needed @Sean-Der as the default behaviour is the desired behaviour. I cherry-picked the change to [v2](https://github.com/pion/ice/commit/1e0d4c6883a8f4d33870b884db1a83c1214f6de7) and tagged it [v2.3.37](https://github.com/pion/ice/releases/tag/v2.3.37) and have updated [livekit dependencies](https://github.com/livekit/livekit/commit/150433ab6477126abdaad4582b1d340af670452e) to use that tag.

boks1971:
I was debating adding a settings engine setting to enable the priority check (the flag added in this PR) and thread it through. But, as we discussed the default behaviour added in this PR looks like the right one and the flag is just a safety net. So, decided to wait to see if there are any ill effects before adding another things to SettingsEngine.

nils-ohlmeier:
> There is this [long standing FIrefox issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1034964) where it does aggressive nomination even if the peer is ice-lite. I guess in that case the selected pair will change a bunch of times and settle on the final one. Is that correct? 

Yeah that one sucks. The problem is that Firefox uses its own ICE stack written in C with lots of pointer magic. Most of the developers don't want to touch it with a ten foot pole. The team is aware of the problem. But my guess is that it is rather unlikely this will get fixed anytime soon. 

The best known workaround is to implement on the ice-lite server side to use the candidate pair where it receives traffic from. I'm not sure if a workaround like that should go into Pion to be more compatible with Firefox. @Sean-Der do you have thoughts on this?

boks1971:
I did that approach in a previous life and it worked well. But, that was a few years ago.

PR #744, opened 2024-11-26
#742 net: pass ipv6 zone param to net.ListenUDP

### LINKED ISSUE #742: ipv6: trying to listen on link-local addresses without interface
### Your environment.
  - Version: 4.0.2
### What did you do?
Hi! We recently updated to pion v4 and during load test noticed high CPU consumption. CPU-profile showed us that problem function was [listenUDPInPortRange](https://github.com/pion/ice/blob/master/net.go#L120) while gathering local candidates. After some research I found out that new version is trying to use ipv6 link-local addresses as candidates. However, as you may know, you can't listen on ipv6 link-local address without specifying interface. If you don't specify it during syscall `socket`, then you get the error `bind: can't assign requested address`. The problem is, the [current implementation of  `listenUDPInPortRange`](https://github.com/pion/ice/blob/master/net.go#L140) is not passing the zone parameter to the `ListenUDP` call. As a result an ICE agent has to check ALL ports from the requested range, which is quite CPU intensive. Fix is straightforward, but maybe there are some other places where the zone parameter is missing.

N1cOs:
#### Description
Pass ipv6 zone param to `net.ListenUDP`. Also checked that it's specified in all other calls of `net.ListenUDP`. Tested it locally with ipv6 link-local and global unicast addresses. 

However, should I add a unit test for that @Sean-Der? I could add it in the [gather_test.go](https://github.com/pion/ice/blob/master/gather_test.go#L33), but it would be host specific and require an assigned ipv6 address.

Also don't forget to backport to v3, because ipv6 link-local addresses appeared in v3.

#### Reference issue
Fixes #742

### DIFF
--- a/net.go
+++ b/net.go
@@ -137,8 +137,13 @@ func listenUDPInPortRange(n transport.Net, log logging.LeveledLogger, portMax, p
 	portStart := globalMathRandomGenerator.Intn(j-i+1) + i
 	portCurrent := portStart
 	for {
-		lAddr = &net.UDPAddr{IP: lAddr.IP, Port: portCurrent}
-		c, e := n.ListenUDP(network, lAddr)
+		addr := &net.UDPAddr{
+			IP:   lAddr.IP,
+			Zone: lAddr.Zone,
+			Port: portCurrent,
+		}
+
+		c, e := n.ListenUDP(network, addr)
 		if e == nil {
 			return c, e //nolint:nilerr
 		}


Sean-Der:
@N1cOs Can you try pion/webrtc/v4@v4.0.3 now?

N1cOs:
I will check in a few days, but on v4.0.2 ignoring link-local addresses definitely fixes the high CPU consumption on connection initiation

Issue #753, opened 2025-01-14
Add Support for Parsing Extensions in Candidates


joeturki:
The UnmarshalCandidate function should be improved to align more y with the ICE Candidate parsing rules outlined in [RFC 5245, Section 15.1](https://datatracker.ietf.org/doc/html/rfc5245#section-15.1). 

In addition, candidateBase should support access to extension values. This includes:

Parsing and storing extensions in a structured way.
Matching specific extensions for validation or other purposes.
Serializing (marshaling) the candidate, including its extensions, back into a string.

This is required for many issues like https://github.com/pion/webrtc/issues/2993

I'm making a PR.

PR #754, opened 2025-01-14
Parse Candidate Extensions (RFC5245)

joeturki:
#### Description
- Rewrote `UnmarshalCandidate` to better align with RFC5245.
- Added new methods for extracting candidate extensions.
- Updated `Equal` and `Marshal` to accommodate these changes.
- Added tests for everything. 
- Fixed a "bug" where tcptype would appear before raddr .
- Made a minor change to gather.go; because we can't use the config struct as a map key anymore.
- Added `CandidateExtension`.

I'm aware this is a breaking change but this change is required for many issues like https://github.com/pion/webrtc/issues/2993

Edit: I'll make it non breaking change tomorrow. 

#### Reference issue

https://github.com/pion/ice/issues/753


candidate_base.go
```diff
@@ -10,6 +10,7 @@ import (
 	"hash/crc32"
 	"io"
 	"net"
+	"reflect"
```
Sean-Der, line 13: Just started reading, but would prefer to never use reflection.

Makes things harder on tinygo and we lose safety

candidate.go
```diff
@@ -52,6 +52,17 @@ type Candidate interface {
 	//  candidate, which is useful for diagnostics and other purposes
 	RelatedAddress() *CandidateRelatedAddress
 
+	// Extensions returns a copy of all extension attributes associated with the ICECandidate.
+	// In the order of insertion, *(key value).
+	// Extension attributes are defined in RFC 5245, Section 15.1:
+	// https://datatracker.ietf.org/doc/html/rfc5245#section-15.1
+	Extensions() []string
```
Sean-Der, line 59: What do you think of defining a `Extension` struct? Gives you flexibility to add more fields in the future

candidate_base.go
```diff
@@ -496,103 +503,382 @@ func (c *candidateBase) Marshal() string {
...
-// UnmarshalCandidate creates a Candidate from its string representation
+func (c *candidateBase) Extensions() []string {
+	extensions := make([]string, 0, len(c.extensions))
+	hasTCPTYPE := false
+
+	for i := 0; i < len(c.extensions); i += 2 {
+		key := c.extensions[i]
+
+		if i+1 >= len(c.extensions) {
+			// should never happen
```
Sean-Der, line 530: I am on phone, so don’t fully understand code yet. Can/should we return an error in bad situation?

Sean-Der:
Nice, love these improvements :)

Glad we can do this without API break. It means people don’t update and new versions don’t get improved :(

candidate_base.go
```diff
@@ -406,12 +409,16 @@ func (c *candidateBase) Equal(other Candidate) bool {
...
 	return c.NetworkType() == other.NetworkType() &&
 		c.Type() == other.Type() &&
 		c.Address() == other.Address() &&
 		c.Port() == other.Port() &&
+		// Keeping this for the sake of backwards compatibility
+		// We don't need to compare TCPType anymore, because it's included in the extensions
 		c.TCPType() == other.TCPType() &&
-		c.RelatedAddress().Equal(other.RelatedAddress())
+		c.RelatedAddress().Equal(other.RelatedAddress()) &&
+		reflect.DeepEqual(c.Extensions(), other.Extensions())
```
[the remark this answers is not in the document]
joeturki: I added DeepEqual

candidate_base.go
```diff
@@ -496,103 +503,382 @@ func (c *candidateBase) Marshal() string {
...
+			uint16(component), //nolint:gosec // G115 no overflow we read 5 digits
+			uint32(priority),  //nolint:gosec // G115 no overflow we read 5 digits
+			foundation,
+			parserExt.raddr,
+			parserExt.rport,
+			"",
+			nil,
+			parserExt.extensions,
+		})
 	default:
```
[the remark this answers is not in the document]
joeturki: That was inherited code, fixed thanks :)

candidate.go
joeturki: Done

joeturki:
I added `CandidateExtensions` and `CandidateExtension`, And fixed all the other issues.

@Sean-Der @nils-ohlmeier let me know if there are other problems :)

Thanks.

candidate.go
Sean-Der: Sorry I wouldn’t do `CandidateExtensions‘ as a type. You can do just the slice

Also maybe don’t do a pointer to the slice. A empty slice means no extensions exist is good!

joeturki: @Sean-Der so Extensions() returns a slice of CandidateExtension{key, value} (maybe private?) and GetExtension(key string)?

Sean-Der: I think that makes sense, tell me if you disagree 

* Extensions as a type Allows us to add methods/more fields to Extensions without breaking API
* Returning slice is nice because reading the doc I don’t have to look up  as much. I don’t see a benefit to a type here

candidate_base.go
```diff
@@ -45,6 +45,7 @@ type candidateBase struct {
 
 	remoteCandidateCaches map[AddrPort]Candidate
 	isLocationTracked     bool
+	extensions            *CandidateExtensions
```
Sean-Der, line 48: I would just do `[]CandidateExtension` does not need to be a pointer

nils-ohlmeier (approved):
Looks good to me.

candidateextensions.go
```diff
@@ -0,0 +1,141 @@
...
+type CandidateExtension struct {
+	key   string
+	value string
+}
+
+// UnmarshalCandidateExtensions parses the candidate extensions from the raw string.
+// *(SP extension-att-name SP extension-att-value)
+// Where extension-att-name, and extension-att-value are byte-strings
+// as defined in https://tools.ietf.org/html/rfc5245#section-15.1
+func UnmarshalCandidateExtensions(raw string) (*CandidateExtensions, error) {
```
Sean-Der, line 30: Do you need to expose?

candidate.go
joeturki: Okay, so I'm dumb I thought you meant by struct (CandidateExtensions), not a slice of structs, fixing that 👍

Sean-Der (approved):
Fantastic! This is great.

Sorry about the stress of APIs. It’s just so hard to fix them. We can always fix/improve code though.

Its nice to expose as little as possible until someone requests it :)

joeturki:
@Sean-Der When you have a time can you take a look.

> Sorry about the stress of APIs. It’s just so hard to fix them. We can always fix/improve code though.

It's not stressing at all for me :) 

> Its nice to expose as little as possible until someone requests it :)

This makes sense, Noted :)

candidate.go
```diff
@@ -52,12 +52,28 @@ type Candidate interface {
...
+	// Extension attributes are defined in RFC 5245, Section 15.1:
+	// https://datatracker.ietf.org/doc/html/rfc5245#section-15.1
+	//.
+	Extensions() []CandidateExtension
+
+	// GetExtension returns the value of the extension attribute associated with the ICECandidate.
+	// Extension attributes are defined in RFC 5245, Section 15.1:
+	// https://datatracker.ietf.org/doc/html/rfc5245#section-15.1
+	//.
+	GetExtension(key string) (value string, ok bool)
```
seaduboi-openai, line 66: Mind doing `GetExtension(key string) (value CandidateExtension, ok bool)`

Sean-Der: Sorry this is me, was on work account.

Sean-Der:
@joeturki sorry one last change to keep return types consistent. After that merge it! If we do have bugs/improvements to make we call roll forward :)

sorry for all the rounds on this.

joeturki:
@Sean-Der 
Hopefully, this is my first API change of many :) 

I now have a better understanding of how changes are made to the Pion API, I'll take all feedback into consideration to try to make future iterations shorter. 

Thanks!

PR #763, opened 2025-02-27
Add Req/Res count/time to candidate stats

cnderrauber:
These details will provide information for understanding
connectivity issue.

### DIFF
--- a/candidatepair.go
+++ b/candidatepair.go
@@ -33,7 +33,18 @@ type CandidatePair struct {
 	// stats
 	currentRoundTripTime int64 // in ns
 	totalRoundTripTime   int64 // in ns
-	responsesReceived    uint64
+
+	requestsReceived  uint64
+	requestsSent      uint64
+	responsesReceived uint64
+	responsesSent     uint64
+
+	firstRequestSentAt     atomic.Value // time.Time
+	lastRequestSentAt      atomic.Value // time.Time
+	firstReponseReceivedAt atomic.Value // time.Time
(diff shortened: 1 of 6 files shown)


candidatepair.go
```diff
@@ -141,8 +156,101 @@ func (p *CandidatePair) TotalRoundTripTime() float64 {
...
+func (p *CandidatePair) UpdateRequestSent() {
+	atomic.AddUint64(&p.requestsSent, 1)
+	now := time.Now()
+	p.firstRequestSentAt.CompareAndSwap(nil, now)
+	p.lastRequestSentAt.Store(now)
+}
+
+// UpdateResponseSent increments the number of responses sent.
+func (p *CandidatePair) UpdateResponseSent() {
+	atomic.AddUint64(&p.responsesSent, 1)
```
boks1971, line 247: Is there any value to adding first/last response sent time?

With this PR, we have
- first/last time of request sent
- first/last time of request received
- first/last time of response received

wondering if there is more to learn by recording first/last time of response sent?

boks1971 (approved):
nice!!!

candidatepair.go
cnderrauber: I had it but deleted at last since the agent always send response when request is received, so the timpstamp is useless (the `responseSent` has same issue too but it is in the specification https://www.w3.org/TR/webrtc-stats/#dom-rtcicecandidatepairstats-responsessent so I keep it).

boks1971: got it, thank you @cnderrauber

PR #767, opened 2025-03-17
Always send KeepAlives

oto313:
Pion incorrectly resets the consent timer when sending any traffic. 
The consent timer must only be reset on STUN traffic.

RFC 7675
> Consent expires after 30 seconds.  That is, if a valid STUN binding
> response has not been received from the remote peer's transport
> address in 30 seconds, the endpoint MUST cease transmission on that
> 5-tuple.  STUN consent responses received after consent expiry do not
> re-establish consent and may be discarded or cause an ICMP error.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -594,9 +594,7 @@ func (a *Agent) checkKeepalive() {
 		return
 	}
 
-	if (a.keepaliveInterval != 0) &&
-		((time.Since(selectedPair.Local.LastSent()) > a.keepaliveInterval) ||
-			(time.Since(selectedPair.Remote.LastReceived()) > a.keepaliveInterval)) {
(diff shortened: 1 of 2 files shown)


Sean-Der:
Thank you so much @oto313!

Would it be possible to make a small test out of this? My fear is that this will break again in the future :(

If it isn't easy we can merge now, and add a ticket. Better to have it fixed then block for a test.

thank you!

oto313:
Sure @Sean-Der. Take a look if the test is enough. It fails when this patch is not merged and passes when it is merged. 

Thanks

Sean-Der:
@oto313 Thank you! Tagged as `v4.0.8`

PR #773, opened 2025-04-16
Avoid unnecessary turn address resolution in the turn client

amanakin:
#### Description

The `turn.NewClient` function in the library currently requires TURN server addresses passed to it to be resolvable. This causes an error when using a proxy, as the TURN address might only be resolvable by the proxy server itself, not by the client application initiating the connection.

To avoid this error in proxy scenarios, this PR modifies the initialization to omit the TURN address when creating the `turn.Client`.

Blocker:
This change depends on functionality introduced in `pion/turn` Pull Request [#450](https://github.com/pion/turn/pull/450).

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -690,6 +690,8 @@ func (a *Agent) gatherCandidatesRelay(ctx context.Context, urls []*stun.URI) {
 				relAddr       string
 				relPort       int
 				relayProtocol string
+
+				omitTurnServerAddr bool
 			)
 
 			switch {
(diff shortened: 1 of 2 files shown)


gather.go
```diff
@@ -721,6 +723,11 @@ func (a *Agent) gatherCandidatesRelay(ctx context.Context, urls []*stun.URI) {
 				}
 				locConn = turn.NewSTUNConn(conn)
 
+				// We don't need to resolve address of the turn server, since we are using
+				// proxy dialer.
+				if _, err = a.net.ResolveUDPAddr("udp4", turnServerAddr); err != nil {
```
Sean-Der, line 728: Why can't we always omit the TURN Server Addr now?

Sean-Der:
@amanakin I released a new version of `pion/turn` if you want to update the go.mod!

gather.go
amanakin: In the proxy case, the TURN address is only used only for debug logging. So maybe completely skipping it is too much? Some might find having the address in the logs useful for debugging. What do you think?

amanakin: The only viable approach I can see on the turn library side is to add an `IgnoreTURNResolveError` flag to this configuration struct: https://github.com/pion/turn/blob/master/client.go#L39

And it breaks the minimalism of the config, unfortunately :(

amanakin: @Sean-Der Sorry for the delay.
We can try to do something like that https://github.com/pion/turn/pull/455 on the TURN package.
And then in the ICE package we will just propagate `IgnoreTURNResolveErrors=true` in case of a proxy connection.

WDYT?

Sean-Der (approved):

PR #781, opened 2025-06-02
Allow cancelling TURN Allocate phase, which can block for ~7.8s if TURN server is not responding.

smb-z:
#### Description

Although [Agent.GatherCandidates()](https://github.com/pion/ice/blob/753c2a0fe868006a9e8d341feb11fc6a5d54a45d/gather.go#L39) have a cancellable context, the [turn.Client.Allocate()](https://github.com/pion/turn/blob/61a25e754f7a9c4e6ba5a4b8ee18579899985339/client.go#L325) processing does not use it and can hang for about 7.8 seconds if the TURN server is unreachable.
This happens because Allocate() completes only after the ALLOCATE request retry limit is reached.
Ultimately, PeerConnection.Close() will block while waiting for unnecessary allocation to complete.

PR fixes this behavior by cancelling ALLOCATE phase if context was cancelled.

Caveat: If cancelled, Client.Allocate will nevertheless return error ("transaction closed") which is logged as warning [here](https://github.com/pion/ice/blob/753c2a0fe868006a9e8d341feb11fc6a5d54a45d/gather.go#L842).

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -836,7 +836,17 @@ func (a *Agent) gatherCandidatesRelay(ctx context.Context, urls []*stun.URI) {
 				return
 			}
 
+			allocDoneCh := make(chan struct{})
+			go func() {
+				select {
+				case <-ctx.Done():
+					client.Close()
+				case <-allocDoneCh:
+				}
+			}()
+
 			relayConn, err := client.Allocate()
+			close(allocDoneCh)
 			if err != nil {
 				client.Close()
 				closeConnAndLog(locConn, a.log, "failed to allocate on TURN client %s %s", turnServerAddr, err)
--- a/gather_vnet_test.go
+++ b/gather_vnet_test.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"net"
 	"testing"
+	"time"
 
 	"github.com/pion/logging"
 	"github.com/pion/stun/v3"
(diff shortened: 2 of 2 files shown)


smb-z:
Thanks, @JoeTurki!
It seems you're right, and cancelling client.Allocate() is more tricky than I first thought. Here are some race issues I've found:

1. If the context is cancelled just after client.Allocate() successfully completes but before it returns, we'll end up in a state when err [here](https://github.com/pion/ice/blob/753c2a0fe868006a9e8d341feb11fc6a5d54a45d/gather.go#L840) is nil, but the client is closed by the goroutine. I wasn't able to find any significant problems with this, but it looks unreliable, as the code following the error check is expecting Allocate to complete successfully. Thanks, @aalekseevx for pointing this out!
2. If the context is cancelled before client.Allocate(), it wouldn't have any effect, and we'll still have client.Allocate() blocked for 7.8s.

It looks like a better option is to pass the context to Allocate() and process context cancelling inside the TURN client itself.

Sean-Der:
I am going to add `Context` support to `pion/turn` client. Then will update this PR

Sean-Der:
Thank you for filing this @smb-z :) 

I ended up making the `context` change! I added you as a co-author on the commit since you did a bunch of work.

PR #782, opened 2025-06-14
Change activeTCPConn.close to atomic.Bool

cppcoffee:
Replace manual atomic operations with atomic.Bool type for better type safety and cleaner code. This modernizes the atomic usage pattern from atomic.LoadInt32/StoreInt32 to the newer Load/Store methods on atomic.Bool.

- Update activeTCPConn.closed field type from int32 to atomic.Bool
- Replace atomic.LoadInt32(&a.closed) with a.closed.Load()
- Replace atomic.StoreInt32(&a.closed, 1) with a.closed.Store(true)

All existing functionality preserved with improved type safety.

### DIFF
--- a/active_tcp.go
+++ b/active_tcp.go
@@ -18,7 +18,7 @@ import (
 type activeTCPConn struct {
 	readBuffer, writeBuffer *packetio.Buffer
 	localAddr, remoteAddr   atomic.Value
-	closed                  int32
+	closed                  atomic.Bool
 }
 
 func newActiveTCPConn(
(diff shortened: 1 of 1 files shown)


active_tcp.go
```diff
@@ -18,7 +18,7 @@ import (
 type activeTCPConn struct {
 	readBuffer, writeBuffer *packetio.Buffer
 	localAddr, remoteAddr   atomic.Value
-	closed                  int32
+	closed                  atomic.Int32
```
Sean-Der: Would it be possible to use `atomic.Bool`?

cppcoffee: It is indeed possible to use `atomic.bool`

cppcoffee:
Changed to `atomic.Bool`.

JoeTurki (approved):

PR #784, opened 2025-06-14
Enable any type alias in golangci-lint config

cppcoffee:
Add detection of the any type in the CI review.


Sean-Der:
Nice! Can you change in goassets repo? Then it gets synced to every repo.

PR #788, opened 2025-06-27
Optimize slice allocation in UDPMuxDefault

cppcoffee:
Pre-allocate localAddrsForUnspecified slice with known capacity and use index assignment instead of append to avoid multiple slice reallocations.


Sean-Der:
Thank you @cppcoffee !

Sean-Der (approved):

PR #789, opened 2025-07-16
Implement ICE Role conflict resolution

### LINKED ISSUE #359: How pion webrtc handle ice role conflict
pion webrtc environment.
Version: v2@v2.2.26

What did you do?
we use pion webrtc as  client，ice role is contrlolling，remote webrtc as server，ice role is controlling，in this case，an end-to-end not solve the problem of conflict of ice, pion webrtc no solution to the conflict, causing the ice connection to fail.

What did you expect?
Hope that the pion webrtc can solve the problem of ice role conflict.

Sean-Der:
Detect if remote has a role conflict and resolve it as defined by RFC 8445 section-7.3.1.1

Resolves #359

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -62,7 +62,7 @@ type Agent struct {
 	muHaveStarted sync.Mutex
 	startedCh     <-chan struct{}
 	startedFn     func()
-	isControlling bool
+	isControlling atomic.Bool
 
 	maxBindingRequests uint16
 
@@ -104,7 +104,9 @@ type Agent struct {
 	remoteCandidates map[NetworkType][]Candidate
 
 	checklist []*CandidatePair
-	selector  pairCandidateSelector
+
+	selectorLock sync.RWMutex
+	selector     pairCandidateSelector
 
 	selectedPair atomic.Value // *CandidatePair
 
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -1064,6 +1056,33 @@ func (a *Agent) handleInboundBindingSuccess(id [stun.TransactionIDSize]byte) (bo
...
 
+func (a *Agent) handleRoleConflict(msg *stun.Message, local, remote Candidate, remoteTieBreaker *AttrControl) {
+	localIsGreater := a.tieBreaker > remoteTieBreaker.Tiebreaker
+	a.log.Warnf("Role conflict local and remote same role(%s), localIsGreater(%t)", a.role(), localIsGreater)
+
+	// https://datatracker.ietf.org/doc/html/rfc8445#section-7.3.1.1
+	//
+	// * If the agent's tiebreaker value is larger than or equal to the contents of the ICE-CONTROLLING attribute
+	// * If the agent's tiebreaker value is less than the contents of the ICE-CONTROLLED attribute
+	if (a.isControlling.Load() && localIsGreater) || (!a.isControlling.Load() && !localIsGreater) {
```
britblue: According to RFC, the error code message should be sent and the role retained if controlling and tie breaker greater or equal OR not controlling and tie breaker less. This code is doing the opposite. Can you double check?
Also, what happens when the remote sends the ERROR code to pion? Is that scenario being handled?

Sean-Der: Fixed! Flipped the conditional @britblue thanks for catching.

Sean-Der:
Tested against Chrome and flipped roles as expected. I forced both sides to Controlling and flipped as expected

agent.go
```diff
@@ -1064,9 +1056,39 @@ func (a *Agent) handleInboundBindingSuccess(id [stun.TransactionIDSize]byte) (bo
...
+
+	// https://datatracker.ietf.org/doc/html/rfc8445#section-7.3.1.1
+	//  An agent MUST examine the Binding request for either the ICE-
+	//  CONTROLLING or ICE-CONTROLLED attribute.  It MUST follow these
+	// procedures:
+
+	// If the agent's tiebreaker value is larger than or equal to the contents of the ICE-CONTROLLING attribute
+	// If the agent's tiebreaker value is less than the contents of the ICE-CONTROLLED attribute
+	//  the agent generates a Binding error response
+	if (a.isControlling.Load() && localIsGreater) || (!a.isControlling.Load() && !localIsGreater) {
```
britblue: The RFC explicitly states greater or equal. To avoid future bugs in this logic, I would recommend we update the `localIsGreater` logic above to be `localIsGreaterOrEqual := a.tieBreaker >= remoteTieBreaker.TieBreaker`

Sean-Der: @britblue Fixed! Mind giving it another look. When you are happy with it I will merge.

Do you need a pack port to /v3 for this?

britblue (approved):

PR #799, opened 2025-08-17
Add support for renomination

kevmo314:
#### Description

Adds support for renomination following https://datatracker.ietf.org/doc/html/draft-thatcher-ice-renomination-01.

### DIFF
--- a/renomination.go
+++ b/renomination.go
@@ -1,0 +1,86 @@
+// SPDX-FileCopyrightText: 2025 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package ice
+
+import (
+	"fmt"
+
+	"github.com/pion/stun/v3"
+)
+
+// Default STUN Nomination attribute type for ICE renomination.
+// Following the specification draft-thatcher-ice-renomination-01.
+const (
+	// DefaultNominationAttribute represents the default STUN Nomination attribute.
+	// This is a custom attribute for ICE renomination support.
+	// This value can be overridden via AgentConfig.NominationAttribute.
+	DefaultNominationAttribute stun.AttrType = 0x0030 // Using a value in the reserved range
+)
+
+// NominationAttribute represents a STUN Nomination attribute.
+type NominationAttribute struct {
+	Value uint32
+}
+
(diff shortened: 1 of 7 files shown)


JoeTurki:
It's nice to see you again Kevin <3 
I'll have to read the spec first before I review it.

renomination.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/pion/stun/v3"
+)
+
+// STUN Nomination attribute type for ICE renomination.
+// Following the specification draft-thatcher-ice-renomination-01.
+const (
+	// AttrNomination represents the STUN Nomination attribute.
+	// This is a custom attribute for ICE renomination support.
+	AttrNomination stun.AttrType = 0x0030 // Using a value in the reserved range
```
JoTurk: perhaps we should make this configurable, at least until the attribute for renomination is officially assigned..

JoeTurki (approved):
looks good, thank you :)

renomination.go
kevmo314: Would we be coupled to that API forever since it has to be backwards-compatible?

JoTurk: hmm good point, maybe we could use the settings pattern to set it, and add an option `SetNominationAttr(...) error`, and in the future return an error? idk it's a suggestion :)

kevmo314: I updated it to use an option pattern in the agent constructor. It's a little more verbose but allows some more flexible configuration :)

JoTurk: This is cool, i wanted to do this so I can work on ICE issues, and features requests like port mapping, but i would rather if we got rid of the AgentConfig.

what do you think? ik that's a lot of work sorry, I can help :)

kevmo314: How are you thinking we should migrate? I'm down to help, I can do it in a separate PR. Maybe one idea would be to remove `AgentConfig` from the `WithOptions()` constructor and slowly move everything in that direction?

JoTurk: > , I can do it in a separate PR. Maybe one idea would be to remove AgentConfig from the WithOptions() constructor and slowly move everything in that direction?

Yes this makes sense.

Thank you.

PR #810, opened 2025-09-13
Improve test coverage for various tests

philipch07:
#### Description

Improves code coverage for various tests.

#### Reference issue
n/a

### DIFF
--- a/tcp_mux_multi_test.go
+++ b/tcp_mux_multi_test.go
@@ -7,6 +7,7 @@
 package ice
 
 import (
+	"errors"
 	"io"
 	"net"
 	"testing"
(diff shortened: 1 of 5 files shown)


tcp_packet_conn_test.go
```diff
@@ -0,0 +1,232 @@
...
+
+	under := newBufferedConn(c1, 0, logger)
+
+	bc, ok := under.(*bufferedConn)
+	require.True(t, ok, "expected *bufferedConn")
+
+	_ = bc.buf.SetReadDeadline(time.Unix(0, 0))
+
+	// wait for goroutine to hit Read -> error path
+	time.Sleep(10 * time.Millisecond)
```
JoTurk: This will be a bit flaky if the action is slow, we have some sleep patterns in tests that we can use instead.

tcp_mux_multi_test.go
```diff
@@ -126,3 +127,139 @@ func TestMultiTCPMux_NoDeadlockWhenClosingUnusedPacketConn(t *testing.T) {
...
+
+func TestMultiTCPMux_GetConnByUfrag_NoMuxes(t *testing.T) {
+	multi := NewMultiTCPMuxDefault() // no muxes
+
+	pc, err := multi.GetConnByUfrag("ufrag", false, net.IP{127, 0, 0, 1})
+	require.Nil(t, pc)
+	require.ErrorIs(t, err, errNoTCPMuxAvailable)
+}
+
+func TestMultiTCPMux_GetConnByUfrag_UsesFirstMux(t *testing.T) {
```
JoTurk: Maybe instead of asserting it's the first mux, it should be from any mux, because we change the selection policy I'm not sure if this behavior should be defined, what do you think?

tcp_packet_conn_test.go
philipch07: Would you happen to have any examples of this? I do agree that it would be more ideal to have a different sleeping approach than this.

tcp_mux_multi_test.go
philipch07: Good point, I changed it to check for any mux so we can account for different selection policies. I initially had it assert the first mux as a starting point.

tcp_packet_conn_test.go
JoTurk: In this example you can make a simple proxy struct for net.Conn, 
e.g:
```
type readNotifyConn struct {
	net.Conn
	once       sync.Once
	readCalled chan struct{}
...
```
and proxy `(r *readNotifyConn) Read(b []byte) (int, error) {` to close the readChannel once in a `once.Do(`, and use this interface with `newBufferedConn`.

If it's a lot of work we can always fix it later, It's just that github actions are sometimes slow / flaky, we have many sleep like this but we just try to limit them in future tests :)

JoeTurki (approved):

tcp_packet_conn_test.go
philipch07: I see, thanks for the in depth example. I just added it, let me know what you think!

JoTurk: dope, i like how you validate that we log the error.

PR #811, opened 2025-09-14
Add more tests to improve code coverage

philipch07:
#### Description
Similar to my previous effort, this adds tests to improve code coverage.

The breaking change was moved to #812 
The slices rework was moved to #813 

#### Reference issue
N/A

### DIFF
--- a/active_tcp_test.go
+++ b/active_tcp_test.go
@@ -7,7 +7,9 @@
 package ice
 
 import (
+	"context"
 	"fmt"
+	"io"
 	"net"
 	"net/netip"
 	"sync/atomic"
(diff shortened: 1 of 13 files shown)


active_tcp_test.go
```diff
@@ -290,3 +292,121 @@ func TestActiveTCP_Respect_NetworkTypes(t *testing.T) {
...
+
+	a := newActiveTCPConn(ctx, "127.0.0.1:0", ra, logger)
+	require.NotNil(t, a)
+
+	srvConn, err := tcpListener.Accept()
+	require.NoError(t, err)
+
+	_ = srvConn.Close()
+
+	_, _ = a.WriteTo([]byte("data"), nil)
```
JoTurk: Maybe we need to check / assert this error and the others.

JoeTurki (approved):

active_tcp_test.go
philipch07: Thanks for the heads up, I forgot about this. I just added the checks like you suggested.

JoTurk: Thanks

PR #813, opened 2025-09-14
Replace verbose check with slices.Contains in agent_config

philipch07:
#### Description
Simplifies some logic in `agent_config.go`. I'm assuming it was written before go 1.21 introduced this, but it's equivalent logic and is easier to read.

#### Reference issue
N/A

### DIFF
--- a/agent_config.go
+++ b/agent_config.go
@@ -5,6 +5,7 @@ package ice
 
 import (
 	"net"
+	"slices"
 	"time"
 
 	"github.com/pion/logging"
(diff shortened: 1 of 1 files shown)


JoeTurki:
It can be made through a simple rule guard but it will be a lot of work because we'll have to update it everywhere, merging this is good for now \:)

philipch07:
> It can be made through a simple rule guard but it will be a lot of work because we'll have to update it everywhere, merging this is good for now :)

Good idea, I can look into this.

PR #819, opened 2025-10-05
Set default accept wait for relay-only candidate to 0

### LINKED ISSUE #324: Pion should skip the two-second delay when ICETransportPolicy=ICETransportPolicyRelay
When looping back through a TURN relay, Chrome takes a few milliseconds to connect: https://github.com/jech/galene/blob/master/static/galene.js#L2307

The equivalent code on the server side takes 2 seconds: https://github.com/jech/galene/blob/master/ice/ice.go#L127

The reason for that is that the TURN delay is fixed to two seconds (`ice.defaultRelayAcceptanceMinWait`), and this delay is honoured even when `ICETransportPolicy` is set to relay.

adeithe:
#### Description
Defaults relay-only agent candidate types acceptance wait to zero

#### Reference issue
Fixes #324

### DIFF
--- a/agent_config.go
+++ b/agent_config.go
@@ -39,6 +39,9 @@ const (
 	// defaultRelayAcceptanceMinWait is the wait time before nominating a relay candidate.
 	defaultRelayAcceptanceMinWait = 2000 * time.Millisecond
 
+	// defaultRelayOnlyAcceptanceMinWait is the wait time before nominating with a relay only candidate.
+	defaultRelayOnlyAcceptanceMinWait = time.Duration(0)
+
 	// defaultSTUNGatherTimeout is the wait time for STUN responses.
 	defaultSTUNGatherTimeout = 5 * time.Second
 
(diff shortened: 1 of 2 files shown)


agent_config.go
```diff
@@ -294,19 +301,22 @@ func (config *AgentConfig) initExtIPMapping(agent *Agent) error { //nolint:cyclo
...
+	case CandidateTypeServerReflexive:
 		candiSrflxEnabled := slices.Contains(agent.candidateTypes, CandidateTypeServerReflexive)
 		if !candiSrflxEnabled {
 			return ErrIneffectiveNAT1To1IPMappingSrflx
 		}
+	default:
+		return nil
 	}
 
 	return nil
```
philipch07, line 322: Since this is a switch now, is it still necessary to have the `return nil` at the very end?

adeithe: Since not all cases return a value, yeah

JoeTurki (approved):
lgtm, please squash all the commits into a single one and you should be able to merge it, thank you :)

agent_config.go
JoTurk: yeah, also we can't remove the default case because the linter will complain, this looks good.

philipch07: Oh yeah that makes sense. I wonder if it would be more clear to just use an if/else-if instead, but I'm ok with leaving it as is!

JoTurk: > Oh yeah that makes sense. I wonder if it would be more clear to just use an if/else-if instead, but I'm ok with leaving it as is!

The linter will also complain if it was an if/else-if `could use tagged switch on` It's a default lint this time so not my fault for making the lint strict xd

philipch07: I do recall seeing instances of the tagged switch suggestions but I thought those were just suggestions and not requirements. Oh well, no big deal!

JoTurk: We mute them in the CI after we upgraded to v2. because they case conflicts but maybe on the future we'll enable them again https://github.com/pion/webrtc/blob/0575dfbe15ecd8e489db8c1e67f6bdfc953e29be/.golangci.yml#L84C12-L84C18

JoeTurki:
@adeithe Thank you so much for fixing this, looking forward for your incoming contributions :)

PR #822, opened 2025-10-15
Add automatic renomination

kevmo314:
#### Description

Adds automatic renomination as an ICE agent option which allows the ICE agent to switch to the best available link. The logic for this was scoped out of libwebrtc, which also has a 3s polling interval and continuously pings candidates.

The example is probably the best way to see how this feature works. It establishes two virtual links and then you can see it switch between them based on the RTT, which is controlled by the user.

#### Reference issue
Fixes #...

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -526,13 +526,14 @@ func (a *Agent) setSelectedPair(pair *CandidatePair) {
 	a.selectedPair.Store(pair)
 	a.log.Tracef("Set selected candidate pair: %s", pair)
 
+	// Signal connected: notify any Connect() calls waiting on onConnected
+	a.onConnectedOnce.Do(func() { close(a.onConnected) })
+
+	// Update connection state to Connected and notify state change handlers
 	a.updateConnectionState(ConnectionStateConnected)
 
-	// Notify when the selected pair changes
+	// Notify when the selected candidate pair changes
 	a.selectedCandidatePairNotifier.EnqueueSelectedCandidatePair(pair)
-
-	// Signal connected
-	a.onConnectedOnce.Do(func() { close(a.onConnected) })
 }
 
 func (a *Agent) pingAllCandidates() {


agent_config.go
```diff
@@ -211,6 +211,18 @@ type AgentConfig struct {
 	// switched to that irrespective of relative priority between current selected pair
 	// and priority of the pair being switched to.
 	EnableUseCandidateCheckPriority bool
+
+	// AutomaticRenomination enables automatic renomination of candidate pairs when better pairs
+	// become available after the initial connection. This feature requires both agents to support
+	// renomination (draft-thatcher-ice-renomination-01).
```
nils-ohlmeier: I think a warning here might be helpful: this behavior or ICE extension never got fully standardized and thus I think so far Chrome is the only ICE implementation supporting this behavior.

kevmo314: Good point, I added a note to the comment.

agent_config.go
```diff
@@ -211,6 +211,22 @@ type AgentConfig struct {
...
+	//
+	// Note: libwebrtc is the only other ICE implementation that supports renomination. If you are
+	// connecting to a peer that does not support renomination, this will add some load to your
+	// setup but will have no effect on connectivity.
+	AutomaticRenomination bool
+
+	// RenominationInterval is the minimum time to wait after connection before considering
+	// automatic renomination. This prevents excessive renominations during initial connection setup.
+	// If nil, defaults to 3 seconds (matching libwebrtc behavior).
+	RenominationInterval *time.Duration
```
JoTurk: Since we're dropping this config in the next major release, maybe let's stick with the options. If users want the new feature, they can move to the new API. 
What do you think?

examples/automatic-renomination/README.md
```diff
@@ -0,0 +1,556 @@
...
+Automatic renomination is a feature where the controlling ICE agent continuously monitors candidate pairs and automatically switches to better pairs when they become available. This is particularly useful for:
+
+- **Adapting to network changes**: When network conditions improve or degrade
+- **Optimizing for latency**: Automatically switching to lower-latency paths
+- **Quality of service**: Maintaining the best possible connection quality
+- **Interface failover**: Switching to alternate interfaces when primary path fails
+
+## How It Works
+
+The automatic renomination feature evaluates candidate pairs based on:
```
JoTurk, line 16: Will be cool if we provide a docker image with this example, I made a scuffed one while i was testing, can clean and commit it :)

agent_config.go
kevmo314: Makes sense to me! Removed.

examples/automatic-renomination/README.md
kevmo314: Yeah that would be helpful! Can you push yours?

Definitely the hardest part of testing this is the networking setup.

JoTurk: > Yeah that would be helpful! Can you push yours?

cool, I'll make a PR.

PR #831, opened 2025-11-08
Add port mapping to address rewrite rules

xinze-zheng:
## Summary

- Extend AddressRewriteRule with OriginalPort and NewPort so address and port translation use one configuration API.
- Select port mappings with the rule's existing candidate type, local address, interface, CIDR, network, and precedence filters.
- Match one gathered port when OriginalPort is non-zero, or every gathered port when it is zero. NewPort must be non-zero unless both fields are zero, which disables port rewriting.
- Apply port rewriting to gathered host, server-reflexive, and relay candidates across direct, UDP mux, universal UDP mux, STUN, TURN, and address-rewrite paths.
- Change only the advertised candidate port; the underlying packet connection remains bound to its original port.

This replaces the earlier callback API, so the PR adds no handler or agent state.

## Tests

- Full Go test suite
- golangci-lint
- Pion shared commit-message linter
- Public AddressRewriteRules validation plus wildcard and exact port gathering coverage

## Reference

https://github.com/pion/webrtc/issues/3155


agent_config.go
```diff
@@ -211,6 +211,9 @@ type AgentConfig struct {
 	// switched to that irrespective of relative priority between current selected pair
 	// and priority of the pair being switched to.
 	EnableUseCandidateCheckPriority bool
+
+	// MapPortHanlder is the handler used to compute mapped port for host candidate.
+	MapPortHanlder func(candidate Candidate) int
```
JoTurk: We've stopped adding new configurations to the AgentConfig struct since it's planned for deprecation in favor of a more Pion-aligned API. this feature should instead be exposed as an option, WithMapPortHandler, and used with the new NewAgentWithConfig API. Users who want to use this  should migrate to the new API.

we did the same thing recently with the renomination API: https://github.com/pion/ice/pull/822#discussion_r2463046725

JoTurk (commented):
Thank you so much, I like this direction let me know when it's ready for test and review.

candidate.go
```diff
@@ -46,6 +46,10 @@ type Candidate interface {
 	Address() string
 	Port() int
 
+	// Port mapping support for containers
+	MappedPort() int
```
JoTurk: question, do we have to make this public?

xinze-zheng: Should we? Will webrtc care about this? I was thinking for design of separted ICE (maybe the future ion) this will be somehow used. What do you think?

JoTurk: I think we should keep it private for now, unless we need it, or someone requests that we expose it later.

xinze-zheng:
@JoTurk Hi I've been looking into integrating this feature with webrtc. Seems we need to replace `webrtc.ICECandidate.Port` with `mappedPort` in [webrtc.newICECandidateFromICE](https://github.com/pion/webrtc/blob/d2a251e9384b2bde967ee8bc1bdc9a8941cf1fd9/icecandidate.go#L49) (or the information of mappedPort is lost). I think we can implement via 
```
	port := candidate.Port()
	if candidate.GetMappedPort() != 0 {
		port = candidate.GetMappedPort()
	}
```
we just need to make GetMappedPort public in `ice.candidate`.

agent_options.go
```diff
@@ -960,3 +960,17 @@ func WithLoggerFactory(loggerFactory logging.LoggerFactory) AgentOption {
 		return nil
 	}
 }
+
+func WithMapPortHandler(handler func(cand Candidate) int, candTyp CandidateType) AgentOption {
+	return func(a *Agent) error {
+		a.mapPort = func(candidate Candidate) int {
+			if candidate.Type() == candTyp {
```
JoTurk: I think we'll either need to remove the candidate type filter or make it possible to add multiple filters for different candidates types.
I think removing is better because the user filter have access to the candidate type.

xinze-zheng: Agree! Updated.

Sean-Der (approved):

PR #832, opened 2025-11-09
Finish agent options

JoTurk:
#### Description
slightly changes the behavior for the agents options introduced in https://github.com/pion/ice/pull/799/ so that
1. Called before any I/O begins, allowing the entire config object to be replaced.
2. config.initDefault no longer overwrites it.
3. Moved all config state into the agent struct to allow overriding via options.
4. Added new tests.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -111,8 +111,10 @@ type Agent struct {
 
 	selectedPair atomic.Value // *CandidatePair
 
-	urls         []*stun.URI
-	networkTypes []NetworkType
+	urls             []*stun.URI
+	networkTypes     []NetworkType
+	natCandidateType CandidateType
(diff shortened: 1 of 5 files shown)


agent.go
```diff
@@ -371,22 +419,13 @@ func newAgentWithConfig(config *AgentConfig, opts ...AgentOption) (*Agent, error
...
 	}
 
-	for _, opt := range opts {
-		if err := opt(agent); err != nil {
-			agent.closeMulticastConn()
-			_ = agent.Close()
-
-			return nil, err
-		}
-	}
```
philipch07: Do we not need to handle this anymore, or does it automatically get handled somewhere else?

Issue #780, opened 2025-05-31
support domain name besides IP  sent to the client in ICE candidates


ghost:
## Summary

support domain name besides IP  sent to the client in ICE candidates 

## Motivation

This would allow specify the external domain sent to the client's browser, and the ip should be resolved on the client side to increase flexibility

## Describe alternatives you've considered

A clear and concise description of the alternative solutions you've considered.

## Additional context

https://github.com/pion/ice/blob/753c2a0fe868006a9e8d341feb11fc6a5d54a45d/external_ip_mapper.go#L12

fippo:
note that this was removed from [RFC 8839](https://datatracker.ietf.org/doc/html/rfc8839#section-5.1) (mdns is allowed by an extension spec). Lots of complexity...

aler9:
I used domain names as SDP candidates for a while by mungling SDP and manually adding them. I stopped doing that as soon as i discovered that Chrome is the only browser that supports domain names in the SDP. In all others, domain names are reserved for mdns. This is quite clear if you take a look at Firefox's source code. Browsers will never be updated to support this feature, especially if it has been removed from the RFC.

Just solve domains on server-side and send resolved IPs to users.

JoTurk:
I think we shouldn't have an issue with merging a support for this if someone has a use for them and want to send a patch, pion/ICE is used a lot outside of the webrtc world.

PR #834, opened 2025-11-10
Add AddressRewriteRules

JoTurk:
#### Description

This deprecates NAT 1:1 API.

We're currently reworking the Agent API to use an option pattern API this is the last option before we can use it with WebRTC, I kept this last so I can take this opportunity to add better mapping and settings.

For me I have a test server with 11 mixed ipv4 and ipv6 interfaces some of them behind NAT 1:1 and it was always annoying that i can only use nat1:1 on a single interface. and filter the rest.

But this is also needed for ION smart ICE thing.

NAT 1:1 mapping is configured via **rules**. You can add rules with helpers for the common cases (host/srflx), or provide full custom rules.

Also I added a docker compose example with multiple rules and network examples.

#### TL;DR

* New option: `WithAddressRewriteRules(rules ...AddressRewriteRule) AgentOption`. 
* Rules cover **host**, **server-reflexive**, and **relay** candidates.
* Each rule can be scoped by:
  * local address (`Local`)
  * interface name (`Iface`)
  * CIDR (`CIDR`)
  * network types (`Networks`)
* Each rule chooses whether to **replace** the original candidate or **append** additional ones via `AddressRewriteMode` (`AddressRewriteReplace` / `AddressRewriteAppend`), with sensible defaults per candidate type. 
* NAT 1:1 error identifiers still exist but are now marked deprecated; new address-rewrite logic uses them internally for backwards compatibility. NAT 1:1 will be deprecated with the whole config struct after this PR.
* Empty External rules are intentional:
     - Mode AddressRewriteReplace drops the matched candidate (deny-list style).
     - Mode AddressRewriteAppend keeps the original candidate and adds nothing,
     - which is useful when you combine a catch-all replace with per-interface allow rules.
---

#### New API: Address rewrite rules

```go
// WithAddressRewriteRules appends the provided address rewrite (1:1) rules
// to the Agent configuration.
func WithAddressRewriteRules(rules ...AddressRewriteRule) AgentOption
```

```go
type AddressRewriteMode int

const (
    AddressRewriteReplace AddressRewriteMode = iota + 1
    AddressRewriteAppend
)
```

```go
// AddressRewriteRule represents a rule for remapping candidate addresses.
type AddressRewriteRule struct {
	// External are the 1:1 external addresses to advertise for this rule.
	// For replace mode, an empty list is treated as "drop the matched local
	// address" (no candidate emitted). For append mode, an empty list is a
	// no-op: the original candidate is kept.
	// Empty External rules are intentional:
	//   - Mode AddressRewriteReplace drops the matched candidate (deny-list style).
	//   - Mode AddressRewriteAppend keeps the original candidate and adds nothing,
	//     which is useful when you combine a catch-all replace with per-interface
	//     allow rules.
	External []string
	// Local optionally pins this rule to a specific local address. When set,
	// external IPs map to that address regardless of IP family. When empty,
	// External acts as a catch-all for the family implied by the local scope
	// (CIDR when set, otherwise the external IP family).
	Local string
	// Iface is the optional interface name to limit the rule to, empty = any.
	Iface string
	// CIDR is the optional CIDR to limit the rule to, empty = any.
	CIDR string
	// AsCandidateType is the candidate type to publish as for this rule. Defaults to host
	// when unspecified. Supported values: host, server reflexive, relay.
	AsCandidateType CandidateType
	// Mode controls whether we replace the original candidate or append extra
	// candidates.
	//
	// If Mode is zero, the default is:
	//   - CandidateTypeHost           -> AddressRewriteReplace
	//   - CandidateTypeServerReflexive, CandidateTypeRelay -> AddressRewriteAppend
	// For replace mode, a match with zero external IPs removes the candidate.
	// For append mode, a match with zero external IPs leaves the original
	// candidate untouched.
	Mode AddressRewriteMode
	// Networks is the optional networks to limit the rule to, nil/empty = all.
	Networks []NetworkType
}
```

Implementation notes:

* Rules are pre-validated:

  * `External` must be non-empty, deduplicated, and all valid IP literals.
  * `Local` (when set) must be a valid IP and consistent with `External` and `CIDR`.
  * `AsCandidateType` must not be peer-reflexive.
* Internally, rules are compiled into per-family mappings with fast lookup for:

  * exact `Local` hits
  * per-family "catch-all" mappings (IPv4 / IPv6).

---

#### Behavior / semantics

* **Rule ordering & precedence**

  * Rules are applied in the order they are added (first match wins for a given candidate type + local address).
  * Within a rule, mappings prefer:

    * explicit `Local` bindings
    * then CIDR-scoped bindings
    * then per-family catch-all external IPs. 
* **Candidate types**

  * `AsCandidateType = CandidateTypeHost`

    * Default mode: **replace** – host candidates are rewritten to the configured external IPs.
  * `AsCandidateType = CandidateTypeServerReflexive`

    * Default mode: **append** – server-reflexive candidates are synthesized directly from the rules, so you can advertise srflx addresses even without STUN.
  * `AsCandidateType = CandidateTypeRelay`

    * Default mode: **append** – allows rewriting / normalizing relay addresses for deployments where the TURN server is behind another hop. 

* **Replace vs append**

  * `AddressRewriteReplace`: only the rewritten candidates are advertised for the matching scope.
  * `AddressRewriteAppend`: rewritten candidates are added alongside the originals.
  * Mixed behavior is possible by combining multiple rules with different `Mode`s and scopes.

* **Validation & diagnostics**

  * Invalid combinations (family mismatches between `External`, `Local` and `CIDR`, or unsupported candidate types) fail fast with the existing NAT-1:1 error values (now deprecated but still emitted).
  * Overlapping rules for the same "scope" (candidate type + iface + CIDR + networks + local family) are detected and logged as warnings with a summary of the conflict.

---

#### examples

*host, replace:*

```go
WithAddressRewriteRules(
    ice.AddressRewriteRule{
        External:        []string{"203.0.113.10"},
        AsCandidateType: ice.CandidateTypeHost,
        // Mode omitted -> defaults to AddressRewriteReplace for host
    },
)
```

*srflx, append:*

```go
WithAddressRewriteRules(
    ice.AddressRewriteRule{
        External:        []string{"198.51.100.50"},
        AsCandidateType: ice.CandidateTypeServerReflexive,
        // Mode omitted -> defaults to AddressRewriteAppend for srflx
    },
)
```

*per-iface rule:*

```go
WithAddressRewriteRules(
    ice.AddressRewriteRule{
        External:        []string{"203.0.113.10"},
        AsCandidateType: ice.CandidateTypeHost,
        Iface:           "eth0",
        CIDR:            "10.0.0.0/24",
        Networks:        []ice.NetworkType{ice.NetworkTypeUDP4},
    },
)
```

Semantics around mDNS safety (rejecting host rewrites when mDNS is `QueryAndGather`) and srflx/relay enablement remain consistent with the earlier design.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1022,6 +1022,10 @@ func (a *Agent) addRemoteCandidate(cand Candidate) { //nolint:cyclop
 }
 
 func (a *Agent) addCandidate(ctx context.Context, cand Candidate, candidateConn net.PacketConn) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
 	return a.loop.Run(ctx, func(context.Context) {
 		set := a.localCandidates[cand.NetworkType()]
 		for _, candidate := range set {


agent_options.go
```diff
@@ -31,6 +33,300 @@ func DefaultNominationValueGenerator() NominationValueGenerator {
...
+	if len(publicIPs) == 0 {
+		return nil
+	}
+
+	entries := make([]nat1To1PublicIPEntry, 0, len(publicIPs))
+
+	for _, mapping := range publicIPs {
+		if mapping == "" {
+			continue
+		}
```
philipch07: Should we trim this before checking if it's empty?

agent_options_test.go
```diff
@@ -647,6 +647,349 @@ func TestWithCandidateTypesNAT1To1Validation(t *testing.T) {
...
+
+		//nolint:godox
+		// TODO: remove once AgentConfig.NAT1To1IPs is deprecated.
+		agent, err := NewAgent(&AgentConfig{
+			Net:        stub,
+			NAT1To1IPs: []string{"1.2.3.4", "5.6.7.8"},
+		})
+		require.ErrorIs(t, err, ErrInvalidNAT1To1IPMapping)
+		require.Nil(t, agent)
+	})
```
philipch07, line 1512: When are we deprecating `AgentConfig.Nat1To1IPs`?

JoTurk: when we remove AgentConfig in the next major release.

JoTurk: After we're done, I'll mark it as deprecated, we already did that with the new APIs.

agent_options.go
JoTurk: No because we'll end up trimming twice after we remove the part later, so the behavior won't change but we'll just trim twice.

JoeTurki:
@Sean-Der did you get time to find old threads? I can try finding them too? was it in the slack? I changed the gatherer so it actually works with the new rules was going to do it in a PR, also added a docker compose with some networks inspired by this issue @nab-os issue https://github.com/Glimesh/broadcast-box/issues/447

gather.go
```diff
@@ -171,156 +171,175 @@ func (a *Agent) gatherCandidatesLocal(ctx context.Context, networkTypes []Networ
...
+						}
+					}
+					if len(conns) == 0 {
+						// Didn't succeed with any, try the next network.
+						continue
+					}
+					tcpType = TCPTypePassive
+					// Is there a way to verify that the listen address is even
+					// accessible from the current interface.
+				case udp:
```
philipch07, line 354: Is the second comment correct? I'm not sure what it's describing.

JoTurk: It's more of a 5 years old todo, I didn't want to remove it while restructuring the file here when it was added in 2020 https://github.com/pion/ice/commit/3b9a98917766765015c8d34db37db50c80760ef6

JoeTurki:
> If you do hostname lets us resolve https://github.com/pion/ice/issues/780 also!

I looked at this issue many times, I think if we support this, I think we should also allow for custom resolves / lookup :) 

> Do you think it would be worth allowing a user to also do a relay candidate?

If we're going with explicit replace or append, I think it will make sense to allow any candidate, I'm not sure how it will make sense for prflx but I think we should just allow all candidates, for consistency :)

agent.go
```diff
@@ -111,18 +111,17 @@ type Agent struct {
 
 	selectedPair atomic.Value // *CandidatePair
 
-	urls             []*stun.URI
-	networkTypes     []NetworkType
-	natCandidateType CandidateType
-	natIPs           []string
+	urls                []*stun.URI
+	networkTypes        []NetworkType
```
philipch07: Similarly I wonder if this needs to be a list or if a map may be more appropriate (e.g. if we're primarily checking for existence of a specific network type).

agent_options.go
```diff
@@ -31,6 +33,300 @@ func DefaultNominationValueGenerator() NominationValueGenerator {
...
+	ip, _, err := validateIPString(local)
+	if err != nil {
+		return "family:unknown"
+	}
+
+	if ip.To4() != nil {
+		return "family:ipv4"
+	}
+
+	return "family:ipv6"
```
philipch07: Should this just call `return deriveAddressRewriteFamilyScopeKey(local)`?

JoeTurki:
merging this tonight :)

Sean-Der:
This is so great. I am so grateful you are pushing on all this cruft.

It is such an uphill battle to change existing stuff. Really special work you are doing. This is MUCH harder then writing new code :)

Sean-Der (approved):

gather.go
```diff
@@ -469,48 +539,98 @@ func (a *Agent) gatherCandidatesSrflxMapped(ctx context.Context, networkTypes []
...
+				if shouldFilterLocationTracked(mappedIP) {
+					closeConnAndLog(currentConn, a.log, "external IP is somehow filtered for location tracking reasons %s", mappedIP)
+
+					if idx > 0 {
+						continue
+					}
+
+					// The first mapped IP reuses the original listen socket, stop the loop to avoid
+					// reusing a closed base connection for subsequent mappings.
+					return
```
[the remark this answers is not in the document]
JoTurk: @copilot I refactored the code hopefully it prevents you from talking about this non existent issue

gather.go
```diff
@@ -469,48 +539,92 @@ func (a *Agent) gatherCandidatesSrflxMapped(ctx context.Context, networkTypes []
...
+					newConn, listenErr := listenUDPInPortRange(
+						a.net,
+						a.log,
+						int(a.portMax),
+						int(a.portMin),
+						network,
+						&net.UDPAddr{IP: lAddr.IP, Port: 0},
+					)
+					if listenErr != nil {
+						closeConnAndLog(newConn, a.log, "Failed to listen %s for additional srflx mapping: %v", network, listenErr)
```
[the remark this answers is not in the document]
JoTurk: All the comments are wrong again.
and this concern is not an actual leak. By the time idx > 0 is reached, the base conn from the first mapping has already been handed to the first candidate and must stay open. If listenUDPInPortRange fails for a subsequent mapping, we log and return, there's nothing to close because the newConn wasn't acquired, and the original conn is still in use by the already-added candidate. So we shouldn't close the base socket in that error path.

PR #837, opened 2025-11-13
ECN support

penhauer:
This part is the tricky part for supporting ECN. If the underlying socket is a UDPConn,  `ReadMsgUDP(buf, oob)` is used which returns out of band data like ToS. It works fine for now, but I honestly haven't thought about the cases where the `conn` is not a pure socket, but a class that implements `net.PacketConn` (like a Proxy or a Wrapper), and in this case we don't want to bypass the Wrapper's methods. However, for now, I think this is fine.


JoTurk (approved):
Hello, I finished this and integrated it with the new transport attributes API, I put ECN behind a "unix" flag (linux || darwin || freebsd), I'm not sure how to do it for windows and I don't have a real way to test it on windows right now (I can do that later).

Sean-Der (approved):
IIITTTTSSS HAPPENING! Amazing thank you @JoTurk

PR #840, opened 2025-11-26
Fix a deadlock in TaskLoop

JoTurk:
#### Description
Fixes a rare race condition in taskloop if close fired just while Run was getting called in the tiny window after the err check and just before Run gets called, and because the context is actually set to the loop itself we get a deadlock.

#### Reference issue
I saw it happening this week and I think it Fixes https://github.com/pion/interceptor/issues/384 but not 100% sure

### DIFF
--- a/internal/taskloop/taskloop_test.go
+++ b/internal/taskloop/taskloop_test.go
@@ -1,0 +1,60 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package taskloop
+
+import (
+	"context"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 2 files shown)


internal/taskloop/taskloop_test.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRunReturnsErrClosedWhenLoopClosing(t *testing.T) {
```
philipch07, line 15: Is there a way to check for goroutine leaks in this test? Or maybe in some sort of fuzz test?

JoTurk: We can't really fuzz test this, this test emulates the condition where Run and close are both called at the same time in the small window in the goroutine, We don't close the context on close because it's actually tied to the loop, so yeah the deadlock happens... you can see it in the CI logs.

Sean-Der (approved):

PR #841, opened 2025-11-30
Set linger time for tcp connections

boks1971:
#### Description

It is possible that Close() of a TCP connection to block if there is pending data to be written till it is written out. Set linger to ensure that it does not block.

### DIFF
--- a/tcp_packet_conn.go
+++ b/tcp_packet_conn.go
@@ -284,6 +284,11 @@ func (t *tcpPacketConn) removeConn(conn net.Conn) bool {
 
 	t.closeAndLogError(conn)
 
+	// wait for some time to flush pending writes
+	_ = conn.SetWriteDeadline(time.Now().Add(5 * time.Second))
+	// read deadline as well just in case
+	_ = conn.SetReadDeadline(time.Now().Add(5 * time.Second))
+
 	delete(t.conns, conn.RemoteAddr().String())
 
 	return len(t.conns) == 0
@@ -303,6 +308,12 @@ func (t *tcpPacketConn) Close() error {
 
 	for _, conn := range t.conns {
 		t.closeAndLogError(conn)
+
+		// wait for some time to flush pending writes
+		_ = conn.SetWriteDeadline(time.Now().Add(5 * time.Second))
+		// read deadline as well just in case
+		_ = conn.SetReadDeadline(time.Now().Add(5 * time.Second))
+
 		delete(t.conns, conn.RemoteAddr().String())
 	}
 


tcp_mux_test.go
```diff
@@ -216,7 +217,13 @@ func TestTCPMux_NoLeakForConnectionFromStun(t *testing.T) {
...
 		buf := make([]byte, 1)
 		_, err = conn.Read(buf)
-		require.ErrorIs(t, err, io.EOF)
+		if opErr, ok := err.(*net.OpError); ok { //nolint:errorlint
+			if sysErr, ok := opErr.Err.(syscall.Errno); ok { //nolint:errorlint
+				require.Equal(t, syscall.ECONNRESET, sysErr)
+			}
+		} else {
+			require.ErrorIs(t, err, io.EOF)
+		}
```
boks1971: had to change this expected error because of use of `SetLinger`.

Maybe should use `SetWriteDeadline` (or `SetDeadline`) to unblock?

tcp_packet_conn.go
```diff
@@ -152,6 +152,18 @@ func (t *tcpPacketConn) AddConn(conn net.Conn, firstPacketData []byte) error {
...
 	}
 
+	// ensure that close does not block on pending write data.
+	//
+	// two things of interest:
+	//   1. Close() may or may not block on pending writes - see https://pkg.go.dev/net#Conn
+	//   2. using SetLinger(0) as some OSes can block even if linger time is > 0,
+	//      see https://pkg.go.dev/net#TCPConn.SetLinger
+	if tcpConn, ok := conn.(*net.TCPConn); ok {
+		if err := tcpConn.SetLinger(0); err != nil {
```
JoTurk: But by default it doesn't block because:

> If sec < 0 (the default), the operating system finishes sending the data in the background.

And this change will discard any unsent or ack in the pipes, but it will not change the block behavior (currently will not block), 

> If sec == 0, the operating system discards any unsent or unacknowledged data.

It will just cause more unexpected close at the other peer instead of graceful shutdown, and as for writeDeadline go already sets a write deadline of 5s, we can set it to a smaller deadline, but I'm not sure about this change, because it doesn't directly change how the block happens, and it's kinda of an extreme change to drop graceful shutdown everywhere.

boks1971: Got it. Yeah, this is not great.

Where can I see the 5 second default for write? Yeah, I was thinking of using WriteDeadline.

Trying to reason for the hang on close and this is the only thing I could find. This is happening in a similar environment which showed the SCTP issue in https://github.com/pion/sctp/pull/441.

JoTurk: https://github.com/golang/go/blob/8fd0f83552d3ef9ca38c031bec93a36b189e3e11/src/crypto/tls/conn.go#L1360-L1361

they set writeDeadline but not readDeadline which caused the issue :smile: so we were stuck on read. but also the default writeDeadline of 5s is kinda big.

boks1971: Thank you @JoeTurki . I guess that is not the cause of the issue then. 5 seconds is big, but it should at least not get stuck. May not be worth making a change as it is unlikely to solve the issue. Need to look more.

boks1971: Oh, that 5s is for TLS conn it looks like.

I think it is worth trying setting a write deadline of 5 seconds for TCP conn.

Will make the change and set it to 5s although it is big to give enough room.

JoTurk: makes sense, thank you.

boks1971: Updated it Joe.

JoeTurki (approved):
Thank you, maybe we should add ReadDeadline too?

PR #842, opened 2025-12-03
Cancel gathering before removing candidates

boks1971:
#### Description

When ICE agent is closed, the connections from UDP/TCP mux were removed before gathering was canceled. This led to candidate gathering adding entries to UDP mux which were not cleaned up.

~Also passing into logger to the mux as ICE agent logger could have more context and is useful to see the context.~


agent.go
```diff
@@ -401,6 +401,11 @@ func newAgentWithConfig(agent *Agent, opts ...AgentOption) (*Agent, error) {
 	}
 
 	agent.loop = taskloop.New(func() {
+		agent.gatherCandidateCancel()
```
boks1971, line 404: This is the main change. Other bits are passing around ICE agent logger into TCP mux. The ICE agent logger could have more context and could be useful.

Guess, I should add it to UDP mux also?

boks1971: Oh well, this breaks API checks. Will probably remove it then.

boks1971: okay, cleaned them up. This is the only change now.

PR #851, opened 2025-12-17
Store lastSent and lastReceived as int64

atoppi:
#### Description

This patch will store candidate `lastSent` and `lastReceived` in `int64` in place of `time`. The interface will not be changed, since we just convert to `time` when a reading is requested.
In order to preserve monotonic clock, we save a referenceTime and use it:
- to store nanoseconds since the ref time when writing
- build a monotonic clock from nanoseconds when reading

#### Reference issue
While memory profiling our video SFU, we noticed that on high loads the memory allocations performed by `setLastSent` start to impact. This patch aims at reducing allocations in ice.
pion-ice-allocs

#### Benchmark

```
                                 │  master.txt  │          with-ref-nanos.txt          │
                                 │    sec/op    │   sec/op     vs base                 │
CandidateLastSent/setLastSent-24   17.640n ± 5%   4.996n ± 1%   -71.68% (p=0.000 n=10)
CandidateLastSent/LastSent-24       1.169n ± 1%   3.557n ± 4%  +204.45% (p=0.000 n=10)
geomean                             4.540n        4.216n         -7.14%

                                 │  master.txt  │           with-ref-nanos.txt            │
                                 │     B/op     │    B/op     vs base                     │
CandidateLastSent/setLastSent-24   24.00 ± 0%      0.00 ± 0%  -100.00% (p=0.000 n=10)
CandidateLastSent/LastSent-24      0.000 ± 0%     0.000 ± 0%         ~ (p=1.000 n=10) ¹
geomean                                       ²               ?                       ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean

                                 │  master.txt  │           with-ref-nanos.txt            │
                                 │  allocs/op   │ allocs/op   vs base                     │
CandidateLastSent/setLastSent-24   1.000 ± 0%     0.000 ± 0%  -100.00% (p=0.000 n=10)
CandidateLastSent/LastSent-24      0.000 ± 0%     0.000 ± 0%         ~ (p=1.000 n=10) ¹
geomean                                       ²               ?                       ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean
```

### DIFF
--- a/candidate_base.go
+++ b/candidate_base.go
@@ -32,8 +32,8 @@ type candidateBase struct {
 
 	resolvedAddr net.Addr
 
-	lastSent     atomic.Value
-	lastReceived atomic.Value
+	lastSent     atomic.Int64
+	lastReceived atomic.Int64
 	conn         net.PacketConn
 
 	currAgent *Agent
@@ -454,29 +454,29 @@ func (c *candidateBase) String() string {
 // LastReceived returns a time.Time indicating the last time
 // this candidate was received.
 func (c *candidateBase) LastReceived() time.Time {
-	if lastReceived, ok := c.lastReceived.Load().(time.Time); ok {
-		return lastReceived
+	if lastReceived := c.lastReceived.Load(); lastReceived != 0 {
+		return time.Unix(0, lastReceived)
 	}
 
 	return time.Time{}
 }
 
 func (c *candidateBase) setLastReceived(t time.Time) {
-	c.lastReceived.Store(t)
+	c.lastReceived.Store(t.UnixNano())
 }
 
 // LastSent returns a time.Time indicating the last time
 // this candidate was sent.
 func (c *candidateBase) LastSent() time.Time {
-	if lastSent, ok := c.lastSent.Load().(time.Time); ok {
-		return lastSent
+	if lastSent := c.lastSent.Load(); lastSent != 0 {
+		return time.Unix(0, lastSent)
 	}
 
 	return time.Time{}
 }
 
 func (c *candidateBase) setLastSent(t time.Time) {
-	c.lastSent.Store(t)
+	c.lastSent.Store(t.UnixNano())
 }
 
 func (c *candidateBase) seen(outbound bool) {
--- a/candidate_base.go
+++ b/candidate_base.go
@@ -48,6 +48,19 @@ type candidateBase struct {
 	extensions            []CandidateExtension
 }
 
+// Save a time reference to calculate monotonic time for candidate last sent/received
+var timeRef = time.Now()
+
+// getMonoNanos returns the monotonic nanoseconds of a time t since timeRef
+func getMonoNanos(t time.Time) int64 {
+	return t.Sub(timeRef).Nanoseconds()
+}
+
+// getMonoTime returns a time.Time based on monotonic nanos since timeRef
+func getMonoTime(nanos int64) time.Time {
+	return timeRef.Add(time.Duration(nanos))
+}
+
 // Done implements context.Context.
 func (c *candidateBase) Done() <-chan struct{} {
 	return c.closeCh
@@ -455,28 +468,28 @@ func (c *candidateBase) String() string {
 // this candidate was received.
 func (c *candidateBase) LastReceived() time.Time {
 	if lastReceived := c.lastReceived.Load(); lastReceived != 0 {
-		return time.Unix(0, lastReceived)
+		return getMonoTime(lastReceived)
 	}
 
 	return time.Time{}
 }
 
 func (c *candidateBase) setLastReceived(t time.Time) {
-	c.lastReceived.Store(t.UnixNano())
+	c.lastReceived.Store(getMonoNanos(t))
 }
 
 // LastSent returns a time.Time indicating the last time
 // this candidate was sent.
 func (c *candidateBase) LastSent() time.Time {
 	if lastSent := c.lastSent.Load(); lastSent != 0 {
-		return time.Unix(0, lastSent)
+		return getMonoTime(lastSent)
 	}
 
 	return time.Time{}
 }
 
 func (c *candidateBase) setLastSent(t time.Time) {
-	c.lastSent.Store(t.UnixNano())
+	c.lastSent.Store(getMonoNanos(t))
 }
 
 func (c *candidateBase) seen(outbound bool) {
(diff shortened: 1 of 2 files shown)


candidate_base.go
```diff
@@ -454,29 +454,29 @@ func (c *candidateBase) String() string {
...
+		return time.Unix(0, lastReceived)
 	}
 
 	return time.Time{}
 }
 
 func (c *candidateBase) setLastReceived(t time.Time) {
-	c.lastReceived.Store(t)
+	c.lastReceived.Store(t.UnixNano())
 }
```
JoTurk, line 480: Good catch. but this changes the behavior, we'll drop the monotonic clock, Which will cause some subtle bugs in rare edge cases.

maybe we can store one Time object, and use it to construct and compare monotonic time.

atoppi: You are right, we are losing the monotonic time. Any idea on how to proceed?

JoTurk: We can store a single time.Time for truth, and use `time.Since(base).Nanoseconds()` to capture change, and compare change against that time. The only nuanced part is when we return time in LastReceived but we can do best effort by using `time.Now().Add(-elapsed)` which will still be "correct" and monotonic-safe I think.

atoppi: Isn't the the storing of the first time reference + nanos racy?

JoTurk: I meant a single time object per application, initialized once and never mutated.

atoppi: Done. We are losing something on the read path, but writing is blazing fast and allocations are 0.

atoppi: @JoeTurki the linter is complaining for the global variable. What is the pattern to adopt here?

Sean-Der: @atoppi ` // nolint: gochecknoglobals` is totally fine!

Sean-Der: Pretty cool that you are profiling/debugging :) if you have any other ideas/complaints I would love to hear :)

JoTurk: I think you also removed the test correctness, from this version.

atoppi: Tests should be fine, with nanos elapsed since ref time the previous test is working.

candidate_base.go
```diff
@@ -48,6 +48,20 @@ type candidateBase struct {
...
+
+// getMonoNanos returns the monotonic nanoseconds of a time t since timeRef.
+func getMonoNanos(t time.Time) int64 {
+	return t.Sub(timeRef).Nanoseconds()
+}
+
+// getMonoTime returns a time.Time based on monotonic nanos since timeRef.
+func getMonoTime(nanos int64) time.Time {
+	return timeRef.Add(time.Duration(nanos))
+}
```
JoTurk, line 63: test is failing because we treat 0 as not set, simple way to "fix" this.
 
```suggestion
func getMonoNanos(t time.Time) int64 {
	return t.Sub(timeRef).Nanoseconds() + 1
}

// getMonoTime returns a time.Time based on monotonic nanos since timeRef.
func getMonoTime(nanos int64) time.Time {
	return timeRef.Add(time.Duration(nanos - 1))
}
```

candidate_base.go
JoTurk: Oh i see, I was looking at the old diff by mistake.

candidate_base.go
JoTurk: We shouldn't do this tho, I was just explaining the issue.

atoppi: No, the issue is that the test is flaky. `TestCandidateLastSent` and `TestCandidateLastReceived` might fail for a little nano seconds diff.

atoppi:
```
=== RUN   TestCandidateLastReceived
    /home/atoppi/src/pion/ice/candidate_test.go:196:
                Error Trace:    /home/atoppi/src/pion/ice/candidate_test.go:196
                Error:          Not equal:
                                expected: time.Date(2025, time.December, 17, 17, 41, 59, 709426140, time.Local)
                                actual  : time.Date(2025, time.December, 17, 17, 41, 59, 709426131, time.Local)

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,2 +1,2 @@
                                -(time.Time) 2025-12-17 17:41:59.70942614 +0100 CET m=+0.000883993
                                +(time.Time) 2025-12-17 17:41:59.709426131 +0100 CET m=+0.000883993

                Test:           TestCandidateLastReceived
--- FAIL: TestCandidateLastReceived (0.00s)
FAIL
FAIL    github.com/pion/ice/v4  0.004s
```

JoTurk: I see, i thought i hit getMonoTime() being 0, while testing something locally. we can make the test non flaky but we'll have to add a test to make sure it's monotonic, maybe outside of this scope tho.

atoppi:
```
=== RUN   TestCandidateLastReceived
    /home/atoppi/src/pion/ice/candidate_test.go:196:
                Error Trace:    /home/atoppi/src/pion/ice/candidate_test.go:196
                Error:          Not equal:
                                expected: time.Date(2025, time.December, 17, 17, 41, 59, 709426140, time.Local)
                                actual  : time.Date(2025, time.December, 17, 17, 41, 59, 709426131, time.Local)

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,2 +1,2 @@
                                -(time.Time) 2025-12-17 17:41:59.70942614 +0100 CET m=+0.000883993
                                +(time.Time) 2025-12-17 17:41:59.709426131 +0100 CET m=+0.000883993

                Test:           TestCandidateLastReceived
--- FAIL: TestCandidateLastReceived (0.00s)
FAIL
FAIL    github.com/pion/ice/v4  0.004s
```
Any idea for this randomic tiny error?

JoeTurki:
@atoppi I meant we can after this PR or in the same PR we can add a simple test that calls String and compares the ` m=` value (or something like that), and assert that the monotonic clock is preserved. A test like this would fail with the original version of this PR with unix time.

As for the flaky test, it's fine to use WithinDuration to add some tolerance, since we don't store the exact time object, this is expected.

JoTurk (approved):
Thank you a lot.

atoppi:
> Thank you a lot.

Thank you @JoTurk for everything you're doing for pion and @Sean-Der for this incredible project.

PR #852, opened 2025-12-18
Propagate and implement deadlines

JoTurk:
#### Description
Propagate and implement deadlines, Mainly across the TCP stack.
We found many rare deadlocks and goroutines leaks in some conditions.

### DIFF
--- a/tcp_packet_conn_test.go
+++ b/tcp_packet_conn_test.go
@@ -20,6 +20,37 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type deadlineConn struct {
+	readDeadline  time.Time
+	writeDeadline time.Time
+	deadline      time.Time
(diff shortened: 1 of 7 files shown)


tcp_packet_conn_test.go
```diff
@@ -20,6 +20,37 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type deadlineConn struct {
```
boks1971, line 23: is this type used anywhere?

boks1971 (approved):
Great stuff @JoeTurki . Thank you for doing this :bow:

tcp_packet_conn_test.go
JoTurk: yes it's a mock used in TestTCPPacketConn_SetDeadlines.

Issue #854, opened 2025-12-18
Refactors and code improvements


JoTurk:
Right now in pion/ICE, we kinda have hard time with adding new features, because of code all over the place. duplicated code paths, nolints. and others. 
This is a tracking issue for the fixes and refactors.

PR #855, opened 2025-12-18
Refactor and test handleInbound

JoTurk:
#### Description
Refactor handleInbound and make it more testable. remove nolint.
part of https://github.com/pion/ice/issues/854

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1378,15 +1378,12 @@ func (a *Agent) handleRoleConflict(msg *stun.Message, local, remote Candidate, r
 }
 
 // handleInbound processes STUN traffic from a remote candidate.
-func (a *Agent) handleInbound(msg *stun.Message, local Candidate, remote net.Addr) { //nolint:gocognit,cyclop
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -1378,98 +1378,125 @@ func (a *Agent) handleRoleConflict(msg *stun.Message, local, remote Candidate, r
...
-
-		if remoteCandidate == nil {
-			a.log.Warnf("Discard success message from (%s), no such remote", remote)
-
+	case stun.ClassRequest:
+		var ok bool
+		if remoteCandidate, ok = a.handleInboundRequest(remoteCandidate, local, remote, msg); !ok {
 			return
 		}
+	default:
```
philipch07, line 1404: I think this default is redundant

philipch07 (approved):
a lot cleaner already 👀 just a tiny nit

agent.go
JoTurk: I added it for the linter (exhaustive), i think that linter is stupid in general, but it's important in some case.

philipch07: Oh I missed that, that's fair then

PR #858, opened 2025-12-19
Cleaned up deadline setters in active tcp conns

philipch07:
#### Description
I had an idea when I saw the original impl and thought this might be nicer; but this might have to be undone if each one has to be more different in the future.

#### Reference issue
part of #854


JoTurk:
Go hate for dry, can be cringe sometimes :(

PR #861, opened 2025-12-29
Add DTLS-in-STUN (SPED) attributes

fippo:
defined in https://github.com/pion/stun/pull/260

### DIFF
--- a/.github/.gitignore
+++ b/.github/.gitignore
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+# SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
 # SPDX-License-Identifier: MIT
 
 .goassets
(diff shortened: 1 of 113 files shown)


sped.go
```diff
@@ -0,0 +1,71 @@
...
+	*d = v
+
+	return nil
+}
+
+// DtlsInStunAckAttribute is a STUN attribute for acknowledging the receipt
+// of DTLS packets (embedded in STUN or without embedding).
+type DtlsInStunAckAttribute []uint32
+
+// Acks are 32 bit values, the attribute can carry up to four of these.
```
fippo, line 38: I don't think we have consensus on the maximum length to put in the spec but this matches what is [currently in libwebrtc](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/p2p/dtls/dtls_stun_piggyback_controller.h;l=34;drc=56535b80c32d3a618b8fc8cfd7b1afdd3862e1b2).

fippo:
We have priority.go (priority), icecontrol.go (controlled/controlling) and usecandidate.go (use-candidate) in this repo though.
Which makes sense, ICE is an extension of STUN and these are just the attributes mentioned in https://www.rfc-editor.org/rfc/rfc5245#section-19. sped is similar in that respect.

PR #862, opened 2025-12-30
Expose asynchronous API for connect

fippo:
This adds a nonblocking API for ICE that will allow the PeerConnection's `startTransports` to start ICE,  *not* wait for ICE to connect and start DTLS so the DTLS handshake can be piggybacked into the STUN messages earlier.


jech:
Thanks.  The first commit looks good to me.

fippo:
> @fippo @jech @JoTurk I would like to do a new major version and break APIs to match `dtls`. We did the same to match `crypto/tls`
> 
> * Make `Dial` and `Accept` just return a `Controlling` or `Controlled` Agent.
> * Add `Handshake` that blocks until you go to connected
> * `Read` and `Write` start the handshake automatically.
> 
> Less confusing for existing Go devs.

this would mostly work (apart from read and write) integration as is? I can add comments capturing that plan which SGTM

JoTurk (approved):
I think this is a good change until we're able to break the API to match dtls in v2.

PR #863, opened 2026-01-01
Add API to write to specific candidate pair

kevmo314:
#### Description

Adds a `WriteToPair()` API to write to a specific candidate pair, which is useful for utilizing paths that are not the ICE-nominated path.

This change itself has no effect on any existing functionality, it only exposes a useful API. I will be utilizing the API
to send FEC packets over a secondary path so we utilize multiple links at the same time without jitter buffer pressure.

#### Reference issue
Fixes #...

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -105,7 +105,9 @@ type Agent struct {
 	remotePwd        string
 	remoteCandidates map[NetworkType][]Candidate
 
-	checklist []*CandidatePair
+	checklist  []*CandidatePair
+	nextPairID uint64
+	pairsByID  map[uint64]*CandidatePair
 
 	selectorLock sync.RWMutex
 	selector     pairCandidateSelector
@@ -336,6 +338,7 @@ func createAgentBase(config *AgentConfig) (*Agent, error) {
 		connectionState:                 ConnectionStateNew,
 		localCandidates:                 make(map[NetworkType][]Candidate),
 		remoteCandidates:                make(map[NetworkType][]Candidate),
+		pairsByID:                       make(map[uint64]*CandidatePair),
 		urls:                            config.Urls,
 		networkTypes:                    config.NetworkTypes,
 		onConnected:                     make(chan struct{}),
@@ -661,6 +664,7 @@ func (a *Agent) updateConnectionState(newState ConnectionState) {
 		if newState == ConnectionStateFailed {
 			a.removeUfragFromMux()
 			a.checklist = make([]*CandidatePair, 0)
+			a.pairsByID = make(map[uint64]*CandidatePair)
 			a.pendingBindingRequests = make([]bindingRequest, 0)
 			a.setSelectedPair(nil)
 			a.deleteAllCandidates()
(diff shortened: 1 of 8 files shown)


agent.go
```diff
@@ -1536,6 +1536,20 @@ func (a *Agent) GetSelectedCandidatePair() (*CandidatePair, error) {
...
+	err := a.loop.Run(a.loop, func(_ context.Context) {
+		pairs = make([]*CandidatePair, len(a.checklist))
+		copy(pairs, a.checklist)
+	})
+	if err != nil {
+		return nil
+	}
+
+	return pairs
+}
```
JoTurk: I don't think we should expose the internal candidates states or pointers, This will leak lifetime, and it will be kinda pain to maintain and support. We should prob return snapshots or internal ID, and add WriteTo at the con level, just next to Write, to write to specific pair.
Like WriteToPair in this PR.

JoeTurki (commented):
pretty cool :)

agent.go
kevmo314: That's fair, my concern with that is that we would need to iterate over all the candidates to find the matching one, which seems like a lot of overhead to do every packet. Do you have any ideas for how we could avoid needing to do that?

JoTurk: Maybe hold an internal map?

kevmo314: Done

JoeTurki (commented):
looks good to me, i'll look again and test when i wake up :)

stats.go
```diff
@@ -140,6 +140,27 @@ type CandidatePairStats struct {
...
+
+	// Nominated indicates whether this pair has been nominated.
+	Nominated bool
+
+	// CurrentRoundTripTime is the latest RTT measurement.
+	CurrentRoundTripTime time.Duration
+
+	// Quality is a score indicating the pair's quality (higher is better).
+	// Considers candidate types (host > srflx > relay), measured RTT, and stability.
+	Quality float64
```
JoTurk: This sounds like an application logic or user preference, maybe we should just export the candidate type and RTT and let the user decide?

transport.go
```diff
@@ -113,6 +113,44 @@ func (c *Conn) Write(packet []byte) (int, error) {
...
+		pair = c.agent.pairsByID[pairID]
+	}); err != nil {
+		return 0, err
+	}
+
+	if pair == nil {
+		return 0, ErrCandidatePairNotFound
+	}
+
+	if pair.state != CandidatePairStateSucceeded {
```
JoTurk: I think checking for state outside of loop is racy.

agent.go
```diff
@@ -1536,6 +1543,30 @@ func (a *Agent) GetSelectedCandidatePair() (*CandidatePair, error) {
 	return &CandidatePair{Local: local, Remote: remote}, nil
 }
 
+// GetCandidatePairsInfo returns snapshot information for all candidate pairs.
+// Use the returned ID with Conn.WriteToPair() to write to a specific pair.
+func (a *Agent) GetCandidatePairsInfo() []CandidatePairInfo {
```
JoTurk: Thinking about maybe both APIs write to pair and getCandidatePiarsInfo should be in the same struct (prob conn), but not sure.

stats.go
kevmo314: I renamed it to `RenominationQuality`. I think it's still useful to export the score used for renomination so you know what candidate is "next in line". But I also added the local and remote candidate types, I definitely think letting the user decide the metric they want is a good idea.

transport.go
kevmo314: Done

agent.go
kevmo314: Done, moved it to Conn

PR #866, opened 2026-01-06
Add UpdateOptions for runtime agent configuration

wrangelvid:
#### Description
This will be used implicitly by calling `pc.SetConfiguration` s.t. an ice restarts will use the newly added urls.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -169,8 +169,8 @@ var (
 	// ErrInvalidNetworkMonitorInterval indicates an invalid network monitor interval was provided.
 	ErrInvalidNetworkMonitorInterval = errors.New("network monitor interval must be greater than 0")
 
-	// ErrRunningAgentOptionNotUpdatable indicates an option cannot be updated on a running agent.
-	ErrRunningAgentOptionNotUpdatable = errors.New("option cannot be updated on a running agent")
+	// ErrAgentOptionNotUpdatable indicates an option cannot be updated after construction.
+	ErrAgentOptionNotUpdatable = errors.New("option can only be set during agent construction")
 
 	errAttributeTooShortICECandidate = errors.New("attribute not long enough to be ICE candidate")
 	errClosingConnection             = errors.New("failed to close connection")
--- a/agent_options.go
+++ b/agent_options.go
@@ -51,7 +51,7 @@ func DefaultNominationValueGenerator() NominationValueGenerator {
 func WithAddressRewriteRules(rules ...AddressRewriteRule) AgentOption {
 	return func(agent *Agent) error {
 		if agent.constructed {
-			return ErrRunningAgentOptionNotUpdatable
+			return ErrAgentOptionNotUpdatable
 		}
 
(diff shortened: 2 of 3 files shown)


agent.go
```diff
@@ -1574,6 +1574,14 @@ func (a *Agent) SetRemoteCredentials(remoteUfrag, remotePwd string) error {
...
 }
 
+// UpdateURLs updates the STUN/TURN server URLs used for gathering.
+// The new URLs will take effect on the next call to GatherCandidates.
+func (a *Agent) UpdateURLs(urls []*stun.URI) error {
+	return a.loop.Run(a.loop, func(_ context.Context) {
+		a.urls = urls
+	})
+}
+
```
JoTurk: We recently cleaned up the ICE options API and standardized on functional options. Given that, I think it makes sense to add an UpdateOptions method that accepts ICE options and explicitly whitelists which ones can be updated at runtime.

That would allow patterns like: `agent.UpdateOptions(WithURLs(...))`

Using our already existing options.

https://github.com/wrangelvid/ice/blob/1b616e70dec5633ba929376e7fddbbf6d7616459/agent_options.go

JoTurk (changes requested):

agent.go
wrangelvid: Should we enforce the whitelist or just document it?
If we want to enforce it, we may need a wrapper type. My understanding is that due to opaque functions, this api `func (a *Agent) UpdateOptions(opts ...AgentOption) ` wouldn't allow us to check at runtime which option is what which. However, we can introduce a wrapper type, but that may create more duplication `func (a *Agent) UpdateOptions(opts ...UpdatableAgentOption)`.

JoTurk: Since we do define all the options. we can just return an Error from options that can't be used at the runtime, options can access the agent object. We could even add a simple boolean state for constructed or not. So we don't touch the other agent states.

wrangelvid: Clever! I just gave it a crack.

errors.go
```diff
@@ -169,6 +169,9 @@ var (
 	// ErrInvalidNetworkMonitorInterval indicates an invalid network monitor interval was provided.
 	ErrInvalidNetworkMonitorInterval = errors.New("network monitor interval must be greater than 0")
 
+	// ErrRunningAgentOptionNotUpdatable indicates an option cannot be updated on a running agent.
+	ErrRunningAgentOptionNotUpdatable = errors.New("option cannot be updated on a running agent")
```
JoTurk: I think this should be ErrAgentOptionNotUpdatable or ErrAgentOptionOnlyAtConstruction, because we don't allow update after construction. even if it's not running.

JoTurk (commented):
Thank you, I think we can merge and tag this, after we update the error name. we prob should add a test for the behavior. but i think this can be done later.

errors.go
wrangelvid: Done. I did add a test before. Happy to extend it, but I may need clarification.

JoTurk (approved):
thank you.

PR #868, opened 2026-01-07
Fix ICE Agent gathering notification on Close

boushley:
#### Description

Ensure that when an ICE Agent is closed that the complete handlers fire. As highlighted in https://github.com/pion/webrtc/issues/2507 not firing this handler results in hung go routines when following common patterns.

#### Reference issue
Fixes https://github.com/pion/webrtc/issues/2507

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1214,6 +1214,12 @@ func (a *Agent) GracefulClose() error {
 }
 
 func (a *Agent) close(graceful bool) error {
+	// Ensure that we mark gathering as complete
+	err := a.setGatheringState(GatheringStateComplete)
+	if err != nil {
+		a.log.Warnf("Failed to setGatheringState to complete while closing: %v", err)
+	}
+
 	// the loop is safe to wait on no matter what
 	a.loop.Close()
 
--- a/agent_test.go
+++ b/agent_test.go
@@ -2593,3 +2593,24 @@ func TestAgentUpdateOptions(t *testing.T) {
 		}
 	})
 }
+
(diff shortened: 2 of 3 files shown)


agent.go
```diff
@@ -1208,6 +1208,12 @@ func (a *Agent) GracefulClose() error {
 }
 
 func (a *Agent) close(graceful bool) error {
+	// Ensure that we mark gathering as complete
+	err := a.setGatheringState(GatheringStateComplete)
```
JoTurk, line 1218: I think we'll need to make sure we call this once per a gathering cycle, Restart_During_Gather fails because it expects gather to complete once which sounds more correct.
Other than that we'll need a unit test to make sure we don't break this again.

boushley: What do you mean by "Restart_During_Gather"? I reviewed the `Agent#Restart` method and I don't see anything in there that fails / errors out based on the current state. Am I missing something?

As for ensuring we only do this once per cycle, I'm trying to understand what we're looking to avoid. The `setGatheringState` function won't trigger the broadcast of the `nil` candidate again unless we are actually changing the state (so if it was already Complete this is essentially a no-op).

Can you help me understand the side effect we're trying to prevent or an example of limiting to once per gathering cycle?

boushley:
@JoTurk let me know in the morning what you think of this approach. Sounds like you have a different idea in mind around the generations, but I backed off on trying that after discovering that all of this diverges from the browser's behavior which seems to be paramount.

Also I have no idea why the linter is upset about those function calls without context. I don't see those failures on `master`, but I also didn't modify those files in any way :/

PR #874, opened 2026-01-09
Add gathering generation extension to candidates

boushley:
#### Description
Populate the generation extension on candidates added to ICE Agent.


agent.go
```diff
@@ -1682,6 +1703,34 @@ func (a *Agent) setGatheringState(newState GatheringState) error {
...
+
+	prevState := a.gatheringState
+	if prevState == newState {
+		return
+	}
+
+	if newState == GatheringStateComplete && !a.gatherEndSent {
+		if prevState == GatheringStateGathering {
+			a.candidateNotifier.EnqueueCandidate(nil)
+		}
```
boushley: This is a slight departure from the previous behavior that was part of the base patch from @JoTurk

I'm not sure if this behavior (only send `nil` candidate when transitioning from `Gathering` to `Complete`) is more correct, or if we should keep the original behavior (send `nil` candidate anytime we transition into the `Complete` state)

boushley:
Converted to draft since I missed that I should be piping generation through the call stack from `Agent#GatherCandidates` instead of pulling fresh from `agent` within the individual gather methods. I'll update that.

agent.go
JoTurk: yeah, the reason i added this is because i just wanted to make the test pass (because the previous pr did close => complete), but sending nil regardless is correct i think.

JoTurk:
Thank you I'll check again how chrome and firefox handle generations and review this. this pr is continuation of https://github.com/pion/webrtc/pull/3113

agent.go
JoTurk: this is because we don't go to complete from close, like in the previous pr.

gather.go
```diff
@@ -1215,7 +1216,13 @@ func (a *Agent) startNetworkMonitoring(ctx context.Context) {
...
 		case <-ticker.C:
 			if a.detectNetworkChanges() {
-				a.gatherCandidatesInternal(ctx)
+				generationChan := make(chan uint64, 1)
+				a.loop.Run(a.loop, func(_ context.Context) {
+					newGen := a.bumpGatheringGenerationLocked()
+					generationChan <- newGen
+				})
+				generation := <- generationChan
+				a.gatherCandidatesInternal(ctx, generation)
```
boushley, line 1439: @JoTurk thoughts on this? It feels a bit janky, but I wasn't sure of a cleaner way. I'm assuming that re-triggering candidate gathering because of a network change should trigger a new generation. If not then I can revert this, but if so I'm open to recommendations on cleaner ways to run the bump and then gather candidates again, I didn't want to run the actual `gatherCandidatesInternal` method within the agent lock/loop as that seems like a major change.

JoTurk: I'm not entirely sure, since browsers don't do continual gathering and there isn't really a standard behavior to reference here.  But I think this makes sense, we don't want the remote side attempting to connect to candidates before a network change. Also it's a way to signal that a gather did happen and how many gathers happend.
 
Kinda relates to this https://github.com/pion/webrtc/pull/3323

wrangelvid: I don't think we should bump the generation during continual gathering. Network changes can be **additive**, for example, if a peer on a mobile network connects to wifi, a _new_ path becomes available while the old one remains valid. We want to allow the ICE agent to smoothly _renominate_ the better path without tearing down the session.

My understanding is that generations are useful for hard resets, like ice restarts where credentials change, rendering previous candidates invalid. Since continual gathering uses existing credentials, the generation id must remain the same.

JoTurk: @wrangelvid Generations aren't used to tear down or invalidate already-accepted candidates. They're used to reject new candidates from an older context when we call AddCandidates. In this module, once a candidate has been accepted, bumping the generation won't affect it at all so nothing to worry about.

During a network change we'll still generate candidates for all available paths, including additive, and the ICE agent can continue to smoothly renominate the better path without disrupting the existing session. It will just make sure to reject old generation candidates if they got sent late. and not waste time try to connect to most likely not working candidates.

wrangelvid: Mhh, I might need a bit more help to understand this, but I don't see the advantage yet.
My understanding is that the receiver discards identical candidates automatically. Wouldn't bumping the generation break this deduplication and trigger redundant connectivity checks for paths that are already connected?

Regarding late candidates: if a candidate from the current session arrives late, wouldn't we actually want to accept it? Since the credentials haven't changed, it's a valid backup path that just arrived out of order. I thought we only needed to strictly reject late candidates during ICE restarts because the credentials become invalid.

Also, if an interface is truly gone (e.g. unplugged ethernet), we won't gather new candidates for it anyway, so there's nothing to filter there.

JoTurk: @wrangelvid please look at https://github.com/pion/webrtc/issues/2993 so the problem is that if we use a trickle  signaling protocol that collects and send candidates at once or send candidates out of order, we might apply older candidates by mistakes.
After my fix in https://github.com/pion/webrtc/pull/3113 we ignore candidates using the frag/username, but this alone isn't enough now that we support continual gathering (I stated in the PR that we'll have to do generations eventually).
A real-world case would be a user switching from WifI -> data -> WiFi again, where we end up collecting and signaling candidates from both network changes concurrently. In that situation, we should avoid attempting to connect to candidates from the intermediate data transition once it's no longer valid (just wasting time + resources).

As in your example, this doesn't affect already-accepted candidates, it only impacts AddCandidate. Once we accept a candidate from a newer generation, we should ignore late-arriving candidates from older generations (e.g., delayed by the signaling or collecting path). So user switching from data to data and wifi at the same time, won't invalidate accepted candidates, it'll just add more candidates options.

agent.go
boushley: If I'm [reading this right](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/peerconnection/rtc_ice_transport.cc;l=189-192;drc=279feee6aa7430af23c09d08e16e5a4a0928f8ab) it looks like the previous behavior of sending the `nil` candidate anytime we enter `Complete` as long as the previous state wasn't also `Complete` is in line with Chrome's behavior.

agent.go
```diff
@@ -1682,6 +1703,32 @@ func (a *Agent) setGatheringState(newState GatheringState) error {
...
+		return
+	}
+
+	prevState := a.gatheringState
+	if prevState == newState {
+		return
+	}
+
+	if newState == GatheringStateComplete && !a.gatherEndSent {
+		a.candidateNotifier.EnqueueCandidate(nil)
```
boushley: @JoTurk I wonder if we should also be concerned about calling this from within the "Lock" / AgentLoop here. Specifically the fact that `EnqueueCandidate` starts by acquiring the `handlerNotifier` Lock. I'd be concerned that we could end up delaying the Agent Loop if not fully deadlocking it.

gather.go
wrangelvid: I see your point on network flapping. Though, couldn't the second Wifi gather be identical to the first gather? If we use generations to treat it as "new", wouldn't we trigger unnecessary connectivity checks on a path we are already connected to? Standard ICE deduplication should normally handle this for free.

Also, assuming the wifi drops again (`Wifi (gen 1) -> data (gen 2)-> Wifi (gen 3)-> data (gen 4)`), isn't gen 3 throwing away the candidates of gen 2 right away? We may have dropped a valid warm backup path and have to wait for gen 4 to reconnect, whereas standard ICE would have just kept the gen 2 candidates alive as a fallback.

I definitely agree that we need to avoid a combinatorial explosion of candidate pairs and wasting resources on dead paths. However, strict generation-based filtering at the ingress seems to trade reliability for efficiency in a way that risks connectivity during these edge cases. I just wanted to throw in another perspective after digging myself too deep into renomination and continual gathering. Maybe there is a middle ground that prunes old candidates without hard-rejecting them if they arrive out of order?

JoTurk: @wrangelvid I think there is a miss-understanding here,  generation bump here is not an ICE restart and it doesn't invalidate already-accepted candidates or the nominated pair, it only gates late arrivals at AddCandidate. So gen3 doesn't "throw away" gen2 candidates that were already applied. it just ignores out-of-order/late trickle from an older gather wave (gen2 candidates after we start receiving gen3 candidates), which is exactly what causes pair explosion during flapping + batched signaling. If you're worried about "identical Wi-Fi gather triggers redundant checks," If network didn't change, then cool the new generation will be the same as the old generation, it's just a gate against spending checks on stale network context from older generations.

wrangelvid: I understand that it's not invalidating existing candidate pairs. My concern is that the "gate" may prevent us from acquiring a valid backup path if signaling is out of order. A late arrival of the data candidates doesn't necessarily mean it's a stale network context, just slow signaling, and we'd want those candidates for failover.

JoTurk: We have to decide here:
1. Reject older generation candidates and save network time on most likely not valid network paths. and risk slight delay if we reject a valid candidate (we'll end up receving it again with the new generation anyway).
2. Accept all candidates, in case some of the candidates are valid, and risk wasting a lot of time in cases the user switch between different networks really quick.

I think your scenario will be super rare in the real world, what will be most likely the case is like in whip/whep signaling where we collect many candidates and batch them in one response as in the original issue https://github.com/pion/webrtc/issues/2993 and in cases like this we have candidates from both generations.

wrangelvid: Fair enough, that's a trade-off. We'll try it with generations on mobile robots when we get a chance and compare. Worst case, we'll vendor patch this.

JoTurk: Can always remove the generation extension from the candidate text, if you have a non standard signaling protocol  :)

wrangelvid: Well, we should keep the generation extension for ice restarts ◡̈

boushley: Alright, seems like we've landed on trying the generation bump on network change for now. We can certainly reevaluate in the future.

With that agreement in place are there thoughts on whether there's a better way to be executing that on the Agent Loop and getting the generation value back to this go-routine?

JoTurk:
@boushley sorry I lost track of this, Will try to get back to you tomorrow.

agent.go
JoTurk: Maybe we can keep the transition state, and add "should we emit end of the candidates" as a notifier after the loop lock is released?


(Sorry i was super busy with new job + moving to a new place past few weeks).

boushley:
I'm open to ideas of how we should handle the concurrency issues here. I worry about the current state because it could potentially result in a deadlock where the Agent loop gets hung up trying to acquire the notifier lock, although I don't think there's a place where the notifier lock is dependent on the Agent loop, I haven't confirmed that side.

I also don't see a clean way of getting the notification with the `nil` candidate pulled out of the agent loop especially since this is done from at least two locations that can set the state to Complete :/

Should I dig in on what exactly the Notifier lock can hang on to gain confidence we're safe on that side? Or can you think of a cleaner way of getting the notification out on a less critical go routine?

JoTurk:
@boushley I'm so sorry i lost this from my todo (again), I think your approach is the correct one, If you don't mind fixing the conflicts and I'll get into reviewing and testing this.
I think we'll just need a few unit tests before we merge this.

Sean-Der (approved):

gather.go
Sean-Der:
```
Generation:
All of the candidates conveyed within an ICE session (correlated with a particular Username Fragment and Password combination).
```

I am merging now! We shouldn't do a generation bump. Generations only happen on frag/passwd changes https://www.rfc-editor.org/rfc/rfc8838.html

JoTurk: @Sean-Der the generation extension isn't part of the ice spec, it's originally from https://xmpp.org/extensions/attic/xep-0176-0.13.html but yeah libwebrtc only generates them if ufrag/password changes we can update this code now https://github.com/pion/webrtc/pull/3113 to account for this extension (currently it can race since it doesn't know if ufrag/username are from the old or the new generation)

PR #876, opened 2026-01-19
Implement SPED (STUN Protocol for Embedding DTLS)

fippo:
(based on #914 )

Part of https://github.com/pion/webrtc/issues/3335


cnderrauber:
Thank you for this great feature! I have integrated this into the livekit SFU and made some connectivity checks, found 2 regression cases:
1. The ice-lite agent doesn't ping the remote actively, so the piggyback packets will be sent in response to the remote peer's connectivity check, cost longer than the non-sped case. Fixing this could conflict with spec, maybe we can mark it as an exception.
2. When a remote peer sends a `done` signal after the agent is connected while still having inflight piggyback packets, it will mark the state to `complete` and doesn't have a chance to flush the packets. And also the Piggyback() will swallow all dtls packets but return true in this state (complete after connected).

Both are fixed in our fork https://github.com/livekit/ice/commit/85afccdd578d5f7a282fd8221c1fbda336449dc1, https://github.com/livekit/ice/commit/5eec925b050f5558a223e75a89fdb87da496b912. Will merge them back to upstream once this PR is merged if they make sense.

fippo:
@cnderrauber oops, forgot to respond. 

(1) the tl;dr is that you can send DTLS unembedded once ice is writable which for lite means you can send after receiving the first valid binding check. https://fippo.github.io/warp-snap-sped/draft-hancke-webrtc-sped.html#section-3.3.1 bullet (3) covers that

(2) good catch!

piggyback.go
```diff
@@ -0,0 +1,236 @@
...
+	defer a.piggyback.mu.Unlock()
+	if a.piggyback.state == PiggybackingStateOff {
+		return a.connectionState != ConnectionStateConnected
+	}
+
+	if packet != nil {
+		// If we receive a packet after the end of a flight we need
+		// to clear the outgoing list.
+		if a.piggyback.newFlight {
+			a.piggyback.packets = []packetWithCrc{}
```
JoTurk: should we reset packetsIndex too?

piggyback.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// Piggyback stores a packet to be picked in a round-robin fashion.
+// Returns `true` if packet is to be consumed.
+func (a *Agent) Piggyback(packet []byte, end bool) bool {
+	a.piggyback.mu.Lock()
+	defer a.piggyback.mu.Unlock()
+	if a.piggyback.state == PiggybackingStateOff {
+		return a.connectionState != ConnectionStateConnected
+	}
```
JoTurk: shouldn't we queue / store packets before `connected`, wouldn't this affect early flight dtls packets?

piggyback.go
```diff
@@ -0,0 +1,236 @@
...
+		// to clear the outgoing list.
+		if a.piggyback.newFlight {
+			a.piggyback.packets = []packetWithCrc{}
+		}
+		a.piggyback.newFlight = end
+		crc := crc32.ChecksumIEEE(packet)
+		a.piggyback.packets = append(a.piggyback.packets, packetWithCrc{packet, crc})
+	} else {
+		a.piggyback.state = PiggybackingStatePending
+	}
```
JoTurk: Sorry if this comment is dumb but shouldn't Piggyback(nil, true) clear packets and state or do we clear at another point?

piggyback.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// Piggyback stores a packet to be picked in a round-robin fashion.
+// Returns `true` if packet is to be consumed.
+func (a *Agent) Piggyback(packet []byte, end bool) bool {
+	a.piggyback.mu.Lock()
+	defer a.piggyback.mu.Unlock()
+	if a.piggyback.state == PiggybackingStateOff {
+		return a.connectionState != ConnectionStateConnected
```
JoTurk: reading connectionState isn't safe outside of agent's loop. not sure if this will cause issues.

JoTurk (commented):
Sorry about ghosting this for too long,  I have a few comments mainly about how Piggyback interacts with the agent.

piggyback.go
fippo: Good catch, updating!

piggyback.go
fippo: better safe than sorry... added a flag for this.

piggyback.go
fippo: Yes, @zshang-oai found this too. This needs to go into the capture + flush part. I have a fix but keeping this open, this needs more :eyes:

piggyback.go
fippo: No, (nil, true) means our local DTLS emitted all packets in [the webrtc glue](https://github.com/pion/webrtc/pull/3362/changes#diff-b13505853b1a1e1b02b40dea692421a19cd825553c703b44dbc067acef9a99b4R589)
This does not mean we know the peer has received all the packets.

fippo:
updated!

piggyback.go
```diff
@@ -0,0 +1,240 @@
...
+
+// Piggyback stores a packet to be picked in a round-robin fashion.
+// Returns `true` if packet is to be consumed.
+// A nil packet signals that the local DTLS handshake completed.
+func (a *Agent) Piggyback(packet []byte, end bool) bool {
+	a.piggyback.mu.Lock()
+	defer a.piggyback.mu.Unlock()
+	if a.piggyback.state == PiggybackingStateOff && a.piggyback.connected {
+		return false
+	}
```
JoTurk, line 124: Shouldn't this also include complete to fix the issue reported / fixed by livekit with late retransmits? @cnderrauber https://github.com/livekit/ice/commit/85afccdd578d5f7a282fd8221c1fbda336449dc1 this maybe should also extend to address packets that were already queued.

fippo: https://github.com/pion/ice/pull/876#discussion_r3624283924 -- was thinking of having that done as a f'up.

piggyback.go
```diff
@@ -0,0 +1,240 @@
...
+		if a.piggyback.newFlight {
+			a.piggyback.packets = []packetWithCrc{}
+			a.piggyback.packetsIndex = 0
+		}
+		a.piggyback.newFlight = end
+		crc := crc32.ChecksumIEEE(packet)
+		a.piggyback.packets = append(a.piggyback.packets, packetWithCrc{packet, crc})
+	} else if a.piggyback.state != PiggybackingStateOff {
+		a.piggyback.state = PiggybackingStatePending
+	}
```
JoTurk, line 141: This is triggered on dtls handshake right? from the comment "// A nil packet signals that the local DTLS handshake completed." 
what happen if dtls handshake happens after complete? do we go back to pending or is it impossible because of something I'm missing?

https://github.com/pion/ice/pull/876/changes#diff-bed5e122cd63ea9db2c3fd9eb61077670debbebb87c1d898c640d9057f288842R146-R149

fippo: wait, is this a deja-vu? https://github.com/pion/ice/pull/876#discussion_r3624355998 😂

JoTurk: Sorry i misworded my comment, My concern here is different from my old comment, Is it possible for the state to be "complete" when (nil, true) gets called? and if that happens do we move back to pending?

JoTurk: (Also from the glue this is called in completeStart which is called after full dtls handshake, this gets triggered after Finished in 1.2 and 1.3) so my concern is what happens if dtls handshake happens after sped handshake and is it possible?

JoTurk: In other words what happens if pion receives and process the stun message before the pion/dtls' hanshake channel unlocks? do we enter complete => pending race?

piggyback.go
fippo: (hopefully fixed because of resync against libWebRTC M144)

fippo:
I gave up and resynced with C++ logic ...

piggyback.go
```diff
@@ -0,0 +1,297 @@
...
+
+// Piggyback stores a packet to be picked in a round-robin fashion.
+// Returns `true` if packet is to be consumed.
+// A nil packet signals that the local DTLS handshake completed.
+func (a *Agent) Piggyback(packet []byte, end bool) bool {
+	a.piggyback.mu.Lock()
+	defer a.piggyback.mu.Unlock()
+	if a.piggyback.state == PiggybackingStateOff && a.piggyback.connected {
+		return false
+	}
```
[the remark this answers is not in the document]
JoTurk: From my understanding the original libwebrtc does not suppress the normal ICE SendPacket, so bugs like this aren't possible, is there a limitation in pion that prevents us from having the same behavior as libwebrtc?

piggyback.go
```diff
@@ -0,0 +1,297 @@
...
+			a.piggyback.packetsIndex = 0
+		}
+		a.piggyback.newFlight = end
+		crc := crc32.ChecksumIEEE(packet)
+		a.piggyback.packets = append(a.piggyback.packets, packetWithCrc{packet, crc})
+	} else if a.piggyback.state != PiggybackingStateOff && a.piggyback.state != PiggybackingStateComplete {
+		// As DTLS 1.2 client we have nothing more to send while as server we
+		// need to keep the last flight around until it gets acknowledged.
+		// DTLS 1.3 reverses this.
+		if a.piggyback.isDtlsClient {
```
JoTurk: Can we make `SetDtlsRole` take DTLS version so we don't have to break the API when we release DTLS 1.3 soon?

piggyback.go
JoTurk: Seems fixed now :)

piggyback.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+		// If we receive a packet after the end of a flight we need
+		// to clear the outgoing list.
+		if a.piggyback.newFlight {
+			a.piggyback.packets = []packetWithCrc{}
+			a.piggyback.packetsIndex = 0
+		}
+		a.piggyback.newFlight = end
+		crc := crc32.ChecksumIEEE(packet)
+		a.piggyback.packets = append(a.piggyback.packets, packetWithCrc{packet, crc})
```
JoTurk: We should either copy here or document that the caller must not modify the buffer after Piggyback returns.

JoTurk (commented):
sorry for all the back and forth :(

piggyback.go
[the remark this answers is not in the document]
fippo: I think copilot made this up, I can't see how this would happen.

piggyback.go
fippo: setDtlsRole is pre-handshake, the negotiated DTLS version is only known afterwards. Like libWebRTC we need glue code that signals the completion together with the version.

fippo:
still reasonable given the amount of cross-repo work required. Now it also seems we need to plumb ice-options if this is to be negotiated with ice-options ...

piggyback.go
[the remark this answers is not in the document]
JoTurk: Hmm, classic clankers...

piggyback.go
JoTurk: Sounds good, we expose the version in dtls but maybe we expose it late, I'll make sure expose it early in the handshake

piggyback.go
```diff
@@ -0,0 +1,299 @@
...
+		// to clear the outgoing list.
+		if a.piggyback.newFlight {
+			a.piggyback.packets = []packetWithCrc{}
+			a.piggyback.packetsIndex = 0
+		}
+		a.piggyback.newFlight = end
+		crc := crc32.ChecksumIEEE(packet)
+		// Copy the packet as the caller may reuse the underlying buffer.
+		data := make([]byte, len(packet))
+		copy(data, packet)
```
JoTurk: Nit: can use data := bytes.Clone(packet)

piggyback.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	a.piggyback.state = PiggybackingStateOff
+}
+
+// SetDtlsHandshakeComplete signals that the local DTLS handshake completed and
+// carries the negotiated DTLS role and version. The party that sends the last
+// flight has to keep it around until it gets acknowledged; that is the server
+// in DTLS 1.2 and the client in DTLS 1.3. The other party has nothing more to
+// send and drops its outgoing packets.
+func (a *Agent) SetDtlsHandshakeComplete(isClient, isDtls13 bool) {
```
JoTurk: Maybe instead of isdtls13 we should just have uint16 value of the version as a go "enum"
```go
type dtlsVersion uint16

const (
	DTLS12 dtlsVersion = 0xfefd
	DTLS13 dtlsVersion = 0xfefc
)
```

JoTurk (approved):
Just two nits, looks great, thank you and sorry again

piggyback.go
fippo: this becomes an export hassle and I do not think we will have DTLS 1.4 anytime soon (and nobody likes 1.0)

JoTurk: I agree but isdtls13 just doesn't feel "right" for a public api, but I don't have a hard feeling about this :)

JoTurk:
Also I wonder if we should move the sped methods into a new optional object that the user can pass to options With sped(controller) so we don't have methods on agents but this isn't a big deal because ice is a due for an api rework and refactor and we can do all of this in the next major release. And release this for now.

piggyback.go
fippo: yeah but ideally DTLS defines its version but ice can't import from DTLS and lots of conversion and ... :-|

piggyback.go
fippo: I checked how it behaves in libWebRTC and it is simple, gets emitted after the handshake is complete and all packets have been sent.

piggyback.go
JoTurk: We export the versions in dtls/pkg/protocol https://github.com/pion/dtls/blob/main/pkg/protocol/version.go and ice can import them and they are released with v3 so we don't have to wait for v4 :)

piggyback.go
JoTurk: Sounds good, the version should be available through dtls state by then

piggyback.go
fippo: yes but that creates a rather awkward dep from ice on dtls...

JoTurk: Ice already imports dtls directly xd https://github.com/pion/ice/blob/main/go.mod

fippo: whyyyyyyyyy...

(will update, awkwardness is not on me!)

zshang-oai:
From my Codex and internal security tool, just for your reference @fippo 

>  Security concern: reportPiggybackingFromMessage extracts DTLS-IN-STUN and DTLS-IN-STUN-ACK without checking whether those attributes appear before MESSAGE-INTEGRITY.
Since STUN integrity only authenticates attributes preceding MESSAGE-INTEGRITY, an attacker could append SPED attributes after a valid integrity field without invalidating the existing HMAC. Those unauthenticated attributes would still reach ReportPiggybacking, potentially injecting DTLS data or modifying ACK state.
Could we either enable strict STUN decoding or explicitly reject SPED attributes appearing after MESSAGE-INTEGRITY? Regression tests should cover both appended DTLS and ACK attributes.

fippo:
https://github.com/pion/stun/pull/273 should be on by default?

JoTurk:
@fippo we're going to tag a new major release for STUN soon because of https://github.com/pion/stun/pull/284 maybe we can switch strict mode to be on by default?

fippo:
@JoTurk  ah... we didn't ship that. Yes, lets do it as part of a breaking change.

PR #877, opened 2026-01-20
Updated received trickled candidates handling

### LINKED ISSUE #622: Candidate type is not changed from peer-reflexive to host
### Your environment.
  - Version: pion v3.1.60, pion/ice v2.3.2

### What did you do?
Two pion-based peers establish WebRTC connection with ICE trickle enabled.

### What did you expect?
When both peers are on the same network (especially in tests where both are in the same process), both of them selects host-host pair as a best one and reports it via `peer.SCTP().Transport().ICETransport().OnSelectedCandidatePairChange()` callback. I am interested in last reported pair.

### What happened?
When ICE trickle is enabled, there is race condition between ICE candidate sent via websocket and STUN request. When latter one arrives first, pion/ice selects host-prflx pair as a best one and reports it via callback mentioned above.

My code differentiated between direct host-host connections and other ones, so sometimes it works incorrectly.

Related part of ICE Trickle RFC: https://www.rfc-editor.org/rfc/rfc8838#section-11-2.4.2.1

sirzooro:
Implements RFC 8838 §11.4 (“Receiving Trickled Candidates”) behavior so that when a remote peer-reflexive (prflx) candidate is discovered first, a later signaled equivalent candidate (host/srflx/relay) replaces it in-place while preserving checklist/pair behavior.

## Key Changes

- **prflx replacement on trickle (RFC 8838 §11.4)**
  - Replace redundant remote prflx candidates when a signaled candidate with the same transport address arrives.
  - Upgrade existing checklist pairs in-place (remote candidate swap), while preserving the pair priority.
  - Update local candidate remote-candidate caches so subsequent lookups use the signaled candidate.
  - Avoid creating duplicate pairs when the replacement occurs.

- **prflx priority from inbound STUN (RFC 5245 §7.1.3.2.1)**
  - When a prflx candidate is discovered via an inbound Binding Request, set its priority from the STUN `PRIORITY` attribute.

~~- **Candidate pair priority correctness (RFC 5245)**~~
  ~~- Fix pair-priority computation to use the RFC 5245 formula with $2^{32}$ (not $2^{32}-1$).~~
  ~~- Add an internal helper for the computation and make it overflow-safe via saturation.~~
  ~~- Add a pair-level priority override so prflx→signaled replacement can preserve the previously computed pair priority.~~
(Restored original code, this change was not needed)

- **Handler notifier deadlock/starvation fix**
  - Split the notifier’s single `running` flag into independent running state for connection-state, candidate, and selected-pair queues.
  - Prevent callback starvation that could cause tests (and user code) waiting on connection-state transitions to hang.

~~- **Test stability improvements (Active TCP)**~~
  ~~- Prefer non-link-local IPv6 addresses and skip IPv6 cases when only link-local addresses exist.~~
  ~~- Use mDNS on non-Windows platforms but disable it on Windows to reduce flakiness.~~
  ~~- Increase the per-subtest timeout to reduce intermittent CI/dev timeouts.~~
(Moved to another PR)

#### Reference issue
Fixes #622


.gitignore
```diff
@@ -14,6 +14,7 @@
 bin/
 vendor/
 node_modules/
+.vscode/
```
JoTurk: Can you please add this to .goassets instead? it will get overwritten the next time the bot does sync.

active_tcp_test.go
```diff
@@ -53,17 +55,13 @@ func ipv6Available(t *testing.T) bool {
...
+		if !addr.addr.IsLinkLocalUnicast() {
+			return true
 		}
-
-		return false
 	}
 
-	return len(localAddrs) > 0
+	return false
 }
```
JoTurk, line 60: I added this for mac only because I wasn't sure if it affects all the platforms we support. anyways, this should be done in a separate PR.

active_tcp_test.go
```diff
@@ -94,29 +92,31 @@ func TestActiveTCP(t *testing.T) {
 	}
 
 	if ipv6Available(t) {
+		// mDNS on Windows is flaky in CI/dev environments; prefer it elsewhere.
+		// if we don't use mDNS, we will very likely be filtering out location tracked ips.
+		useMDNS := runtime.GOOS != "windows"
+
```
JoTurk: I think all the test adjustments should be done in separate PRs.

agent.go
```diff
@@ -954,6 +954,43 @@ func (a *Agent) requestConnectivityCheck() {
...
+
+	addrA := removeZoneIDFromAddress(a.Address())
+	addrB := removeZoneIDFromAddress(b.Address())
+
+	return a.NetworkType() == b.NetworkType() &&
+		a.Component() == b.Component() &&
+		addrA == addrB &&
+		a.Port() == b.Port() &&
+		a.TCPType() == b.TCPType()
+}
```
JoTurk: Why we need another equality helper when we have candidate-base's Equal which also compares RelatedAddress.
If it's about fixing the equality for ipv6 addresses with zones, I think we should fix the original helper, I didn't check if works for Ipv6 addresses with zones, but we did fix it for turn recently.

agent.go
```diff
@@ -1015,9 +1052,24 @@ func (a *Agent) addRemotePassiveTCPCandidate(remoteCandidate Candidate) {
...
+	if cand.Type() != CandidateTypePeerReflexive {
+		for i := 0; i < len(set); i++ {
+			existing := set[i]
+			if existing.Type() == CandidateTypePeerReflexive && sameTransportAddress(existing, cand) {
+				replacedPrflx = append(replacedPrflx, existing)
+				set = append(set[:i], set[i+1:]...)
+				i--
+			}
+		}
+	}
```
JoTurk: Nit: we're in the process in cleaning ICE #854 and it would be cool to clean new code, and not add new nolints for cyclop or gocognit.

I think this should be refactored to a another function but it's not a blocker.

.gitignore
sirzooro: Done: https://github.com/pion/.goassets/pull/263

active_tcp_test.go
sirzooro: Ok, will move this to another PR.

active_tcp_test.go
sirzooro: Ok, will move this to another PR.

agent.go
sirzooro: `Equal` also compares `Type` and `RelatedAddress`, so it is not suitable here. My version also checks if components are equal (maybe Equal should do this too). I added part to remove IPv6 zones after I got comment to ignore them in one of my recent PRs here.

agent.go
sirzooro: Done, moved code to replaceRedundantPeerReflexiveCandidates

JoTurk (approved):
Thank you.

PR #906, opened 2026-04-06
Add generic STUN send hook for outbound message rewriting

zshang-oai:
## Background
RFC 8489 Section [14.2](https://www.rfc-editor.org/rfc/rfc8489.html#section-14.2) expects a STUN Binding Success response to report the client's transport address in `XOR-MAPPED-ADDRESS`. That becomes incorrect when the ICE server is behind a proxy or relay hop and only sees the proxy-facing tuple.

This PR reframes the original request from a narrow address-override callback into a broader outbound STUN extension point. The immediate use case is allowing upstream integrations such as `pion/webrtc` to rewrite Binding Success responses based on proxy metadata, but the hook is generic enough for other STUN response/request customization.

## Summary
- replace the narrow `WithBindingSuccessXORMappedAddressHandler(...)` hook with a generic `WithSTUNSendHandler(...)`
- call the handler for outbound STUN sends, including Binding Success responses
- run the handler for Binding Success before `MESSAGE-INTEGRITY` and `FINGERPRINT` are added, so callers can safely modify attributes like `XOR-MAPPED-ADDRESS`
- keep the existing `WithBindingRequestHandler(...)` behavior unchanged
- add focused tests for option wiring and Binding Success rewriting via the new hook

## Testing
- `go test ./... -run 'TestWithSTUNSendHandler|TestSTUNSendHandlerOverridesBindingSuccessResponseAddress' -count=1`
- `go test ./...`
- `golangci-lint run`

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -38,6 +38,22 @@ type bindingRequest struct {
 	nominationValue *uint32 // Tracks nomination value for renomination requests
 }
 
+// STUNSendHandler allows applications to inspect or modify outbound Binding
+// responses before they are sent.
+//
+// The callback receives the outbound message plus the inbound request that
+// triggered it.
+type STUNSendHandler func(
+	outbound, inbound *stun.Message,
+	local, remote Candidate,
+) error
+
+type stunSetterFunc func(*stun.Message) error
+
+func (f stunSetterFunc) AddTo(message *stun.Message) error {
+	return f(message)
+}
+
 // Agent represents the ICE agent.
 type Agent struct {
 	loop *taskloop.Loop
@@ -134,6 +150,8 @@ type Agent struct {
 	// Callback that allows user to implement custom behavior
 	// for STUN Binding Requests
 	userBindingRequestHandler func(m *stun.Message, local, remote Candidate, pair *CandidatePair) bool
+	// Callback that allows user to inspect or modify outbound STUN messages.
+	stunSendHandler STUNSendHandler
 
 	gatherCandidateCancel func()
 	gatherCandidateDone   chan struct{}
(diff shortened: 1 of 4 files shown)


agent_options.go
```diff
@@ -789,6 +789,20 @@ func WithBindingRequestHandler(
...
+	return func(a *Agent) error {
+		if a.constructed {
+			return ErrAgentOptionNotUpdatable
+		}
+
+		a.bindingSuccessXORMappedAddressHandler = handler
+
+		return nil
+	}
+}
```
JoTurk: I'm sorry, but unless we have to, I don't think we should add ad-hoc narrower callbacks like this to ICE anymore especially that we're going with a refactor soon, maybe we make another version of `WithBindingRequestHandler` and / or bake proxy/address logic inside ICE?

JoTurk (changes requested):

agent_options.go
Sean-Der: @JoTurk What do you think of something like `BindingResponseHandler`? Or maybe even more general, let someone override `sendSTUN`?

I agree this is such a unique use case it's only going to have one or two users at most. How can we make this general enough that it benefits everyone (and their unique use cases). 

`STUNSendHandler` maybe?

JoTurk: > What do you think of something like BindingResponseHandler?

Yeah this is what i meant by "another version of WithBindingRequestHandler" (we added it in the new options API) https://github.com/pion/ice/blob/9acc9f60f0468caab765821e396b594819898b47/agent_options.go#L778
Maybe we just need an enhanced version with more generic control and more data? Then in the next major release (should be soon for batching and single write path) we can get rid of the current one. Or STUNSendHandler yeah. 

> I agree this is such a unique use case it's only going to have one or two users at most.

Also another reason, it's hard to keep small features from breaking, and it's a lot easier to support generic APIs that many users use, even better if we use it in webrtc.

agent.go
```diff
@@ -1420,7 +1438,7 @@ func (a *Agent) sendBindingRequest(msg *stun.Message, local, remote Candidate) {
 	a.sendSTUN(msg, local, remote)
 }
 
-func (a *Agent) sendBindingSuccess(m *stun.Message, local, remote Candidate) {
+func (a *Agent) sendBindingSuccess(request *stun.Message, local, remote Candidate) {
 	base := remote
```
JoTurk, line 1442: I think we should do it closer to the actual send path, instead of just `sendBindingSuccess`.

PR #909, opened 2026-04-14
Follow RFC 8445 to skip connectivity checks in lite mode

### LINKED ISSUE #96: Implement ICE Lite controlling agent
ICE Lite may be controlling when the other agent also implements ICE Lite. This is a very uncommon case. See See RFC 8445 S6.1.1 and S6.2.

erivni:
#### Description
Lite ICE agents are STUN-server-only — they must not generate connectivity checks ([RFC 8445 §7](https://www.rfc-editor.org/rfc/rfc8445.html#section-7)). This change fixes two gaps in the current lite mode implementation for the full + lite agent scenario:

* Suppress triggered checks (controlledSelector.HandleBindingRequest): lite agents no longer call PingCandidate in response to an inbound Binding Request, per [RFC 8445 §7](https://www.rfc-editor.org/rfc/rfc8445.html#section-7)
* Direct nomination (controlledSelector.HandleBindingRequest): a lite controlled agent accepts USE-CANDIDATE and sets the selected pair immediately, without requiring the pair to have first reached Succeeded state via a triggered check that was never sent, per [RFC 8445 §7.3.2](https://www.rfc-editor.org/rfc/rfc8445.html#section-7.3.2).
Tests added for both behaviours.

> The both-lite controlling path (liteSelector.ContactCandidates) is not changed here and remains tracked in #96.

#### Reference issue
related to #96
See also #907


selection_test.go
```diff
@@ -1491,3 +1491,182 @@ func TestControllingSideRenomination(t *testing.T) {
...
+			stun.NewShortTermIntegrity(agent.localPwd),
+			stun.Fingerprint,
+		)
+		require.NoError(t, err)
+
+		ls.HandleBindingRequest(msg, local, remote)
+
+		assert.Equal(t, pair, agent.getSelectedPair(),
+			"lite controlled agent must accept nomination and set selected pair")
+		// Still no triggered check emitted
```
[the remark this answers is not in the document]
erivni: done

JoTurk (approved):
looks good, thank you. 
I'll try to look at the rest of the code again, before we merge or tag this to make sure we don't introduce bugs.

selection.go
```diff
@@ -449,11 +449,10 @@ func (s *controlledSelector) HandleBindingRequest(message *stun.Message, local,
 			return
 		}
 
-		if pair.state == CandidatePairStateSucceeded {
-			// If the state of this pair is Succeeded, it means that the check
-			// previously sent by this pair produced a successful response and
-			// generated a valid pair (Section 7.2.5.3.2).  The agent sets the
-			// nominated flag value of the valid pair to true.
+		if pair.state == CandidatePairStateSucceeded || s.agent.lite {
```
JoTurk: This leaves lite pair as Waiting while still selecting it. I'll push a commit to fix this.

JoTurk (changes requested):

selection.go
JoTurk: fixed

JoTurk:
I also updated the PR description as this doesn't solve #96

JoTurk:
Found another bug with custom BindingRequestHandler where we don't switch the state from waiting when we select a pair.

I pushed another commit to fix it.

JoTurk (approved):
Looks good now, I tested this manually.

PR #914, opened 2026-04-24
Make it possible to conditionally insert attributes

fippo:
which will be used in #876 where the presence of some
attributes depends on whether they were used in the
original request.

(split from that to make review simpler, no functional changes)

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1574,15 +1574,20 @@ func (a *Agent) sendBindingSuccess(m *stun.Message, local, remote Candidate) {
 		return
 	}
 
-	if out, err := stun.Build(m, stun.BindingSuccess,
+	attributes := []stun.Setter{
+		m,
+		stun.BindingSuccess,
 		&stun.XORMappedAddress{
 			IP:   ip.AsSlice(),
 			Port: port,
 		},
+	}
+	attributes = append(attributes,
 		stun.NewShortTermIntegrity(a.localPwd),
-		stun.Fingerprint,
-	); err != nil {
-		a.log.Warnf("Failed to handle inbound ICE from: %s to: %s error: %s", local, remote, err)
+		stun.Fingerprint)
+
+	if out, err := stun.Build(attributes...); err != nil {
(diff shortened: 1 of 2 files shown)


agent.go
```diff
@@ -1574,14 +1574,19 @@ func (a *Agent) sendBindingSuccess(m *stun.Message, local, remote Candidate) {
...
 		},
+	}
+	attributes = append(attributes,
 		stun.NewShortTermIntegrity(a.localPwd),
-		stun.Fingerprint,
-	); err != nil {
+		stun.Fingerprint)
+
+	if out, err := stun.Build(attributes...); err != nil {
 		a.log.Warnf("Failed to handle inbound ICE from: %s to: %s error: %s", local, remote, err)
```
fippo: This log is a bit odd since it is an error trying to build a STUN message, not an eror handling an inbound one? From the next chunk below the right thing should be `s.log.Error(err, toError)`?

JoTurk: +1

JoTurk (approved):

agent.go
```diff
@@ -1574,15 +1574,20 @@ func (a *Agent) sendBindingSuccess(m *stun.Message, local, remote Candidate) {
...
+	}
+	attributes = append(attributes,
 		stun.NewShortTermIntegrity(a.localPwd),
-		stun.Fingerprint,
-	); err != nil {
-		a.log.Warnf("Failed to handle inbound ICE from: %s to: %s error: %s", local, remote, err)
+		stun.Fingerprint)
+
+	if out, err := stun.Build(attributes...); err != nil {
+		a.log.Error(err.Error())
```
JoTurk: maybe we should include some details? e.g Failed to build stun attributes: %w etc

JoTurk: This is because we currently don't have tracing (Something we should work on) so it's really hard to track or debug errors without some details...

fippo: Updated at least in all the places in the diff

JoTurk: thank you :)

PR #916, opened 2026-04-28
[codex] Fix renomination attribute ordering

### LINKED ISSUE #915: renomination adds attributes after message-integrity and fingerprint
Spotted by @zshang-oai while looking at #876 

sendNominationRequest [here](https://github.com/pion/ice/blob/main/agent.go#L2049) is adding the renomination attributes *after* M-I and fingerprint have been added to the message (which should lead to a programming error as stated [here](https://webrtc-review.googlesource.com/c/src/+/466000/2)

This is a violation of the STUN spec and means renomination is not covered by M-I or fingerprint.
This does lead to test failures after https://github.com/pion/stun/pull/273 which is not tagged yet and will have to be done more carefully.
This might be a good excuse to change the ice-option and attribute type @juberti?

Fixing the *order* in renomination should be safe though. @zshang-oai do you want to make a PR?

zshang-oai:
## Summary

- move the optional renomination attribute ahead of `MESSAGE-INTEGRITY` and `FINGERPRINT`
- assert the serialized outbound STUN order in the existing renomination test

## Why

`sendNominationRequest` appended the renomination attribute after `MESSAGE-INTEGRITY` and `FINGERPRINT`, which leaves the attribute outside the integrity-protected portion of the STUN message and violates the expected ordering.

Fixes #915.

## Testing

- `go test ./... -run TestSendNominationRequest -count=1`

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -2042,8 +2042,6 @@ func (a *Agent) sendNominationRequest(pair *CandidatePair, nominationValue uint3
 		UseCandidate(),
 		AttrControlling(a.tieBreaker),
 		PriorityAttr(pair.Local.Priority()),
-		stun.NewShortTermIntegrity(a.remotePwd),
-		stun.Fingerprint,
 	}
 
 	// Add nomination attribute if renomination is enabled and value > 0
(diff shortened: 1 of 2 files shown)


JoTurk (approved):

fippo (commented):
🍰

PR #917, opened 2026-04-28
Reject oversized DTLS-in-STUN ACK lists

zshang-oai:
## Summary

- cap `DTLS-IN-STUN-ACK` decoding at the same four ACK values accepted by encoding
- derive the byte limit from the value count so the two limits stay in sync
- add a regression case for the maximum valid ACK attribute size

## Why

`DtlsInStunAckAttribute.AddTo` rejects more than four `uint32` ACK values, but `GetFrom` currently allows up to 32 bytes. Since each ACK is four bytes, that accepts eight ACK values on receive. This makes the inbound parser more permissive than the outbound writer and allows oversized ACK lists through.

## Testing

- `go test -run TestDtlsInStunAckAttribute -count=1 -timeout 30s .`

### DIFF
--- a/sped.go
+++ b/sped.go
@@ -35,8 +35,11 @@ func (d *DtlsInStunAttribute) GetFrom(m *stun.Message) error {
 // of DTLS packets (embedded in STUN or without embedding).
 type DtlsInStunAckAttribute []uint32
 
-// Acks are 32 bit values, the attribute can carry up to four of these.
-const ackSizeBytes, ackSizeValues = 32, 4
+// ACKs are 32-bit values, and the attribute can carry up to four of them.
+const (
+	ackSizeValues = 4
+	ackSizeBytes  = ackSizeValues * 4
+)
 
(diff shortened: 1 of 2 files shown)


JoTurk (commented):
(If the draft doesn't enforce this cap, then we prob shouldn't enforce it, we try to follow RFC-9413 I would recommend enforcing this cap in the spec first).

zshang-oai:
@JoTurk 

[draft-hancke-webrtc-sped](https://www.ietf.org/archive/id/draft-hancke-webrtc-sped-00.html) §3.3.2.2 says implementations SHOULD cap the number of uint32 ACK entries and recommends a cap of 4. But §3.3.3's MTU table sizes DTLS-IN-STUN-ACK as 4-20 bytes under that recommended cap. 3.3.3 says MUST so I'm not super sure, I will let @juberti and @philipch07 chime in

fippo (approved):
the intent was (and is matched by libWebRTC) 4 uint32 which is 4*4 = 16 bytes, not 32

Good catch!

PR #918, opened 2026-05-08
Return ErrNoCandidatePairs before ICE connects

zshang-oai:
## Summary

Return `ErrNoCandidatePairs` from `Conn.Write` when ICE has not selected a pair yet and there is no valid fallback pair available.

## Why

`Conn.Write` previously returned `(0, nil)` in that case because the task-loop lookup itself succeeded even though no packet could be written. That makes an early caller believe the write was accepted even though zero bytes were sent.

## Changes

- return `ErrNoCandidatePairs` when `Conn.Write` has no selected or valid pair
- add a regression test for writes immediately after `StartDial`
- update affected retry-style tests to treat the pre-connection error as transient
- update the binding-request test to assert the explicit no-pair failure

## Tests

- `go test -run '^TestStartDialConnWriteBeforeConnectReturnsError$' -count=1 -timeout 120s .`
- `go test -run '^TestBindingRequestHandler$' -count=1 -timeout 120s .`
- `go test -run '^TestWriteUseValidPair$' -count=1 -timeout 120s .`
- `go test -run '^TestConn_Write_RejectsSTUN$' -count=1 -timeout 120s .`


JoTurk (changes requested):
Hello, I agree with this change, but I'm worried that it might break something at the end user, or at pion/webrtc, so I'm going to soft-block this until i get to manually test pion/webrtc with this change, (feel free to test it instead).
It shouldn't break pion/webrtc because we do handle this error at the mux writer https://github.com/pion/webrtc/blob/9654161f9b698b2d9a2df3dd5a676b20e881a7b3/internal/mux/endpoint.go#L59 but you can't really know for sure sometimes :)

zshang-oai:
@fippo and cc @JoTurk, the mux behavior mentioned above will become a bug after WARP.

If happens when

- DTLS finishes from piggybacked handshake traffic
- ICE has not yet published a selected pair, or has no best valid pair available to ordinary writes
- SCTP/SNAP immediately starts writing post-handshake traffic

And due to SNAP, we consider data channel to be open immediately and the mutex will swallow the wrote silently because ICE is actually not writable. 

We will fix the bug for our scenario but if you agree, we are happy to upstream it. (I guess @JoTurk  want to be more safe :) )

JoTurk:
@zshang-oai Thank you for the report and fix, I just want to make sure that weren't breaking our normal path in webrtc, our dtls handshake/fsm is fragile and you can never know. We had a similar change few months ago that broke fragmented dtls handshake. I'll try to test this with pion/webrtc soon. If you can test it yourself (or maybe codex) that would be also cool.

JoTurk (approved):
Thank you. Sorry I lost track of this, there is a small change of behavior for direct ICE users, but it doesn't break webrtc (i tested and checked manually).
And I think this is the correct direction.
I'll merge this once the CI passes.

PR #920, opened 2026-05-18
[codex] Allow ICE-lite SPED response before nomination

zshang-oai:
## What changed

- Treat an authenticated inbound BindingRequest as a valid pre-nomination pair for ICE-lite controlled agents only while DTLS-in-STUN piggybacking is active.
- Keep the existing non-SPED ICE-lite behavior unchanged.
- Add regression coverage for both the SPED and non-SPED paths.

## Why

In ICE-lite controlled mode, the agent does not have a successful outbound connectivity check before nomination. With DTLS-in-STUN piggybacking, that is a problem when the controlled side is also the DTLS client: its first DTLS handshake flight may need to be returned on the BindingResponse that precedes `USE-CANDIDATE`.

Previously, the pair was still not considered valid at that point, so writes that depended on a valid pair could not use it before nomination. That can stall the handshake before nomination has a chance to arrive.

This change narrows the exception to the piggybacking window only. Normal ICE behavior still waits for the usual nomination flow.

## Validation

- `GOCACHE=/private/tmp/ice-sped-gocache go test ./... -run 'TestLiteControlledSelector_(SpedInboundCheckMakesPairWritableBeforeNomination|NonSpedInboundCheckDoesNotMakePairWritableBeforeNomination)$' -count=1`
- `GOCACHE=/private/tmp/ice-sped-gocache go test ./... -run 'TestSped|TestControlledSelector_(NoTriggeredCheckAfterConnected|TriggeredCheckDuringChecking)$' -count=1`

Note: a full `go test ./...` run still times out in `TestMulticastDNSOnlyConnection` in this local environment because mDNS socket setup is unavailable; that appears unrelated to this change.


zshang-oai:
Ah the branch is dirty, fixing

zshang-oai:
Opened against the wrong base branch; superseded by a clean one-commit branch in the SPED fork.

PR #921, opened 2026-05-19
Avoid triggered checks from ICE-lite agents

zshang-oai:
## Summary
- add an ICE-lite `HandleBindingRequest` path that responds without originating triggered checks
- mark accepted `USE-CANDIDATE` / nomination requests as succeeded before reusing the controlled nomination handling
- add a regression covering accepted ICE-lite nomination without an outbound triggered check

## Why
ICE-lite agents act only as STUN servers. Reusing the full controlled-agent request path caused the lite side to originate a triggered connectivity check, adding an unnecessary RTT before the pair became selected. For accepted nomination requests, RFC 8445 section 7.3.2 lets the lite side treat the pair as valid/nominated from the inbound request path instead.

## Validation
- `go test ./... -run 'TestLiteSelector_NoTriggeredCheckForAcceptedNomination|TestControlledSelector_TriggeredCheckDuringChecking|TestControlledSelector_NoTriggeredCheckAfterConnected' -count=1`
- attempted `go test ./...`; it timed out in `TestMulticastDNSOnlyConnection` in this local environment after about 117s

### DIFF
--- a/selection.go
+++ b/selection.go
@@ -424,6 +424,15 @@ func (s *controlledSelector) HandleSuccessResponse(m *stun.Message, local, remot
 }
 
 func (s *controlledSelector) HandleBindingRequest(message *stun.Message, local, remote Candidate) { //nolint:cyclop
+	s.handleBindingRequest(message, local, remote, true)
+}
+
+//nolint:cyclop
+func (s *controlledSelector) handleBindingRequest(
+	message *stun.Message,
+	local, remote Candidate,
+	pingCandidate bool,
+) {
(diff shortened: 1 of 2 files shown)


JoTurk:
@zshang-oai There is an already a PR to solve this issue https://github.com/pion/ice/pull/909 I just need to test it more, We also need to be careful because we don't handle ice-lite controlling mode yet.

zshang-oai:
@JoTurk Oh did not realize that, sorry and thank you

PR #922, opened 2026-05-21
Support append rewrite rule for tcpmux

cnderrauber:
Support append rewrite rule for tcpmux.

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -332,6 +332,13 @@ func (a *Agent) gatherCandidatesLocal(ctx context.Context, networkTypes []Networ
 		return
 	}
 
+	// Detect duplicate candidate configs before calling addCandidate(), since
+	// TCPMux returns the same underlying tcpPacketConn for multiple mapped IPs
+	// that share a real local IP. Without this, addCandidate() would treat the
+	// second registration as a duplicate and close the shared conn,
+	// invalidating the first candidate.
+	existingConfigs := make(map[CandidateHostConfig]struct{})
+
 	for _, info := range localAddrs {
 		addr := info.addr
 		ifaceName := info.iface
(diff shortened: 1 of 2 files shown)


gather.go
```diff
@@ -357,12 +364,6 @@ func (a *Agent) gatherCandidatesLocal(ctx context.Context, networkTypes []Networ
 			}
 
 			for network := range networks {
-				// TCPMux maintains a single listener per interface. Avoid duplicating passive TCP candidates
```
cnderrauber, line 360: @JoTurk You mentioned that tcpmux does not support append rules because connections cannot be shared between candidates. Fom what I can see, tcppacketconn seems to support this without issues, and the test cases also confirm it works.

Is there any edge case I might be missing?

boks1971 (commented):
lgtm! but my understanding is weak here, would be great to get more 👀 on this.

gather.go
JoTurk: This is done because addcandidate is not safe with shared connection, and will close connections and report duplicated address. As you wrote in your comment.

It's not safe to change this without having a different ownership and duplication handling for mux-based and shared addresses  (planned for the next ICE major version), or at least we soften by filtering out candidates that will cause addCandidate to close the shared connection, your pr is close, but you shouldn't match with CandidateHostConfig it should match addCandidate duplication handling and use candidate tuple.

(We don't support append in UDP mux too).

sorry forgot to /cc  @cnderrauber @boks1971

JoTurk:
(Also we prob should change this PR title as this doesn't add support for append for tcpmux in general, it just makes it possible if the address tuple is unique, assuming you change the CandidateHostConfig map to a candidate tuple).

gather.go
cnderrauber: Makes sense to have a shared connection for append rules, stops working on this pr.  I will look at muxes to draft a new pr for this feature.

JoTurk: Sounds good, we eventually will need to tell ice that some sockets are owned by the user, and the user handles the duplication and the close logic, but this isn't possible in a clean way without breaking the api, a shared connection is good enough for now I think.

PR #923, opened 2026-05-26
Support append rewrite rule for UDP/TCP mux

cnderrauber:
Wrap the conn returned by UDP/TCP mux GetConn in a ref-counted sharedPacketConn so AddressRewriteAppend can hand out multiple alias host candidates that share a single underlying mux conn. The underlying conn is closed only when the last wrapper is released. Close mux conn in error paths in gather to avoid leak.

Close #922

### DIFF
--- a/shared_packet_conn.go
+++ b/shared_packet_conn.go
@@ -1,0 +1,132 @@
+// SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package ice
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net"
+	"os"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+type muxedPacketConn interface {
+	net.PacketConn
+	readFromContext(ctx context.Context, b []byte) (int, net.Addr, error)
+}
+
+// sharedPacketConn is a reference-counted wrapper around an underlying
+// muxedPacketConn owned by a mux. The mux can hand out several wrappers that
+// share one underlying connection (for example, alias host candidates
+// produced by an AddressRewrite append rule).
+//
+// Each wrapper owns its own context. Close() cancels that context, which
+// unblocks any in-flight ReadFrom on this wrapper — without affecting
+// siblings that still hold their own references. The underlying connection
+// itself is closed only when the last wrapper is released.
+type sharedPacketConn struct {
+	underlying muxedPacketConn
+	refs       *atomic.Int32
+
+	ctx       context.Context //nolint:containedctx
+	cancel    context.CancelFunc
+	closeOnce sync.Once
+
+	readDeadline atomic.Pointer[time.Time]
+}
+
+// newSharedPacketConn increments the shared refcount and returns a wrapper.
+// Each returned wrapper must have Close called exactly once.
+func newSharedPacketConn(u muxedPacketConn, refs *atomic.Int32) *sharedPacketConn {
+	refs.Add(1)
+	ctx, cancel := context.WithCancel(context.Background())
+
+	return &sharedPacketConn{
+		underlying: u,
+		refs:       refs,
+		ctx:        ctx,
+		cancel:     cancel,
+	}
+}
+
+func (s *sharedPacketConn) ReadFrom(b []byte) (int, net.Addr, error) {
+	ctx := s.ctx
+	if ctx.Err() != nil {
+		return 0, nil, io.ErrClosedPipe
+	}
+
+	if p := s.readDeadline.Load(); p != nil && !p.IsZero() {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithDeadline(s.ctx, *p)
+		defer cancel()
+	}
+
+	n, addr, err := s.underlying.readFromContext(ctx, b)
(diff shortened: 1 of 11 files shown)


shared_packet_conn.go
```diff
@@ -0,0 +1,132 @@
...
+
+	return nil
+}
+
+func (s *sharedPacketConn) SetWriteDeadline(t time.Time) error {
+	if s.ctx.Err() != nil {
+		return io.ErrClosedPipe
+	}
+
+	return s.underlying.SetWriteDeadline(t)
```
boks1971, line 106: should this be per context too?

shared_packet_conn.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (s *sharedPacketConn) Close() error {
+	var err error
+	fired := false
+	s.closeOnce.Do(func() {
+		fired = true
+		s.cancel()
+		if s.refs.Add(-1) <= 0 {
+			err = s.underlying.Close()
```
boks1971, line 124: can this race? What happens when there is a `newSharedConn` call happening at the same time?

boks1971 (commented):
lgtm! but in the same boat still as last review, don't know the full depth of the nuances here, would be good to get more 👀 . 

Some small comments.

shared_packet_conn.go
cnderrauber: It is hard to add a context for writing.. There are two options:
* add a write context in sharedPacketConn and use a goroutine for every write call to wait for the underlying write or context timeout.
* implements writeWithContext in the underlying connection, also uses a goroutine for every write call.

Neither is ideal as the cost is not negligible and `SetWriteDeadline` of internal mux connection is not used in this repo or pion stack, so left it here.

shared_packet_conn.go
cnderrauber: The gather will get/create all mux conn first then close them if addcandidate failed later or agent is closed, so the case should not happen..

udp_muxed_conn.go
```diff
@@ -89,6 +98,8 @@ func (c *udpMuxedConn) ReadFrom(b []byte) (n int, rAddr net.Addr, err error) {
 		case <-c.notify:
 		case <-c.closedChan:
 			return 0, nil, io.EOF
+		case <-ctx.Done():
+			return 0, nil, ctx.Err()
 		}
 	}
 }
```
JoTurk, line 106: Correct me if I'm wrong but udpMuxedConn will still fail with two aliases and create a race, if two aliases with their recive loop in sharedPacketConn  that wraps udpMuxedConn are waiting, one reader can change the state from waiting to open and receive a notify while the other reader get stuck on an open connection without ever receiving a notify?

JoTurk (commented):
Code looks good to me, thank you, just one concern.

udp_muxed_conn.go
cnderrauber: Not that bad that candidate_base uses a recvloop to call conn.Read, so there will always be one alias's read func change the udpMuxedConn's state to waiting, and no data is lost. I have changed the state to a waiting counter in https://github.com/pion/ice/pull/923/commits/441b2c9d19773cb2841e80f6af4eda9e044c552e to make it looking better.

udp_muxed_conn.go
```diff
@@ -76,19 +79,28 @@ func (c *udpMuxedConn) ReadFrom(b []byte) (n int, rAddr net.Addr, err error) {
...
 
 			return 0, nil, io.EOF
 		}
 
-		c.state = udpMuxedConnWaiting
+		c.readWaiting.Add(1)
 		c.mu.Unlock()
 
 		select {
 		case <-c.notify:
```
JoTurk, line 92: Super nit: this can case a super rare race with two or more readers since the channel is buffered to only one slot, and if the notifiy isn't installed yet, the readers can miss notify if the writer send it in this gap. I think if this happens it will be super rare and not worth blocking this over it, but it's prob worth fixing it in the future.

JoTurk (approved):
Thank you.

udp_muxed_conn.go
cnderrauber: yeah, that's what I mean `better` in the change, it still has a very low chance to miss notify and buffer the packet as you said, but the next read in the candidate's recvloop will get the packet quickly. Not perfect in semantic of net.PacketConn's Read, but works in this repo.

cnderrauber:
MDNS test in macOS failed constantly in this pr, but it should be unrelated to the change. I also see failure when rerunning the test in the main branch https://github.com/pion/ice/actions/runs/25922078986/job/77831961162, guess environment of the runner cause the failure.

JoTurk:
@cnderrauber 
> MDNS test in macOS failed constantly in this pr, but it should be unrelated to the change. I also see failure when rerunning the test in the main branch https://github.com/pion/ice/actions/runs/25922078986/job/77831961162, guess environment of the runner cause the failure.

These failures are related to the recent macOS CI upgrade, which introduced changes to macOS network permissions. https://github.com/pion/.goassets/pull/277 

We have been fixing the tests across pion, but nothing is actually broken in the libraries them-self

https://github.com/pion/webrtc/pull/3429
https://github.com/pion/mdns/pull/275

PR #926, opened 2026-06-02
Fix UDPMuxDefault not reflecting network changes

aler9:
#### Description

UDPMuxDefault was internally caching the list of local IPs, making connectivity impossible after a network change or a local IP change. This was also redundant given the fact that IPs and interfaces are already cached by their provider, which is stdnet.Net / transport.Net.

This patch removes this cache layer, allowing UDPMuxDefault to receive upstream interface and IP changes, and fixing support for network changes.

#### Reference issue

https://github.com/bluenviron/mediamtx/issues/5097

### DIFF
--- a/udp_mux.go
+++ b/udp_mux.go
@@ -44,8 +44,8 @@ type UDPMuxDefault struct {
 
 	mu sync.Mutex
 
-	// For UDP connection listen at unspecified address
-	localAddrsForUnspecified []net.Addr
+	// whether the UDP connection listens on an unspecified address
+	isUnspecified bool
 }
 
 // UDPMuxParams are parameters for UDPMux.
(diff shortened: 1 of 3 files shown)


udp_mux.go
```diff
@@ -61,50 +61,24 @@ type UDPMuxParams struct {
...
 		params.Logger = logging.NewDefaultLoggerFactory().NewLogger("ice")
 	}
 
-	var localAddrsForUnspecified []net.Addr
-	if udpAddr, ok := params.UDPConn.LocalAddr().(*net.UDPAddr); !ok { //nolint:nestif
+	var isUnspecified bool
+	if udpAddr, ok := params.UDPConn.LocalAddr().(*net.UDPAddr); !ok {
 		params.Logger.Errorf("LocalAddr is not a net.UDPAddr, got %T", params.UDPConn.LocalAddr())
 	} else if ok && udpAddr.IP.IsUnspecified() {
 		// For unspecified addresses, the correct behavior is to return errListenUnspecified, but
```
JoTurk, line 73: There is another cache layer in pion/transport that caches the network interfaces https://github.com/pion/transport/blob/1bb98ade0d7759304c0b88bd33914b42bc7f481b/stdnet/net.go#L70-L72 so localInterfaces will just return the same cached interfaces.

maybe we should share the refresh logic with nomination  https://github.com/pion/ice/blob/bd194f61218e2d59923f53b57fb269fdfdff4e3b/gather.go#L1402-L1405 and call updateInterfaces?

JoTurk (commented):
The idea of removing the cache interface looks good to me, maybe we need to make a small helper for detecting network changes across different systems in pin/transport, so we don't have to relay on library logic or polling, this was the plan for nomination anyways.


@cnderrauber this reverts some of https://github.com/pion/ice/commit/a931199dc1d1d43761c8283fa5c9dda18f980b93 does it look good for you?

aler9:
@JoTurk i did an [assessment](https://github.com/bluenviron/mediamtx/issues/5097#issuecomment-4215089513) on the network change issue some time ago that confirms what you wrote about the two separate cache layers.

As you mentioned there's also a refresh mechanism that is unfortunately placed in the wrong component, which is the agent inside each `PeerConnection`, while it should be shared among every component.

In my opinion, the right place in which to put the refresh mechanism is inside a shared instance of `stdnet.Net` / `transport.Net`, and that's what i did in my [overall fix](https://github.com/bluenviron/mediamtx/pull/5818) - i didn't contribute it because it's a breaking change.

Anyway, a common step to any solution is to remove the cache layer inside `UDPMuxDefault` (which is this PR).

cnderrauber (approved):
Looks good!!

udp_mux.go
```diff
@@ -137,8 +111,37 @@ func (m *UDPMuxDefault) LocalAddr() net.Addr {
...
+		case udpAddr.IP.To4() != nil:
+			networks = []NetworkType{NetworkTypeUDP4}
+		case udpAddr.IP.To16() != nil:
+			networks = []NetworkType{NetworkTypeUDP4, NetworkTypeUDP6}
+		default:
+			return []net.Addr{m.LocalAddr()}
+		}
+
+		_, addrs, err := localInterfaces(m.params.Net, nil, nil, networks, true)
+		if err != nil {
```
cnderrauber, line 133: Add a warning log on error?

aler9: thanks for the review, i added log messages in order to report errors that would have otherwise be completely lost since GetListenAddresses doesn't return errors.

JoTurk (approved):
thank you @aler9

PR #938, opened 2026-07-11
Add pre-insertion local candidate filter

xojigsx:
Adds a generic LocalCandidateFilter agent option invoked after duplicate detection and before candidate startup, insertion, pair creation, connectivity checks, or notification. Rejected candidates and their connections are closed without surfacing a gather error. Tests cover accepted and rejected host, server-reflexive, and relay candidates, cleanup, notification/pair suppression, and duplicate behavior. Full and focused race suites pass.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -151,10 +151,11 @@ type Agent struct {
 	udpMux      UDPMux
 	udpMuxSrflx UniversalUDPMux
 
-	interfaceFilter func(string) (keep bool)
-	ipFilter        func(net.IP) (keep bool)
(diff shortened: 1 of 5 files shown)


JoTurk (changes requested):
no-host/srflx use case is already covered by ICETransportPolicyNoHost and WithCandidateTypes. A generic post-construction candidate filter would need a concrete use case and it seems to be a footgun to filter candidates this late.

xojigsx:
@JoTurk took a second look and you're right - I can use ICETransportPolicyNoHost with WithCandidateTypes - thanks!

PR #939, opened 2026-07-13
Ignore malformed nomination attributes

JoTurk:
#### Description
pion/ice had this bug for a while where it enter controlled nomination handling when we receive a malformed nomination-only binding request even without use-candidate. This isn't a big issue because we used connectivity checks to promote waiting candidates, but after https://github.com/pion/ice/pull/909 we don't do checks for LITE and we promote them directly to Succeeded, and this can cause problems with non standard ICE clients or clients that send malformed nominations.
I found this by looking at the selection code while reviewing and fixing #909 and it deserved its own PR / commit because it wasn't directly related.

### DIFF
--- a/selection.go
+++ b/selection.go
@@ -445,16 +445,20 @@ func (s *controlledSelector) HandleBindingRequest(message *stun.Message, local, 
 	}
 	pair.UpdateRequestReceived()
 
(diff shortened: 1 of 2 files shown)


kevmo314 (approved):

JoTurk:
@kevmo314 thank you for the fast review :)

Issue #697, opened 2024-05-09
Doubts about discarding monotonic time


jech:
Hi,

I'm looking at @paulwe optimisations to pion/ice right now, thanks a lot for the work, Paul.

I have a doubt about commit e1c2d85530f0f9e96935c3598cc68b6c596c1a39, where the `lastSent` and `lastReceived` fields of `candidateBase` have been changed from `time.Time` to Unix time.  In doing so, they no longer contain monotonic time, and therefore become susceptible to stepping clocks.

Paul, can you confirm that this change will not cause trouble if the system clock is stepped?  If not, then I suggest that either the commit should be reverted, or we should use a technique similar to what I do in https://github.com/jech/storrent/blob/master/mono/mono.go, where I had exactly the same problem (albeit with just second granularity).

paulwe:
in practice i don't think it matters. on any well behaved system the aberrations in wall clock time should be smaller than the tolerance for timeout/keepalive scheduling. 

because it's part of the public api we can't really do anything more clever. go ahead and revert it if you like. none of this had a measurable impact on my target metric.

jech:
> in practice i don't think it matters. on any well behaved system the aberrations in wall clock time should be smaller than the tolerance for timeout/keepalive scheduling.

What happens if the system clock is stepped?

Some systems have no battery-backed clock.  On such systems, the system clock is stepped after NTP has had time to converge, typically  a couple of minutes after boot.  If I understand your patch correctly, it will cause Pion to break after the clock is stepped.

> none of this had a measurable impact on my target metric.ou

Did you test while stepping the clock?

> go ahead and revert it if you like

@Sean-Der could you please consider PRs #698 and #699?  They unbreak Pion on systems with no battery-backed clock (such as Raspberry Pis and other SBCs).

paulwe:
sorry i thought you had commit or i would have done it myself. like i said, this didn't fix the issue i was trying to solve so i'm not super attached to it.

jech:
Thanks!

PR #941, opened 2026-07-15
Reduce read and write allocations

kcaffrey:
Remove two per-packet write allocations and one per-packet read allocation:
 - Avoid boxing `time.Time` for last packet sent/received
 - Avoid heap allocating `pair` for a closure declared inside a fallback path

The pattern established by #851 is used to maintain monotonic time while avoiding boxing `time.Time`. See also #690, #697, and #699 for more history on this pattern.

A test is added to pin allocations to zero for the write path, at least outside of `net.PacketConn`. The full write path still does allocate inside of `(net.PacketConn).WriteTo`. `net.UDPConn` has `WriteToUDPAddrPort` which doesn't allocate, but supporting that is a larger change that I plan on chipping away at later. In the meantime, the test uses a mock `PacketConn` so that it can verify that everything else is zero alloc.

The closure-related heap allocation took me a while to hunt down and (to me) is very non-obvious, so I put an overly-large comment explaining the oddity. That plus the test verifying zero allocations is hopefully enough to prevent future developers altering that code from introducing an allocation regression.

The benchmark tests the full end-to-end path with a real connection, and demonstrates a reduction from 5 allocs/op to 2 allocs/op for the write and paired read. The remaining 2 allocs are both in PacketConn, and the fix for both would be to use the `netip.AddrPort` variant.

|  | wall time | bytes | allocs |
|---|---|---|---|
| main | 15694 ns/op | 108 B/op | 5 allocs/op |
| this PR | 15798 ns/op | 52 B/op | 2 allocs/op |

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1167,8 +1167,8 @@ func replacePairRemote(pair *CandidatePair, remote Candidate) *CandidatePair {
 	atomic.StoreUint64(&replacement.responsesReceived, atomic.LoadUint64(&pair.responsesReceived))
 	atomic.StoreUint64(&replacement.responsesSent, atomic.LoadUint64(&pair.responsesSent))
 
-	copyAtomicValue(&replacement.lastPacketSentAt, &pair.lastPacketSentAt)
-	copyAtomicValue(&replacement.lastPacketReceivedAt, &pair.lastPacketReceivedAt)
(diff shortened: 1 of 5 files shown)


JoTurk (approved):
Thank you, I'll need to test this a bit before i merge it

PR #942, opened 2026-07-15
Fix gathering state wedged by restart mid-gather

maum-meliora:
Restarting an ICE agent while a gathering cycle is still in flight can leave `gatheringState` permanently stuck, after which every subsequent `GatherCandidates()` is rejected with `ErrMultipleGatherAttempted`. It shows up when an agent restarts back-to-back, or restarts right after it begins gathering.

## Root Cause

All of the agent's mutable state is serialized through the task loop, but a gathering cycle's `gatherCandidates` goroutine updates `gatheringState` via `setGatheringState`, which runs as its *own* task. Two orderings leak past a restart:

1. `Restart` cancels the in-flight cycle and resets `gatheringState` to `New`, but that cycle's trailing `setGatheringState(Complete)` task can still run *after* Restart's task and overwrite the fresh `New` with a stale value.
2. `Restart` only canceled the previous cycle while it was already `Gathering`. A gather goroutine that had started but not yet marked `Gathering` was left uncanceled, survived the restart, and set `Gathering` afterwards.

Either way `gatheringState` ends up non-`New`, and the next `GatherCandidates()` fails.

## Fix

- Thread the owning cycle's context into `setGatheringState` and drop the write when that context is canceled, inside the task loop (closes ordering 1) — a superseded cycle can no longer touch `gatheringState`.
- Cancel the previous cycle unconditionally in `Restart` (closes ordering 2).

Both parts are needed: without the first, a canceled cycle's write still lands; without the second, a not-yet-`Gathering` goroutine escapes cancellation. There are no exported API changes, and the now-unconditional `gatherCandidateCancel()` is safe — it is initialized to a no-op and is idempotent.

## Testing

Added `TestAgentRestartThenGatherRepeatedly` (300 back-to-back `Restart` + `GatherCandidates`):

- Fails on `main` with `ErrMultipleGatherAttempted`.
- Passes on this branch, including `-race` and under `GOMAXPROCS=1/2/8`.
- `test.CheckRoutines` confirms no gather-goroutine leak.

```sh
go test -race -run TestAgentRestartThenGatherRepeatedly -count=20 .
```


Sean-Der:
Nice debugging/testing @maum-meliora!

agent.go
```diff
@@ -1983,15 +1984,23 @@ func (a *Agent) Restart(ufrag, pwd string) error { //nolint:cyclop
...
+// cannot clobber the New state Restart set and wedge the next gather.
+func (a *Agent) setGatheringState(gatherCtx context.Context, newState GatheringState) error {
 	done := make(chan struct{})
-	if err := a.loop.Run(a.loop, func(context.Context) {
+	if err := a.loop.Run(a.loop, func(context.Context) { //nolint:contextcheck
+		defer close(done)
+
+		if gatherCtx.Err() != nil {
+			return
+		}
```
JoTurk, line 1998: I don't think it's safe to make setGatheringState  early return like this without change to gatherCandidates https://github.com/maum-meliora/ice/blob/8a441e87ba060831666398b5198fb01bd5222d77/gather.go#L146-L152 because this will trigger `gatherCandidatesInternal` even tho the state didn't change.

JoTurk (changes requested):
Thank you the fix looks alright, we just need to address the side effect.

agent.go
JoTurk: We can return the new state from setGatheringState and return in gatherCandidates if it didn't change. or just return an ok bool.

gather_restart_test.go
```diff
@@ -0,0 +1,36 @@
+// SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
```
JoTurk: can you please move this test to a pre-existing test file?
thank you.

JoTurk (approved):
Thank you.

PR #949, opened 2026-08-04
Require 4/22 ice-chars for local ICE credentials

fippo:
An ice-char carries 6 bits, not 8, so the RFC 8445 randomness
minimums are exactly the RFC 8839 grammar minimums of 4 and 22
characters. The old *8 check accepted a 16-char password carrying
~96 bits against a 128-bit requirement, and values too short to
be legal in SDP.

RFC 5245 description is still the best:
 https://datatracker.ietf.org/doc/html/rfc5245#section-15.4

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1942,11 +1942,8 @@ func (a *Agent) Restart(ufrag, pwd string) error { //nolint:cyclop
 		}
 	}
 
-	if len([]rune(ufrag))*8 < 24 {
-		return ErrLocalUfragInsufficientBits
-	}
-	if len([]rune(pwd))*8 < 128 {
-		return ErrLocalPwdInsufficientBits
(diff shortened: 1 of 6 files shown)


JoTurk (approved):
good catch :)

fippo:
asked Claude about something else, it blurbed too much in response and said something that sounded wrong ;-)

JoTurk:
@fippo should we also validate that it's valid ice-chars? i added a helper a while ago while i was implementing the extensions parser https://github.com/pion/ice/blob/c0c488f5ce9648e4fa9bfc8e267ccb3ac7dc76b6/candidate_base.go#L1003-L1026

fippo:
for remote that does not work since a lot of people insist on base64. We might for local which we  control

PR #950, opened 2026-08-05
Increase timeouts for lite agents

JoTurk:
#### Description
The current 5 second disconnect timeout is not safe for ICE Lite. RFC 7675 recommends sending connectivity checks every 4-6 seconds (which Firefox uses). Pion can consistently mark otherwise healthy ICE Lite connections as disconnected when there is no active traffic (e.g idle data channel connections with no media tracks).

This changes the disconnected timeout to 10s for ICE-lite, the other change is that we change the failed timeout to roughly 35s this fixes a rare issue we observed at my work and I was able to reproduce it manually where a throttled Firefox in the background can push the connectivity interval up to 20-30s which cases pion to mark the connection as failed. (I think it's related to how Firefox schedule timers).

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -91,9 +91,10 @@ type Agent struct {
 
 	candidateTypes []CandidateType
 
-	// How long connectivity checks can fail before the ICE Agent
-	// goes to disconnected
-	disconnectedTimeout time.Duration
+	// How long the selected pair can receive no traffic before the ICE Agent
+	// goes to disconnected.
+	disconnectedTimeout         time.Duration
+	disconnectedTimeoutExplicit bool
 
 	// How long connectivity checks can fail before the ICE Agent
 	// goes to failed
@@ -474,6 +475,8 @@ func newAgentWithConfig(agent *Agent, opts ...AgentOption) (*Agent, error) {
 		}
 	}
 
+	agent.applyICELiteDisconnectedTimeoutDefault()
+
 	agent.connectionStateNotifier = &handlerNotifier{
 		connectionStateFunc: agent.onConnectionStateChange,
 		done:                make(chan struct{}),
@@ -581,6 +584,12 @@ func newAgentWithConfig(agent *Agent, opts ...AgentOption) (*Agent, error) {
 	return agent, nil
 }
 
+func (a *Agent) applyICELiteDisconnectedTimeoutDefault() {
+	if a.lite && !a.disconnectedTimeoutExplicit {
+		a.disconnectedTimeout = defaultLiteDisconnectedTimeout
+	}
+}
+
 func mDNSLocalAddressFromTCPMux(tcpMux TCPMux, networkTypes []NetworkType) net.IP {
 	if tcpMux == nil || !allNetworkTypesTCP(networkTypes) {
 		return nil
@@ -674,6 +683,7 @@ func (a *Agent) startConnectivityChecks(isControlling bool, remoteUfrag, remoteP
 func (a *Agent) connectivityChecks() { //nolint:cyclop
 	lastConnectionState := ConnectionState(0)
 	checkingDuration := time.Time{}
+	checkingTimeout := a.initialCheckingTimeout()
 
 	contact := func() {
 		if err := a.loop.Run(a.loop, func(_ context.Context) {
@@ -692,8 +702,8 @@ func (a *Agent) connectivityChecks() { //nolint:cyclop
 					checkingDuration = time.Now()
 				}
 
-				// We have been in checking longer then Disconnect+Failed timeout, set the connection to Failed
-				if time.Since(checkingDuration) > a.disconnectedTimeout+a.failedTimeout {
+				// The initial checking deadline has elapsed, so set the connection to Failed.
+				if time.Since(checkingDuration) > checkingTimeout {
 					a.updateConnectionState(ConnectionStateFailed)
 
 					return
@@ -746,6 +756,15 @@ func (a *Agent) connectivityChecks() { //nolint:cyclop
 			return
 		}
 	}
+}
+
+func (a *Agent) initialCheckingTimeout() time.Duration {
+	disconnectedTimeout := a.disconnectedTimeout
+	if a.lite && !a.disconnectedTimeoutExplicit {
+		disconnectedTimeout = defaultDisconnectedTimeout
+	}
+
+	return disconnectedTimeout + a.failedTimeout
 }
 
 func (a *Agent) updateConnectionState(newState ConnectionState) {
--- a/agent_config.go
+++ b/agent_config.go
@@ -23,6 +23,10 @@ const (
 	// defaultDisconnectedTimeout is the default time till an Agent transitions disconnected.
 	defaultDisconnectedTimeout = 5 * time.Second
 
+	// defaultLiteDisconnectedTimeout uses the RFC 7675 consent-expiry period as a
+	// conservative receive-idle window for lite Agents.
+	defaultLiteDisconnectedTimeout = 10 * time.Second
+
 	// defaultFailedTimeout is the default time till an Agent transitions to failed after disconnected.
 	defaultFailedTimeout = 25 * time.Second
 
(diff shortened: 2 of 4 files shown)


Sean-Der:
Nice catch! I have seen this for years and never investigated (ICE connection state flapping even in Chrome)

What do you think of setting this in the `selector` instead? I don't know if the code will be better, but I am thinking

* We already have lite/non-lite split up
* I don't want to add a bunch of conditions around the code

Then maybe we can drop the 'was explicitly set' check?

JoTurk:
@Sean-Der I tested with moving it to the selector but I think it will be more code because we need to pass the timeouts through validateSelectedPair and connectionStateForDisconnection. maybe there is a better a way that i'm missing?

disconnectedTimeoutExplicit is needed because it lets us keep the default of 5 + 25s during the initial checks, and 30 + 25s after connection, unless the user sets an explicit value that applies to both. I didn't want to change the initial check to 55s because it would delay failure for connections that can never be established to around one minute, maybe there is a better way to do this too?

JoTurk:
@Sean-Der This is how it looks in the selector and without `disconnectedTimeoutExplicit`, we'll have to pass a lot of timeouts in parameters, and not having  an explicit `disconnectedTimeoutExplicit` doesn't reduce the conditions unless we make it 55s for inintal checks too, If you're fine with this I'll change the PR to this instead:

```diff
diff --git a/agent.go b/agent.go
index 09cd5f98..06a9c5aa 100644
--- a/agent.go
+++ b/agent.go
@@ -91,10 +91,13 @@ type Agent struct {
 
 	candidateTypes []CandidateType
 
-	// How long the selected pair can receive no traffic before the ICE Agent
-	// goes to disconnected.
-	disconnectedTimeout         time.Duration
-	disconnectedTimeoutExplicit bool
+	// How long connectivity checks can fail or a full Agent's selected pair can
+	// receive no traffic before the ICE Agent goes to disconnected.
+	disconnectedTimeout time.Duration
+
+	// How long a lite Agent's selected pair can receive no traffic before the
+	// ICE Agent goes to disconnected.
+	liteDisconnectedTimeout time.Duration
 
 	// How long connectivity checks can fail before the ICE Agent
 	// goes to failed
@@ -475,8 +478,6 @@ func newAgentWithConfig(agent *Agent, opts ...AgentOption) (*Agent, error) {
 		}
 	}
 
-	agent.applyICELiteDisconnectedTimeoutDefault()
-
 	agent.connectionStateNotifier = &handlerNotifier{
 		connectionStateFunc: agent.onConnectionStateChange,
 		done:                make(chan struct{}),
@@ -584,12 +585,6 @@ func newAgentWithConfig(agent *Agent, opts ...AgentOption) (*Agent, error) {
 	return agent, nil
 }
 
-func (a *Agent) applyICELiteDisconnectedTimeoutDefault() {
-	if a.lite && !a.disconnectedTimeoutExplicit {
-		a.disconnectedTimeout = defaultLiteDisconnectedTimeout
-	}
-}
-
 func mDNSLocalAddressFromTCPMux(tcpMux TCPMux, networkTypes []NetworkType) net.IP {
 	if tcpMux == nil || !allNetworkTypesTCP(networkTypes) {
 		return nil
@@ -759,12 +754,7 @@ func (a *Agent) connectivityChecks() { //nolint:cyclop
 }
 
 func (a *Agent) initialCheckingTimeout() time.Duration {
-	disconnectedTimeout := a.disconnectedTimeout
-	if a.lite && !a.disconnectedTimeoutExplicit {
-		disconnectedTimeout = defaultDisconnectedTimeout
-	}
-
-	return disconnectedTimeout + a.failedTimeout
+	return a.disconnectedTimeout + a.failedTimeout
 }
 
 func (a *Agent) updateConnectionState(newState ConnectionState) {
@@ -919,6 +909,10 @@ func (a *Agent) findPair(local, remote Candidate) *CandidatePair {
 // validateSelectedPair checks if the selected pair is (still) valid
 // Note: the caller should hold the agent lock.
 func (a *Agent) validateSelectedPair() bool {
+	return a.validateSelectedPairWithTimeout(a.disconnectedTimeout)
+}
+
+func (a *Agent) validateSelectedPairWithTimeout(disconnectedTimeout time.Duration) bool {
 	selectedPair := a.getSelectedPair()
 	if selectedPair == nil {
 		return false
@@ -929,19 +923,22 @@ func (a *Agent) validateSelectedPair() bool {
 	// Only allow transitions to failed if a.failedTimeout is non-zero
 	totalTimeToFailure := a.failedTimeout
 	if totalTimeToFailure != 0 {
-		totalTimeToFailure += a.disconnectedTimeout
+		totalTimeToFailure += disconnectedTimeout
 	}
 
-	a.updateConnectionState(a.connectionStateForDisconnection(disconnectedTime, totalTimeToFailure))
+	a.updateConnectionState(a.connectionStateForDisconnection(
+		disconnectedTime, disconnectedTimeout, totalTimeToFailure,
+	))
 
 	return true
 }
 
 func (a *Agent) connectionStateForDisconnection(
 	disconnectedTime time.Duration,
+	disconnectedTimeout time.Duration,
 	totalTimeToFailure time.Duration,
 ) ConnectionState {
-	disconnected := a.disconnectedTimeout != 0 && disconnectedTime > a.disconnectedTimeout
+	disconnected := disconnectedTimeout != 0 && disconnectedTime > disconnectedTimeout
 	failed := totalTimeToFailure != 0 && disconnectedTime > totalTimeToFailure
 
 	switch {
@@ -2049,7 +2046,10 @@ func (a *Agent) setSelector() {
 		s = &controlledSelector{agent: a, log: a.log}
 	}
 	if a.lite {
-		s = &liteSelector{pairCandidateSelector: s}
+		s = &liteSelector{
+			pairCandidateSelector: s,
+			disconnectedTimeout:   a.liteDisconnectedTimeout,
+		}
 	}
 
 	s.Start()
diff --git a/agent_config.go b/agent_config.go
index 97bb8210..41548baf 100644
--- a/agent_config.go
+++ b/agent_config.go
@@ -299,10 +299,11 @@ func (config *AgentConfig) initWithDefaults(agent *Agent) { //nolint:cyclop
 
 	if config.DisconnectedTimeout == nil {
 		agent.disconnectedTimeout = defaultDisconnectedTimeout
+		agent.liteDisconnectedTimeout = defaultLiteDisconnectedTimeout
 	} else {
 		agent.disconnectedTimeout = *config.DisconnectedTimeout
+		agent.liteDisconnectedTimeout = *config.DisconnectedTimeout
 	}
-	agent.disconnectedTimeoutExplicit = config.DisconnectedTimeout != nil
 
 	if config.FailedTimeout == nil {
 		agent.failedTimeout = defaultFailedTimeout
diff --git a/agent_options.go b/agent_options.go
index 90cb2c7c..3111546d 100644
--- a/agent_options.go
+++ b/agent_options.go
@@ -386,7 +386,7 @@ func WithDisconnectedTimeout(timeout time.Duration) AgentOption {
 		}
 
 		a.disconnectedTimeout = timeout
-		a.disconnectedTimeoutExplicit = true
+		a.liteDisconnectedTimeout = timeout
 
 		return nil
 	}
diff --git a/selection.go b/selection.go
index 845dec47..5bc3791a 100644
--- a/selection.go
+++ b/selection.go
@@ -55,9 +55,13 @@ func (s *controllingSelector) isNominatable(c Candidate) bool {
 }
 
 func (s *controllingSelector) ContactCandidates() {
+	s.contactCandidates(s.agent.disconnectedTimeout)
+}
+
+func (s *controllingSelector) contactCandidates(disconnectedTimeout time.Duration) {
 	switch {
 	case s.agent.getSelectedPair() != nil:
-		if s.agent.validateSelectedPair() {
+		if s.agent.validateSelectedPairWithTimeout(disconnectedTimeout) {
 			s.log.Trace("Checking keepalive")
 			s.agent.checkKeepalive()
 
@@ -519,17 +523,18 @@ func (s *controlledSelector) HandleBindingRequest(message *stun.Message, local,
 
 type liteSelector struct {
 	pairCandidateSelector
+	disconnectedTimeout time.Duration
 }
 
 // A lite selector should not contact candidates.
 func (s *liteSelector) ContactCandidates() {
-	if _, ok := s.pairCandidateSelector.(*controllingSelector); ok {
+	if v, ok := s.pairCandidateSelector.(*controllingSelector); ok {
 		//nolint:godox
 		// https://github.com/pion/ice/issues/96
 		// TODO: implement lite controlling agent. For now falling back to full agent.
 		// This only happens if both peers are lite. See RFC 8445 S6.1.1 and S6.2
-		s.pairCandidateSelector.ContactCandidates()
+		v.contactCandidates(s.disconnectedTimeout)
 	} else if v, ok := s.pairCandidateSelector.(*controlledSelector); ok {
-		v.agent.validateSelectedPair()
+		v.agent.validateSelectedPairWithTimeout(s.disconnectedTimeout)
 	}
 }
```

Sean-Der (approved):

JoTurk:
@Sean-Der i changed the timeouts to safer value of 10 disconnect / 35 failed, instead of 30/55s which was extreme :)

PR #958, opened 2026-08-24
Prevent ICE-lite agents from sending connectivity checks

### LINKED ISSUE #510: Why does lite agent dial when checking connectivity?
I am learning the ice logic about ice-lite and found an interesting case `TestConnectivityLite `in `agent_test.go`

There are 2 agents, aAgent works as full ice agent and bAgent works as lite ice agent, in `TestConnectivityLite` case. They connect each other in function `connectWithVNet`

```golang
	cfg0 := &AgentConfig{
		Urls:             []*URL{stunServerURL},
		NetworkTypes:     supportedNetworkTypes(),
		MulticastDNSMode: MulticastDNSModeDisabled,
		//Net:              v.net0,
	}

	aAgent, err := NewAgent(cfg0)
	require.NoError(t, err)
	require.NoError(t, aAgent.OnConnectionStateChange(aNotifier))

	cfg1 := &AgentConfig{
		Urls:             []*URL{},
		Lite:             true,
		CandidateTypes:   []CandidateType{CandidateTypeHost},
		NetworkTypes:     supportedNetworkTypes(),
		MulticastDNSMode: MulticastDNSModeDisabled,
		//Net:              v.net1,
	}

	bAgent, err := NewAgent(cfg1)
	require.NoError(t, err)
	require.NoError(t, bAgent.OnConnectionStateChange(bNotifier))

	aConn, bConn := connectWithVNet(aAgent, bAgent)
```

In `connectWithVNet`, the 2 agent start to check the connectivity after they exchange their candidates, like below code. But why, **bAgent** start to `Dial` instead of `aAgent` calling `Accept`? The bAgent is a lite agent, maybe it should be listener according RFC 5245.

```
# RFC 5245 https://www.rfc-editor.org/rfc/rfc5245#page-30

   One agent full, one lite:  The full agent MUST take the controlling
      role, and the lite agent MUST take the controlled role.  The full
      agent will form check lists, run the ICE state machines, and
      generate connectivity checks.  That agent will execute the logic
      in Section 8.1 to nominate pairs that will be selected by ICE, and
      use the logic in Section 8.1.2 to end ICE.  The lite
      implementation will just listen for connectivity checks, receive
      them and respond to them, and then conclude ICE as described in
      Section 8.2.  For the lite implementation, the state of ICE
      processing for each media stream is considered to be Running, and
      the state of ICE overall is Running.
```

```golang

	accepted := make(chan struct{})
	var aConn *Conn

	go func() {
		var acceptErr error
		aConn, acceptErr = aAgent.Accept(context.TODO(), bUfrag, bPwd)
		check(acceptErr)
		close(accepted)
	}()

	bConn, err := bAgent.Dial(context.TODO(), aUfrag, aPwd)
	check(err)
	
```

Sean-Der:
liteSelector embedded pairCandidateSelector, which promoted the full
    agent methods onto the lite selector. An inbound Binding request took
    this path:
    
    Agent.handleInbound
      -> controlledSelector.HandleBindingRequest
      -> PingCandidate
      -> sendBindingRequest
    
    The lite agent therefore responded to an incoming check and originated
    another one.
    
    Store the wrapped selector in a named field and implement the selector
    interface explicitly. Controlled lite agents can process Binding
    requests without exposing active ping and success-response behavior.
    Keep the existing delegation for the controlling fallback.
    
    Add an end-to-end test that verifies the full agent sends checks while
    the lite agent only receives them.
    
    Fixes #510.

### DIFF
--- a/agent_test.go
+++ b/agent_test.go
@@ -1429,6 +1429,75 @@ func TestConnectivityLite(t *testing.T) {
 	<-bConnected
 }
 
+// TestLiteAgentDoesNotSendConnectivityChecks verifies that an ICE-lite agent
+// responds to checks from a full agent without originating checks itself.
+func TestLiteAgentDoesNotSendConnectivityChecks(t *testing.T) {
+	defer test.CheckRoutines(t)()
+
+	fullAgentNATType := &vnet.NATType{
+		MappingBehavior:   vnet.EndpointIndependent,
+		FilteringBehavior: vnet.EndpointIndependent,
+	}
+	liteAgentNATType := &vnet.NATType{Mode: vnet.NATModeNAT1To1}
+	virtualNet, err := buildVNet(fullAgentNATType, liteAgentNATType)
+	require.NoError(t, err)
+	defer virtualNet.close()
+
+	fullAgent, err := NewAgent(&AgentConfig{
+		Urls: []*stun.URI{{
+			Scheme: SchemeTypeSTUN,
+			Host:   vnetSTUNServerIP,
+			Port:   vnetSTUNServerPort,
+			Proto:  stun.ProtoTypeUDP,
+		}},
+		NetworkTypes:     supportedNetworkTypes(),
+		MulticastDNSMode: MulticastDNSModeDisabled,
+		Net:              virtualNet.net0,
+	})
+	require.NoError(t, err)
+	defer func() { require.NoError(t, fullAgent.Close()) }()
+
+	liteAgent, err := NewAgent(&AgentConfig{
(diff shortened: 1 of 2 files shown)


JoTurk:
@Sean-Der the "controlledSelector.HandleBindingRequest => PingCandidate" issue was fixed in #909 we didn't send connectivity checks since this (except if both peers were controlling). I think this PR adds good defensive design, but I think the PR description is wrong the issue was already fixed :)
Sorry that I didn't try to find all the old issues when #909 was merged.

PR #961, opened 2026-08-26
Do not validate controlled ICE-lite pairs from timer ticks

zshang-oai:
## Problem

#959 lets a controlled ICE-lite agent's timer mark its highest-priority available pair `Succeeded`, without a received check or nomination. The peer may be a full agent: the current API supplies remote credentials and the local role, not the peer's lite/full mode or a finalized candidate exchange.

This creates false write readiness. In the new vanilla-ICE regression, a reachable pair receives an authenticated request, while a blocked higher-priority pair receives none. The timer nevertheless promotes the blocked pair, and ordinary `Conn.Write` sends through it instead of reporting `ErrNoCandidatePairs` before nomination.

## Change

Remove only controlled-side timer promotion. Keep selected-pair liveness maintenance, accepted nomination/custom-handler selection, and the prohibition on lite-originated checks. The controlling-lite local-selection code and the `WriteToPair` validity gate are unchanged. No SPED, dependency, or public API changes.

Add unit tests for untouched, delayed, and already-valid/selected pairs, plus a real full-to-lite VNet regression. It blocks the higher-priority path and temporarily withholds nomination, verifies the error before nomination, then requires both `AwaitConnect` calls, both Connected notifications/states, correct selected addresses, and normal bidirectional `Conn.Write` after genuine nomination.

## Intentional behavior change and remaining both-lite work

**This is a narrow safety correction, not complete both-lite support.** It removes the controlled-side `WriteToPair` shortcut introduced by #959: a still-unvalidated pair now returns `ErrCandidatePairNotSucceeded`. Ordinary controlled-side both-lite startup was already incomplete; simply selecting a pair on that side would incorrectly connect full-to-lite sessions too.

The former `TestLiteMode_LiteControlling_Integration` is explicitly renamed as controlling-side-selection-only coverage. Its controlling connection, role, zero-check, and one-way-delivery assertions remain. The controlled response shortcut is replaced with assertions that the pair stays unvalidated and both write APIs reject it. A short bounded `AwaitConnect` check records the missing completion path. The separate full-to-lite test still verifies real bidirectional startup and traffic.

Known both-lite peers may establish valid pairs without checks; multiple-pair completion requires reconciled candidate exchange. [RFC 8445 §8.2](https://www.rfc-editor.org/rfc/rfc8445.html#section-8.2). Completing that support needs a separately reviewed per-session remote-mode and exchange/selection contract, including delayed candidates and restart. This PR does not resolve the unchanged controlling-side multi-pair selection assumptions or introduce an unagreed signaling API.

## Validation

Base: `dc79ba9`. Candidate: `ace9d67`. Go 1.24.0, macOS ARM64.

- Same tests, module graph, and environment: the new routing regression fails **3/3** on unmodified production and passes **20/20** with this correction. The companion controlled-selector tests pass all 20 repetitions with `-race`.
- The original #959 `WriteToPair` shortcut passes **3/3** before the correction and fails **3/3** afterward with the expected pair-state error, before adapting its assertions. This behavior change is intentional and retained in the evidence.
- Focused lite/selector/role/transport tests pass three times both with and without the race detector.
- The **full race suite without the UDP-fixture overlay did not complete**: an existing UDP-mux fixture received 13 bytes where it expected 8192, then timed out. The same failure reproduced on vanilla `dc79ba9`; the fixture's fixed sleep does not ensure its initial probe is dropped before address registration. It constructs no ICE Agent or selector.
- A separate full race run with only three UDP **test files** overlaid to wait for the actual drop decision passed: **947 test/subtest passes, three existing skips**. That run required local UDP access outside the sandbox after sandboxed sends returned `operation not permitted`. The overlay changes no production code and is **not included in this PR**. The original failed runs remain recorded separately.
- `go vet ./...`, formatter checks on the four changed files, and diff-scoped `golangci-lint` pass. Full-repository lint is not claimed clean: vanilla upstream reports 187 issues with the installed golangci-lint 2.13.1.

```sh
go test -race -count=1 ./...
go test -race -count=20 \
  -run '^(TestLiteControlledSelectorContactCandidates.*|TestLiteControlledDoesNotUseUnverifiedPairBeforeNomination)$' .
```

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -113,6 +113,7 @@ type Agent struct {
 
 	remoteUfrag      string
 	remotePwd        string
+	remoteLite       bool
 	remoteCandidates map[NetworkType][]Candidate
 
 	checklist  []*CandidatePair
@@ -1921,6 +1922,15 @@ func (a *Agent) SetRemoteCredentials(remoteUfrag, remotePwd string) error {
 	})
 }
 
+// SetRemoteICELite sets whether signaling indicated that the remote agent uses ICE-lite.
+// Call this before starting connectivity checks. Remote agents are assumed to use full ICE by default.
+func (a *Agent) SetRemoteICELite(lite bool) error {
+	return a.loop.Run(a.loop, func(_ context.Context) {
+		a.remoteLite = lite
+		a.requestConnectivityCheck()
+	})
+}
+
 // UpdateOptions applies the given options to the agent at runtime.
 // Only a subset of options can be updated after agent creation:
 //   - WithUrls: updates STUN/TURN server URLs (takes effect on next GatherCandidates call)
@@ -2054,7 +2064,7 @@ func (a *Agent) setSelector() {
 		s = &controlledSelector{agent: a, log: a.log}
 	}
 	if a.lite {
-		s = &liteSelector{pairCandidateSelector: s}
+		s = &liteSelector{pairCandidateSelector: s, agent: a}
 	}
 
 	s.Start()
--- a/selection.go
+++ b/selection.go
@@ -518,46 +518,24 @@ func (s *controlledSelector) HandleBindingRequest(message *stun.Message, local, 
 }
 
 type liteSelector struct {
-	pairCandidateSelector pairCandidateSelector
-}
-
-func (s *liteSelector) Start() {
-	s.pairCandidateSelector.Start()
+	pairCandidateSelector
+	agent *Agent
 }
 
 func (s *liteSelector) ContactCandidates() {
-	if controlled, ok := s.pairCandidateSelector.(*controlledSelector); ok {
-		if !controlled.agent.validateSelectedPair() {
-			if pair := controlled.agent.getBestAvailableCandidatePair(); pair != nil {
-				pair.state = CandidatePairStateSucceeded
-			}
-		}
-
+	if s.agent.validateSelectedPair() || !s.agent.remoteLite {
 		return
 	}
 
-	controlling, ok := s.pairCandidateSelector.(*controllingSelector)
-	if !ok || controlling.agent.getSelectedPair() != nil {
-		if ok {
-			controlling.agent.validateSelectedPair()
-		}
-
-		return
-	}
-
-	pair := controlling.agent.getBestAvailableCandidatePair()
+	pair := s.agent.getBestAvailableCandidatePair()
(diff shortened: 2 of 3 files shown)


Sean-Der:
I need to look at this later, but my guess is that `ContactCandidates` should be a no-op.

A `ice-lite` implementation should only be responding to traffic. If the remote is `ice-lite` (and we are controlling) only then `ContactCandidates` should be doing something.

I will look into this tonight. But reading the description in the PR I don't follow/understand the intent yet.

lite_selection_regression_test.go
```diff
@@ -0,0 +1,247 @@
+// SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
```
JoTurk: Please merge these new test files (lite_selection_regression_test.go and lite_selector_safety_test.go) with any existing test file you see fit, we try to limit how much files we have on /

JoTurk (commented):
@Sean-Der the issue is that now we mix "lite and controlled" and "both are lite", and then we mark candidates as `Succeeded` without even receiving a successful STUN (this affects LITE x full-agents) 

IMO since we didn't tag this. I think we should revert #959 and redo it under the condition that the other peer is also LITE rather then merging this fix which I think might introduce other regressions. LITE and LITE should be rare enough and it should be safe to gate such fix. we shouldn't also touch the controlled-lite path. SO it should be only gated behind (local-lite && remote-lite && controlling).

Thank you @zshang-oai

Sean-Der:
```
    Add explicit configuration for remote ice-lite
    
    A controlled ICE-lite selector marked its best available pair as
    succeeded on each ContactCandidates tick without knowing whether its
    peer was full or lite. This made an unvalidated route writable before a
    full controlling peer nominated it.
    
    Add SetRemoteICELite so signaling can identify lite peers while
    defaulting to full ICE. Lite peers select their pair locally without
    connectivity checks; otherwise, controlled agents wait for an
    authenticated USE-CANDIDATE request.
    
    Cover both-lite operation end to end by requiring both agents to
    connect, exchange data in both directions, and send no connectivity
    checks.
```

Added an API for users to set if remote is `ice-lite` will get webrtc using it also! Hopefully no one is running lite against lite though :)

Sean-Der (approved):

PR #972, opened 2026-09-08
Fix UniversalUDPMux initialization race

doraemonkeys:
STUN packets arriving before or during UniversalUDPMux creation can be processed before initialization finishes, causing a data race or a possible nil-pointer panic.

This change initializes the embedded UDP mux before starting the reader. Regression tests cover both ReadFrom and ReadFromAddrPort.

Validation:
- Both tests reproduce the race before the fix.
- Both pass 100 runs with the race detector after the fix.
- The full Go test suite passes with the race detector enabled.

### DIFF
--- a/udp_mux.go
+++ b/udp_mux.go
@@ -90,6 +90,13 @@ type UDPMuxParams struct {
 
 // NewUDPMuxDefault creates an implementation of UDPMux.
 func NewUDPMuxDefault(params UDPMuxParams) *UDPMuxDefault {
+	mux := newUDPMuxDefault(params)
+	go mux.connWorker()
+
+	return mux
+}
+
+func newUDPMuxDefault(params UDPMuxParams) *UDPMuxDefault {
 	if params.Logger == nil {
 		params.Logger = logging.NewDefaultLoggerFactory().NewLogger("ice")
 	}
@@ -127,7 +134,6 @@ func NewUDPMuxDefault(params UDPMuxParams) *UDPMuxDefault {
 		isUnspecified: isUnspecified,
 	}
 	mux.addrPortConn = asAddrPortReaderWriter(params.UDPConn)
-	go mux.connWorker()
 
 	return mux
 }
--- a/udp_mux_universal.go
+++ b/udp_mux_universal.go
@@ -89,7 +89,8 @@ func NewUniversalUDPMuxDefault(params UniversalUDPMuxParams) *UniversalUDPMuxDef
 		UDPConn: mux.params.UDPConn,
 		Net:     mux.params.Net,
 	}
-	mux.UDPMuxDefault = NewUDPMuxDefault(udpMuxParams)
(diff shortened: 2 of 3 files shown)


JoTurk:
This does not seem like a realistic race in the normal UniversalUDPMux lifecycle. We cannot send a stun request through the mux until NewUniversalUDPMuxDefault has returned, so a corresponding stun response should not be able to arrive during construction.

The race technically exists in the code if the supplied "PacketConn" already has a STUN packet queued, but I'm not sure that's a scenario we realistically need to handle.

Was this found by an AI agent (per our new policy we require such finding to be backed by a prove that It can happen)? Or is there an actual code where this triggers a race?

Sean-Der:
@JoTurk I was thinking it *might* be a problem if users are creating them on a known IP/Port?

If I am a starting up servers that listen on a singular Port and the attacker knows to spray X. Some companies run `-race` on a small number of servers and this could hit it.

Super unlikely, but it's a inconsequential change so I think ok?

Sean-Der (approved):

Sean-Der:
@JoTurk Feel free to close or merge! W/e you think is best.

JoTurk:
@Sean-Der I think it looks okay and the race does exist, and I think we should merge it, I just feel that it's similar to many similar findings we get and it just doesn't affect any user, I'm just going to point the new policy in similar prs even when we merge them hopefully I'm not super annoying, please tell me if I should stop :)

Sean-Der:
@JoTurk Keep doing it! The description’s are especially hard because they confuse me/understand less.

I have always been regretfully passive. pion/sdp at one point got a mega rewrite I regret not pushing back on. Or not saying more with ion.

anyway I stand by anything you are passionate about!

doraemonkeys:
@JoTurk AI helped diagnose and fix this after [WindShare’s CI reported the race](https://github.com/WindShare/windshare/actions/runs/33959704276/job/101289317313) in an existing test using real UDP sockets. The queued-STUN regression test was added afterward. Sorry, I should’ve included that context and the original failure log in the PR description.

JoTurk:
@doraemonkeys thank you for the context this is more clear now, so it's a Go race and not a logical race, or stun packet race.
The ci you linked runs with -race and constructs the API in a artificial way his is why it triggered it still not likely to trigger in the real prod
https://github.com/WindShare/windshare/blob/75425f2/connectivity%2Fsocketauthority%2Fauthority.go#L88
https://github.com/WindShare/windshare/blob/53912ae16a2fd8a79959e9813412a0aacdbbcf57/internal/testloopback/pion.go#L51

I think this is a good fix and we should merge it but this is a good example why we need to update the policy and also add examples like @Sean-Der  suggested in discord, it would have been a lot easier to understand if this was described as "breaks this network test with a link to the test itself" rather than using an agent to try to describe the problem, also the queued stun regression test it added isn't related to what broke the CI you linked.

JoTurk (approved):

PR #975, opened 2026-09-09
Range check candidate component-id and priority

youdie006:
`UnmarshalCandidate` rejects a port over 65535, but a component-id over 65535
and a priority over 2^32-1 are truncated instead of rejected. Both of these
parse with `err == nil` on current main:

```
4207374051 99999 udp 500 127.0.0.1 80 typ host
  -> Component() == 34463
  -> Marshal()   == "4207374051 34463 udp 500 127.0.0.1 80 typ host"

4207374051 1 udp 5000000000 127.0.0.1 80 typ host
  -> Priority() == 705032704
  -> Marshal()  == "4207374051 1 udp 705032704 127.0.0.1 80 typ host"
```

So the caller ends up holding a candidate with a number the remote peer never
sent. That matters because this is the function pion/webrtc runs over every
`a=candidate:` line of a remote SDP (`sdp.go:958`), and a truncated priority
then feeds candidate-pair ordering.

`readCandidatePort` already range checks its value (`candidate_base.go:1069`).
Component-id and priority go through the same `readCandidateDigitToken`, which
only limits the digit *count* - 5 digits still admits 99999 and 10 digits still
admits 9999999999, and neither fits the `uint16`/`uint32` that `Component()`
and `Priority()` return (`candidate.go:37`, `candidate.go:50`).

So I added `readCandidateComponent` and `readCandidatePriority` next to
`readCandidatePort` and range check there, returning the existing
`errParseComponent` / `errParsePriority`. `readCandidateDigitToken` now
accumulates into a `uint64` so a 10-digit priority cannot overflow before the
check runs on the i386 job.

That also drops the four `//nolint:gosec // G115 no overflow we read 5 digits`
comments. The conversions they covered are gone, and the claim was not correct
anyway - it also says "5 digits" on the 10-digit priority field.

I bounded on uint16/uint32 rather than RFC 8445 5.1.2's tighter priority range
of 1..2^31-1, so this only rejects values the API cannot represent at all and
doesn't start refusing candidates that implementations actually send today.

One bit of test churn worth flagging: several existing invalid-input rows in
`TestCandidateMarshal` use `99999` as filler in the component slot. After this
change those rows would be rejected on the component rather than on the thing
they are there to test, so I changed the filler to `1`.

Checks I ran locally: `golangci-lint run ./...` at v2.10.1 (the version
`lint.yaml` pins) reports 0 issues, `gofmt`/`go vet ./...`/`GOARCH=386 go vet
./...` are clean, and the new tests fail on main and pass with the fix. I could
not get the full `./...` suite to finish on my machine - `TestMultiUDPMux`
hits the `test.TimeOut` watchdog at `udp_mux_multi_test.go:73` - but it aborts
at the identical test and line on an unmodified checkout too, so it is a local
networking limitation rather than something this change causes. The rest of the
package (the tests that don't need that networking setup) is green before and
after, as are `internal/stun` and `internal/taskloop`.

Disclosure: I used AI assistance while working on this. I ran every check quoted
above myself, and I'll handle review discussion personally.

### DIFF
--- a/candidate_base.go
+++ b/candidate_base.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"hash/crc32"
 	"io"
+	"math"
 	"net"
 	"net/netip"
 	"strconv"
@@ -840,7 +841,7 @@ func UnmarshalCandidate(raw string) (Candidate, error) { //nolint:cyclop
 	}
 
 	// component-id ( 1*5DIGIT )
-	component, pos, err := readCandidateDigitToken(raw, pos, 5)
+	component, pos, err := readCandidateComponent(raw, pos)
 	if err != nil {
 		return nil, fmt.Errorf("%w: %v in %s", errParseComponent, err, raw) //nolint:errorlint // we wrap the error
 	}
@@ -857,7 +858,7 @@ func UnmarshalCandidate(raw string) (Candidate, error) { //nolint:cyclop
 	}
 
 	// priority ( 1*10DIGIT ) SP
-	priority, pos, err := readCandidateDigitToken(raw, pos, 10)
+	priority, pos, err := readCandidatePriority(raw, pos)
 	if err != nil {
 		return nil, fmt.Errorf("%w: %v in %s", errParsePriority, err, raw) //nolint:errorlint // we wrap the error
 	}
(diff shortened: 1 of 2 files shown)


Sean-Der:
Great change, thank you @youdie006 !

Sean-Der (approved):

PR #977, opened 2026-09-12
gather: provide descriptive context on IPv6 STUN gathering warnings

### LINKED ISSUE #23: Make IPv6 error message more descriptive
### What did you do?

Started an ipv6 ICE candidate in a network that doesn't support ipv6

### What did you expect?

A friendly error message

### What happened?

This scary error message was logged:

`ice WARNING: 2019/04/14 01:35:50 could not allocate udp6 stun:stun.stunprotocol.org:3478: failed to create STUN client`

We should let people know why this message happened and whether it can be ignored.

(reported by @adwpc)

wangyusheng1985:
### Context

When running ICE candidate gathering in networks without IPv6 connectivity, attempts to allocate or query STUN servers over IPv6 result in warning log messages that can appear alarming to users.

### Changes

- In `gather.go`, for `udp6` network types during server-reflexive candidate gathering (both standard and UDPMuxSrflx paths), add descriptive context informing users that IPv6 might not be supported on their network and that the warning can usually be ignored if IPv6 connectivity is not expected.

Validation observed for this change:
- `go test -count=1 -run TestGatherCandidatesSrflx ./...`
- `go test -count=1 ./...`
- `go vet ./...`

Fixes #23

### DIFF
--- a/gather.go
+++ b/gather.go
@@ -12,6 +12,7 @@ import (
 	"reflect"
 	"slices"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Sean-Der:
Closing this and #23 as a wontfix

I think the complexity/abstraction will end up frustrating someone in the future. We should just give users the output from Go stdlib so they can debug easily. 

Sorry @wangyusheng1985 and thank you for the contribution.

PR #978, opened 2026-09-13
Allow empty address rewrite mappings

Sean-Der:
Match the docs on AddressRewriteRule


JoTurk (commented):
I added this check as a quick fix before merging the PR originally because it was breaking the old behavior for the nat1:1 API and was creating edge cases, sorry for not updating the documentation

Sean-Der:
@JoTurk ah cool makes sense! I was reading the address rewrite stuff (great job on it!) and just noticed. I will update the doc instead.

PR #979, opened 2026-09-13
Remove redundant address rewrite tests

Sean-Der:
Replace repeated mapper, option, gather, and VNet tests with public agent and end-to-end coverage.

Statement coverage remains above the original suite. Address rewrite rules continue to require an external address, and the public documentation now says so explicitly.

Also close relay resources when filtering leaves no usable mapped address.


JoTurk (commented):
@Sean-Der sorry about adding a billion test, this wasn't my plan for this API internally and when you talked about changing the legacy API, it was just before 4.2 with the RACK change and all of that, and i was so scared to break anything https://github.com/pion/ice/pull/834#issuecomment-3576776053

so i added more tests than prod loc and it all worked out, because nobody reported any bugs

JoTurk:
@Sean-Der I consolidated the tests instead of just removing them, reviewed one by one and made sure we're not dropping any behavior test.

PR #986, opened 2026-09-16
Stop leaking a goroutine in a gathering test

maum-meliora:
`TestContinualGatheringPolicy` collects candidates through a buffered channel it reads only until it has what it needs. An agent that gathers more candidates than the buffer holds leaves the handler blocked on the send; closing the agent cannot release a send into a channel the test owns, so the goroutine outlives the test and fails the startup routine check of every later test that has one.

This only shows on hosts with more than ten local IPv4 addresses (one host candidate each) - a machine with several Docker bridges reproduces it on every full run, while a typical CI runner never does. Lowering the channel buffer to 1 reproduces it anywhere.

Send without blocking. The tests only assert on the gathering state (and, for continual gathering, that at least one candidate arrives), so candidates beyond the buffer can be dropped.

## Testing

- Full `go test -race ./...` on this branch: 0 "Unexpected routines on test startup" failures (68 tests failed that way on `main` from this host).
- With the channel buffer lowered to 1: `main` fails the next startup routine check, this branch passes.
- The one remaining `-race` failure, `TestUDPConnReadWriteDoesNotAllocate`, fails identically on `main` and is unrelated.


JoTurk (approved):
Thank you

JoTurk:
Please check https://github.com/pion/webrtc/wiki/Contributing#ai-assisted-pull-requests

PR #988, opened 2026-09-16
Stop counting ICE traffic in the allocation test

maum-meliora:
`TestUDPConnReadWriteDoesNotAllocate` starts measuring as soon as its warm-up
round trips finish, but both agents are still running the connectivity checks
from setup. `testing.AllocsPerRun` counts allocations process-wide, so the STUN
messages those checks build and parse are charged to the data path, and the
assertion fails with tens of allocations per round trip even though the path
itself allocates none.

Under the race detector the checks outlast the warm-up every time, so the test
fails on every run of the race lanes. Without it the warm-up usually wins and
the test passes, failing only when the machine is busy.

Wait for the process to stop allocating before measuring. The round trip then
measures zero with the race detector enabled, so the assertion holds there
instead of only reporting noise.

## Testing

- `go test -race -count=10 -run TestUDPConnReadWriteDoesNotAllocate`: 10/10 fail
  on `main`, 0/10 fail on this branch.
- Without `-race`, the same command fails 1/10 on `main` and 0/10 here.
- Full `go test -race ./...` on this branch: 0 failures. This was the one
  remaining failure called out in #986.


Sean-Der:
Instead of time is it possible to exclude certain objects? It would be nice to be explicit

JoTurk (changes requested):
can you just disable keep-alive for this test?

Edit: you can just set KeepaliveInterval in AgentConfig to 0.

JoTurk:
Sorry i commented before i saw Sean's comment :)

maum-meliora:
Dropped the helper. The agents now get one loopback candidate each and no
keepalives, so a single pair is checked once and nothing runs alongside the
measurement. No waiting.

`-count=10` with and without `-race`: 0 failures.

JoTurk (approved):
thank you. It's much better without keep-alive compared to the filter helper

PR #989, opened 2026-09-17
Close the agents a failing test leaves behind

maum-meliora:
## Problem

Tests that build a pair of agents close them on the last line of the body, so an assertion that fails earlier skips the teardown. The agents and every candidate receive loop outlive the test, and the next test to call `test.CheckRoutines` aborts at startup — as does every one after it.

Forcing a single assertion in `TestBindingRequestHandler` to fail:

```
before   14 failing tests
after     1 failing test
```

`pipeWithVNet` has a second gap: it builds both agents and then calls `connectWithVNet`, so a failure while connecting returns before the caller has anything to defer on.

## Fix

Close the conns from a `defer` taken where they are built, and hand the agents to `t.Cleanup` as they are made where the caller cannot yet defer. This is the pattern `pipe()` and `pipeWithVNetUsingOptions()` already use.

Test-only; no library code is touched.

## Verification

- The numbers above, reproduced by forcing the same assertion to fail with and without the change.
- Full suite under `-race`: only `TestUDPConnReadWriteDoesNotAllocate`, which fails identically on `main` on this machine.
- golangci-lint v2.10.1 clean; `go test` green on Go 1.24 and 1.25 across linux, i386, macOS, windows and wasm.


JoTurk (approved):
thank you for all the test fixes

PR #992, opened 2026-09-17
Defer UDP mux write cancellation to 5s

### LINKED ISSUE #991: UDPMux: closing one ICE agent can interrupt another agent’s write
## Summary

The write-cancellation mechanism introduced in [#934](https://github.com/pion/ice/pull/934) can interrupt an unrelated connection sharing the same UDP mux.

In particular, **closing agent A can expire the shared socket’s write deadline while only agent B is writing**. B then receives a write timeout even though B remains open and its own deadline has not expired.

Reproduced against `da1692c11684351604d13afec6616361316f3007`.

## Reproduction sequence

Two agents use separate logical connections from one `UDPMuxDefault`:

1. Start a write from B.
2. Pause that write after it enters the mux’s write accounting, but before the underlying UDP write completes.
3. Close A through its normal agent/candidate shutdown path. A has no active write.
4. A’s shutdown calls `abortWrite()`, which sets the shared physical socket’s write deadline to the current time.
5. Resume B’s write. It fails because of the deadline set by A.

A barrier around the underlying `PacketConn` makes this interleaving deterministic; it does not require reproducing kernel-buffer exhaustion. Closing the logical mux connection directly is not equivalent to the agent shutdown path, which explicitly invokes `abortWrite()`.

The ownership problem reproduces through both `WriteTo` and `WriteToAddrPort`. There is also interference when A and B both have outstanding writes and A closes.

## Why it happens

[Candidate shutdown](https://github.com/pion/ice/blob/da1692c11684351604d13afec6616361316f3007/candidate_base.go#L459) forwards cancellation to the mux.

However, [`UDPMuxDefault.abortWrite`](https://github.com/pion/ice/blob/da1692c11684351604d13afec6616361316f3007/udp_mux.go#L375) checks the **socket-wide active-write count**, without identifying which logical connection owns those writes.

The counter and deadline-reset coordination prevent subsequent writes from entering while cancellation is being processed, but do not isolate cancellation between current writers.

## Impact

An unrelated connection loses an outgoing packet as a consequence of another agent closing.

At the higher ICE connection layer, [`candidateBase.writeTo`](https://github.com/pion/ice/blob/da1692c11684351604d13afec6616361316f3007/candidate_base.go#L475) can suppress the resulting timeout, exposing `(0, nil)` for a nonempty write.

The downstream consequence depends on the consumer:

- Focused pure-Go DTLS tests recovered from the interrupted write through retransmission.
- A native DTLS integration treated the result as a short write; controlled tests demonstrated both handshake failure and established DTLS/SCTP transport closure.

These observations do **not** establish the production frequency or imply that default Pion WebRTC necessarily experiences terminal connection failure.

## Expected behavior

Closing A should cancel A’s pending operations without expiring B’s writes solely because they share a UDP socket.

At the same time, **#934’s original shutdown fix must remain intact**: closing an agent must still unblock its own blocked write instead of hanging while waiting for the task loop.

Useful regression cases would cover:

- Idle A closes while B writes.
- A and B both have outstanding writes when A closes.
- A’s own blocked write is interrupted and shutdown completes.
- B remains usable afterward, without inheriting an expired internal deadline.

I’m reporting the cancellation-ownership issue rather than proposing a particular implementation. I would appreciate the maintainers’ preferred approach to preserving both shutdown progress and logical-connection isolation.

This is related to, but distinct from, #971: it concerns cancellation affecting another connection, rather than a live blocked writer delaying its peers.

@JoTurk, since this follows the cancellation support in #934, could you take a look?

JoTurk:
#### Description
https://github.com/pion/ice/issues/934 introduced close deadline fix for stuck close, stuck close can happen if the OS buffer is saturated this is likely related to a Go runtime change and reproducible with a simple stress-test on linux and macos.
#991 reported that this can drop writes, while the report is AI-generated and failed to demonstrate a real reproduction (edit: they sent an amplifier test), I tried to reproduce a dropped write with a real UDP socket but I failed to reproduce it because writeTo doesn't block in normal writes. However, It's just safer to just gate this shutdown-logic behind a 5s timeout so we don't deadlock forever if close blocks.

Moves many of the handing close tests to a syncTest since we unlock after 5s now.

#### Reference issue
closes #991


JoTurk:
made more sense to me to change the timeout to 5s from 10s

PR #993, opened 2026-09-17
Fix golangci-lint warnings

Sean-Der:
Use tagged switch and `reflect.Ptr` renamed


JoTurk:
I wonder if we can enable tagged switches now https://github.com/pion/.goassets/blob/f62fbb40818de7af67a9aa0d4d0370c5cfa038d5/ci/.golangci.yml#L87 it used to conflict with exhaustive

PR #994, opened 2026-09-18
Fix active TCP candidate registration and pairing

maum-meliora:
Active TCP candidates created for remote passive TCP candidates are
registered from within the agent event loop, so they cannot use
addCandidate without re-entering the loop. Their separate registration
path drifted from normal candidate registration, omitting the ufrag and
generation extensions and bypassing location-tracking filtering.

Later remote candidate registration also paired destination-bound active
TCP connections with unrelated remotes. This change stops creating those
pairs.

Extract addCandidateOnLoop so regular and active TCP candidates share
registration. Let destination-bound connections report which remote they
can write to, and apply that constraint when either side is registered.
This gives each pair's connectivity check its own 5-tuple connection
(RFC 6544 Section 7.1).

Tests cover candidate metadata, location filtering, and both
registration orders.


Sean-Der:
Thanks for the fix @maum-meliora!

Is it possible to make this all go through one path? Reduce drift in the future

agent.go
```diff
@@ -1121,25 +1121,29 @@ func (a *Agent) addRemotePassiveTCPCandidate(remoteCandidate Candidate) {
...
 		a.localCandidates[localCandidate.NetworkType()] = append(
 			a.localCandidates[localCandidate.NetworkType()],
 			localCandidate,
 		)
-		a.candidateNotifier.EnqueueCandidate(localCandidate)
+		if !localCandidate.filterForLocationTracking() {
+			a.candidateNotifier.EnqueueCandidate(localCandidate)
+		}
 
 		a.addPair(localCandidate, remoteCandidate)
```
JoTurk: I would suggest that you consolidate addCandidate and addRemotePassiveTCPCandidate instead of fixing the drifts, I noticed this when i was working on ufrag but i forgot to fix it

JoTurk: (NVM sean made the same comment :)

active_tcp_test.go
```diff
@@ -444,3 +445,270 @@ func TestActiveTCPConn_SetDeadlines_WhenConnected(t *testing.T) {
 	default:
 	}
 }
+
+func TestActiveTCP_CandidateCarriesStandardExtensions(t *testing.T) {
+	defer test.CheckRoutines(t)()
+	defer test.TimeOut(10 * time.Second).Stop()
+
```
JoTurk, line 452: I think we can remove all of these tests since it's a single path now.

maum-meliora: Removed the redundant extension test and narrowed the link-local test. I kept the pairing regression because it covers the new write-target constraint in both registration orders.

JoTurk:
@maum-meliora sorry about the conflicts

Sean-Der (approved):

PR #995, opened 2026-09-18
De-dupe address rewriting code

Sean-Der:
Same code existed for Host/Srflx/Relay candidates


Sean-Der:
Working on https://github.com/pion/ice/issues/780 and noticed

PR #998, opened 2026-09-18
Code cleanups

Sean-Der:
Trying to delete as much code as possible before v5 tag


agent.go
```diff
@@ -1936,30 +1936,13 @@ func (a *Agent) SetRemoteICELite(lite bool) error {
...
-			}
-
-			if optErr = opt(a); optErr != nil {
-				return
-			}
-		}
+// SetURLs updates the STUN/TURN server URLs for the next GatherCandidates call.
+// It copies the slice, but callers must not modify the URI objects while in use.
+// It returns ErrClosed if the agent is closed.
+func (a *Agent) SetURLs(urls []*stun.URI) error {
```
JoTurk, line 1935: we shouldn't remove UpdateOptions, it matches SRTP and it's a good safety net for future changes IMO

Sean-Der: Where do you draw the line on 'Options' vs having methods on a struct? 

If I am trying to use `ice.Agent` and I want to do something at runtime I am going to look through the list of methods. I don't think making users look through a list of options is a good API.

Also it isn't know at compile time what is updatable and what is not from an option?

For other Go libraries/APIs is mutation done via Option? 

---

Being consistent with SRTP is a good reason though!

JoTurk: Right now we only respect URLs change in SetConfiguration, if we want to support more (and actually support other configurations) we'll need to add more methods to Agent (can see a use case to update candidates-type on restart): https://github.com/pion/webrtc/blob/c869fd8f2018a7b6049453ab0f76e7c318811d75/configuration.go#L13-L49 


When UpdateOptions was added for urls https://github.com/pion/ice/pull/866 , i suggested adding a check for every option until we figure out what can be changed safely and what need to be behind a generation / snapshot until the next restart, in theory we can update all options, even ice-lite i can see a reason why someone would want to switch between lite and full agent (the Firefox ice bug is a good example i think).

anyways I don't have a hard feeling about this, I just think that we need a general pattern for options updates

JoTurk (approved):

PR #1001, opened 2026-09-21
Support hostname as ICE candidate

### LINKED ISSUE #512: Support hostname as ICE candidate
## Summary

From what I've come to understand this library does not support using a hostname as an ICE candidate. Is this something that can be added? 

## Motivation

Obvious gains of using hostnames over IPs :smile: 

The only alternative is using a list of static IPs. Works but isn't great.

## Additional context

https://github.com/AlexxIT/go2rtc/issues/146

Sean-Der:
Resolves #512


agent.go
```diff
@@ -763,27 +780,35 @@ func isMulticastDNSCandidate(cand Candidate) bool {
...
-	return nil
+func (a *Agent) resolveAndAddFQDNCandidate(cand Candidate, generation uint64) {
+	resolved, err := a.lookupNetIP(a.loop, "ip", cand.Address())
+	if err != nil || len(resolved) == 0 {
+		a.log.Warnf("Failed to resolve FQDN candidate %s: %v", cand.Address(), err)
+
+		return
+	}
+
+	a.addResolvedCandidate(cand, resolved[0], generation)
```
JoTurk, line 811: What if the host has multiple records? and what if it has A and AAAA records

JoTurk: I think it's a bigger problem if we resolve the IPV6 first, multiple addresses can wait, but if we have ipv4 and ipv6 and have to pick one for now i think we should go with ipv4

Sean-Der: For multiple records we could generate a bunch of PeerReflexive candidates.

I am not sure this is what server operators would want though? I am used to picking one IP out of the pool and publishing as ICE candidate for load balancing. If we hit all via FDQN it will generate a lot more traffic/split brain.

Changing the IPv4! Not sure what users expect either

JoTurk: I think random IP is normal, random address family would be unexpected if an address have both, and I think it can cause weird behavior / silent fail if ipv6 isn't supported, anyways this is just a nit, I think we should respect NetworkTypes at least

JoTurk (approved):

PR #1003, opened 2026-09-21
Optimize agent creation

JoTurk:
#### Description
~~This explores an idea of optimizing agent creation using a cached `stdnet ` that updates on network change + timer using https://github.com/pion/transport/pull/405~~

this feels hacky so I'm not sure super sure about it (also users could already provide cached setNet / watch for change and change it on their side), I added a benchmark for 100 agent creation.
on my machine:
Before: 32.34 ms, 13.02 MB, 46,153 allocations
After: 0.88 ms, 0.72 MB, 9,002 allocations.

37x faster and 94% less allocation


Sean-Der:
What does it look like to get rid of stdnet entirely? Does this exist just for `vnet` ?

JoTurk:
@Sean-Der the expensive part isn't stdnet, stdnet is nice because it lets agent cache interfaces, without it, every gather will do 3 to 4 enumerations, now we do it once per agent but this is still expensive for burst, it's the main bottleneck for agent creation now.

Sean-Der:
ah super cool! I think this is a great change, I would put it in `pion/transport` maybe? 

Instead of a timer would it be better to just have an atomic of 'last refreshed time'? Only then is the list of interfaces change

JoTurk:
> ah super cool! I think this is a great change, I would put it in pion/transport maybe?

cool will do that, that's much better, I was exploring the idea \:) I'll leave this PR open and update pion/transport first, and then update it here

Sean-Der:
In my head this *feels* like a really simple change. You want some time invalidated cache. Some kind of thread safe map/slice that is lockless?

JoTurk:
@Sean-Der other than the timer, the cache itself is just `atomic.Pointer[stdnet.Net]` agents can borrow the current value, the timer can update future values, it looks complex because it has timer + stop after last use logic, but once this is abstracted in pion/transport it will look simpler

agent.go
```diff
@@ -31,6 +31,28 @@ import (
...
+		return defaultNet.Load(), nil
+	}
+
+	time.AfterFunc(20*time.Second, func() {
+		defaultNet.CompareAndSwap(network, nil)
+	})
+
+	return network, nil
+}
+
```
JoTurk, line 55: Should we move this to pion/transport?

Sean-Der:
Would it be possible to use the `nil == default` pattern that Go does in some places? like [Resolver](https://cs.opensource.google/go/go/+/refs/tags/go1.27.1:src/net/lookup.go;l=133-134)

So if user doesn't pass a `stdnet.Net` it calls the default implementation that hits your cache?

JoTurk:
@Sean-Der sorry I'm confused we don't call getSharedNet unless `agent.net == nil` already or do you want to move this check to the actual impl?

https://github.com/pion/ice/pull/1003/files#diff-07afbb40d3dc2699797d28c0689ba544f67ca642d0d3efcff3c02a0bbb3604fdR271-R272

Sean-Der:
@JoTurk We might be able to get rid of `stdnet.NewNet()` entirely. 

A `nil` agent.net is safe to use, it falls back to using globals

JoTurk:
@Sean-Der oh I get what you mean, one downside to this idea is that transport Net is an interface so it will not be clean like Resolver, i can do that quickly tho :)

JoTurk:
@Sean-Der https://github.com/pion/transport/pull/407 but as i said agent.net is an interface so we'll still have to keep a condition here + ugly casting :)

JoTurk:
not safe

